PR #34, opened 2026-03-05
feat: add quality filter to mem_save and passive capture

### LINKED ISSUE #35: Quality filter for mem_save and passive capture as rtk-ai
## Problem

Right now `mem_save` will happily store observations like title "ok" with content "done". Over time this pollutes the database and makes `mem_search` and `mem_context` return noise alongside real memories. The `mem_capture_passive` has a 20-char minimum in `ExtractLearnings` but no word count check, so very short phrases still get through.

  ## Proposed solution

A small quality check before persisting in `handleSave`:
  - Title must be at least 3 chars
  - Content must have at least 5 words
  - Known placeholder values are rejected with a descriptive error

For passive capture, add a word count minimum (4 words) alongside the existing character check in `ExtractLearnings`.

  ## Alternatives considered

Putting the check inside `AddObservation` at the store level so it covers all entry points (CLI, HTTP, MCP). Decided against it because the import and sync paths shouldn't be filtered — they restore existing data.

jmeiracorbal:
## Summary

Add quality filter to mem_save and passive capture.

## Motivation

This idea is extracted from [rtk-ai/rtk](https://github.com/rtk-ai/rtk).

## Changes

- Reject observations with title shorter than 3 chars
- Reject content with fewer than 5 words
- Reject known noise/placeholder values (ok, done, todo, etc.)
- Add word count minimum (4 words) to ExtractLearnings for passive capture
- Update existing tests to use content that meets the quality threshold

## Test plan

- [x] `go test ./...` passes

Closes #35


Alan-TheGentleman (commented):
Hey, appreciate the contribution! I like part of this but want to scope it down.

The **ExtractLearnings word count** — keep that. Passive capture parses unstructured text so filtering noise there makes total sense.

The **handleSave quality gate** — I'd like to remove this part. Here's my thinking:

- `mem_save` is called by the agent, not humans. If the agent sends garbage, the real fix is better protocol instructions, not server-side validation on AI-generated content.
- The 5-word minimum is too low to catch actual quality problems. "decided to use JWT" passes — is that quality? The real noise is verbose observations that say nothing useful, and this filter won't catch those.
- The noise set is English-only. We have Spanish-speaking users ("listo", "hecho", "dale" all bypass it). It'd need constant maintenance across languages.
- Failed quality checks mean the agent retries with more text = more tokens burned for something that shouldn't happen if the protocol works.

**What to keep**: `minLearningWords` in ExtractLearnings + related test updates.

**What to remove**: `quality.go`, the `checkObservationQuality` call in handleSave, the `TestCheckObservationQuality*` / `TestHandleSaveRejects*` tests, and revert the test content padding in `TestHandleSaveCreatesProjectScopedSession` and `TestExplicitSessionIDBypassesDefault` back to their originals (those were only padded to pass the gate being removed).

Let me know if you have questions!

Alan-TheGentleman (changes requested):

PR #40, opened 2026-03-06
fix: filterByProject ignores entity-level Project fields

drccastillo:
## Summary

- `filterByProject` in `sync.go` filtered observations and prompts **only by their session's project**, ignoring each entity's own `Project` field. This caused chunks to be silently dropped during sync when an observation had the correct project but its session had an empty or different project (e.g., after topic-key upsert preserved a stale `session_id`).
- The new algorithm checks each entity's own `Project` field first, falling back to session-based matching only when the entity's project is nil/empty. Sessions are now included as dependencies of matched entities rather than as the primary filter.
- Also fixes `RecordSyncedChunk` to use `execHook` for consistency with the rest of `store.go`.

## Scenarios fixed

| Scenario | Before | After |
|----------|--------|-------|
| Observation with own project, session has empty project | **excluded** | included |
| Observation with own project, session has different project (stale session_id) | **excluded** | included |
| Observation with nil project in non-matching session | excluded | excluded |
| Prompt with own project, session has empty project | **excluded** | included |

## Test plan

- [x] New `TestFilterByProjectEntityLevel` covers all scenarios above
- [x] Existing `TestFilterFunctionsAndTimeNormalization` passes (backward compat)
- [x] Existing `TestExportImportFlowWithProjectFilter` passes
- [x] Full suite `go test ./...` passes

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -1418,7 +1418,7 @@ func (s *Store) GetSyncedChunks() (map[string]bool, error) {
 
 // RecordSyncedChunk marks a chunk as imported/exported so it won't be processed again.
 func (s *Store) RecordSyncedChunk(chunkID string) error {
-	_, err := s.db.Exec(
+	_, err := s.execHook(s.db,
 		"INSERT OR IGNORE INTO sync_chunks (chunk_id) VALUES (?)",
 		chunkID,
 	)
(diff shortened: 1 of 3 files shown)


Alan-TheGentleman (approved):
LGTM

Alan-TheGentleman (approved):
Great catch — this was silently dropping observations during sync whenever an entity had its own project but its session didn't match. Classic case of the filter being too naive about where project ownership lives.

The entity-first, session-fallback approach is the right call. Tests cover the edge cases well. The `execHook` consistency fix is a nice bonus.

This is sync layer so it benefits all agents equally. 👍

PR #43, opened 2026-03-07
fix: load deferred memory tools at session start in Claude Code

### LINKED ISSUE #42: Bug: mem_save never called engram tools are deferred in Claude Code plugin system
## Description

After installing the engram plugin in Claude Code and running sessions normally, `engram stats` consistently shows 0 observations despite sessions being created. The session-start hook fires (sessions appear in the DB), but no memory is ever saved.

## Steps to reproduce

1. Install engram via `engram setup claude-code`
2. Open Claude Code in any project
3. Work normally and make decisions, fix bugs, complete tasks
4. Run `engram stats` or `engram context` to check.

## Expected behavior

Observations are saved proactively by the agent after decisions, bug fixes, and discoveries, as described in the Memory Protocol injected by the session-start hook and in the SKILL.md.

## Actual behavior

Sessions are created, HTTP hook works fine, but `Observations: 0` is always present, no change. The agent never calls `mem_save`, `mem_context` or any other memory tool.

The Claude Code plugin system places **all** tools from any plugin into the deferred pool, regardless of the `defer_loading` annotation on the MCP tool definition. This includes the four core tools (`mem_save`, `mem_search`, `mem_context`, `mem_session_summary`) that are intentionally NOT marked with `WithDeferLoading(true)` in `internal/mcp/mcp.go`.

Because the tools are deferred, the agent does not know they exist until a `ToolSearch` call discovers them. Neither `session-start.sh` nor `SKILL.md` instructs the agent to perform this discovery step, so the tools remain unknown for the entire session and no saves ever happen.

This can be confirmed by running `engram mcp --tools=agent` and inspecting the `tools/list`. The four core tools have no `defer_loading` field, while the plugin system still treats them as deferred. The same behavior is observable with other Claude Code plugins (e.g. context7), confirming it is a platform-level constraint engram must work around on its side.

## Environment

- OS: macOS 15 (Darwin 24.6.0)
- engram version: 1.7.1
- Agent: Claude Code

jmeiracorbal:
## Summary

- Instruct the agent to call ToolSearch at session start so core memory tools are available before any proactive save triggers fire.
- Applied in both `session-start.sh` (hook context) and `SKILL.md` (always-active skill).

## Motivation

Closes #42

All Claude Code plugin tools are placed in the deferred pool by the platform, regardless of the `defer_loading` MCP annotation. The agent receives the Memory Protocol telling it to call `mem_save`, but the tool is unknown until ToolSearch discovers it. This results in 0 observations saved across all sessions.

## Changes

- `plugin/claude-code/scripts/session-start.sh`: added `FIRST ACTION` section to the injected protocol with the exact ToolSearch query for the four core tools
- `plugin/claude-code/skills/memory/SKILL.md`: added `FIRST ACTION` section at the top of the protocol so the agent loads tools even when the hook context  is not available (e.g. after compaction)

## Test plan

- [x] `go test ./...` passes
- [x] Manually tested: opened a fresh Claude Code session, confirmed ToolSearch was called at session start, confirmed `mem_save` successfully persisted an observation, verified with `engram stats` that `Observations` is no longer 0

### DIFF
--- a/plugin/claude-code/scripts/session-start.sh
+++ b/plugin/claude-code/scripts/session-start.sh
@@ -47,7 +47,7 @@ This protocol is MANDATORY and ALWAYS ACTIVE.
 
 ### FIRST ACTION — load memory tools
 Memory tools are deferred and must be loaded before use. Call ToolSearch NOW with:
-select:mcp__plugin_engram_engram__mem_save,mcp__plugin_engram_engram__mem_context,mcp__plugin_engram_engram__mem_search,mcp__plugin_engram_engram__mem_session_summary
+select:mcp__plugin_engram_engram__mem_save,mcp__plugin_engram_engram__mem_context,mcp__plugin_engram_engram__mem_search,mcp__plugin_engram_engram__mem_session_summary,mcp__plugin_engram_engram__mem_session_end
 
 ### PROACTIVE SAVE — do NOT wait for user to ask
 Call `mem_save` IMMEDIATELY after ANY of these:
--- a/plugin/claude-code/skills/memory/SKILL.md
+++ b/plugin/claude-code/skills/memory/SKILL.md
@@ -11,7 +11,7 @@ This protocol is MANDATORY and ALWAYS ACTIVE — not something you activate on d
 ## FIRST ACTION (session start)
 
 Memory tools are deferred and must be loaded before use. At the start of every session, call ToolSearch with:
-`select:mcp__plugin_engram_engram__mem_save,mcp__plugin_engram_engram__mem_context,mcp__plugin_engram_engram__mem_search,mcp__plugin_engram_engram__mem_session_summary`
+`select:mcp__plugin_engram_engram__mem_save,mcp__plugin_engram_engram__mem_context,mcp__plugin_engram_engram__mem_search,mcp__plugin_engram_engram__mem_session_summary,mcp__plugin_engram_engram__mem_session_end`
 
 ## PROACTIVE SAVE TRIGGERS (mandatory — do NOT wait for user to ask)
 


Alan-TheGentleman (commented):
The fix is pragmatic and solves the immediate problem (0 observations saved in Claude Code sessions). Two issues to address:

### 1. Missing `mem_session_end` in the select list

The ToolSearch select loads 4 tools:
\`mem_save, mem_context, mem_search, mem_session_summary\`

But \`mem_session_end\` is missing. This tool marks sessions as ended and is the MCP-level fallback if the session-stop hook (#47) fails. Without loading it, the agent can never call it — even if the Memory Protocol instructs it to.

**Requested change**: Add \`mcp__plugin_engram_engram__mem_session_end\` to the select list in both \`session-start.sh\` and \`SKILL.md\`.

### 2. Claude Code-only scope (not a blocker, just documenting)

This fix is specific to Claude Code's deferred-tools platform constraint. The other agents (OpenCode, Gemini, Codex) don't have this problem because:
- **OpenCode**: Tools are registered directly by the plugin, not deferred
- **Gemini/Codex**: MCP tools are available immediately via stdio transport

This is fine — the fix belongs in the Claude Code plugin. Just documenting for context.

### Verdict

Approve after adding `mem_session_end` to the select list.

Tech-Code1:
Hey @jmeiracorbal — the fix is solid and addresses a real blocker. To unblock the approval, Alan's requested change is a one-liner in both files:

Add `mcp__plugin_engram_engram__mem_session_end` to the select query, so the final string becomes:

```
select:mcp__plugin_engram_engram__mem_save,mcp__plugin_engram_engram__mem_context,mcp__plugin_engram_engram__mem_search,mcp__plugin_engram_engram__mem_session_summary,mcp__plugin_engram_engram__mem_session_end
```

Apply that in both `plugin/claude-code/scripts/session-start.sh` and `plugin/claude-code/skills/memory/SKILL.md` and it should be ready to approve.

jmeiracorbal:
> This solves the immediate problem — without it, mem_save literally never gets called in Claude Code. Good find and clean fix.
> 
> One change needed: add `mcp__plugin_engram_engram__mem_session_end` to the select list in both session-start.sh and SKILL.md. Right now you're loading 4 tools but missing the one that marks sessions as ended. If the session-stop hook (#47) fails for any reason, the agent has no fallback because it doesn't know the tool exists.
> 
> Everything else looks good. I'll approve once `mem_session_end` is added to the select.

Hello!

Added `mcp__plugin_engram_engram__mem_session_end` to the ToolSearch select in `session-start.sh` and SKILL.md.

PR #47, opened 2026-03-07
fix: call sessions end endpoint from session-stop hook

### LINKED ISSUE #46: Bug: sessions never get ended_at set — session-stop.sh is an empty placeholder
## Description

After using engram with Claude Code for a while, all sessions in the database have `ended_at = NULL`. The `engram sessions` list grows indefinitely and there is no way to tell from the data when a session actually ended.

The session stop: `plugin/claude-code/scripts/session-stop.sh` was only a placeholder:

```bash
# Nothing heavy to do here — placeholder
exit 0
```

The server already exposes `POST /sessions/{id}/end` (sets `ended_at = datetime('now')`), and the Stop hook already receives `session_id` via stdin. The script just needs to call the endpoint.

## Steps to reproduce

1. Install engram via `engram setup claude-code`
2. Open and close Claude Code multiple times across different projects
3. Query the database: `sqlite3 ~/.engram/engram.db "SELECT id, started_at, ended_at FROM sessions ORDER BY started_at DESC LIMIT 10;"`

## Expected behavior

When a Claude Code session ends, `ended_at` is set on the session record so the duration is known and the session is clearly closed.

## Actual behavior

Every session has `ended_at = NULL`. The Stop hook defined in `hooks.json` runs `session-stop.sh`, but that script contains only a placeholder comment and `exit 0`. The HTTP endpoint `POST /sessions/{id}/end` exists and is ready to use, but nothing calls it from the hook.

The Memory Protocol instructs the agent to call `mem_session_summary` before ending, but this relies on the agent cooperating. If the user closes the terminal, switches projects, or if the deferred-tools bug prevents `mem_session_summary` from being available, the session never gets closed at the DB level.

## Environment

- OS: macOS 15 (Darwin 24.6.0)
- engram version: 1.7.1
- Agent: Claude Code

jmeiracorbal:
## Summary

- Implement `session-stop.sh` to call `POST /sessions/{id}/end` when Claude Code stops, so `ended_at` is always set regardless of agent cooperation

## Motivation

Closes #46

`session-stop.sh` was a placeholder that did nothing. The server already exposed `POST /sessions/{id}/end`, and the Stop hook already received `session_id` via stdin, but the connection was never made. As a result, every session in the database has `ended_at = NULL` and sessions grow unbounded with no duration data.

The fix mirrors the pattern already used in `session-start.sh`: read `session_id` from stdin, call the HTTP API.

## Changes

- `plugin/claude-code/scripts/session-stop.sh` — replaced placeholder with a curl call to `POST /sessions/{id}/end`. Reads `session_id` from stdin (same as `session-start.sh`), exits cleanly if no session_id is present. Runs async (already configured in `hooks.json`) so it never blocks Claude's response.

## Test plan

- [x] `go test ./...` passes
- [x] Manually tested end-to-end:
  1. Created a session via `POST /sessions`
  2. Confirmed `ended_at = NULL` in the database
  3. Simulated the hook: `echo '{"session_id":"test","cwd":"/tmp"}' | bash session-stop.sh`
  4. Confirmed `ended_at` was set correctly in the database

### DIFF
--- a/plugin/claude-code/scripts/session-stop.sh
+++ b/plugin/claude-code/scripts/session-stop.sh
@@ -1,11 +1,23 @@
 #!/bin/bash
 # Engram — Stop hook for Claude Code (async)
 #
-# Tracks tool call count per session. Runs async so it doesn't
-# block Claude's response.
+# Marks the session as ended via the HTTP API.
+# Runs async so it doesn't block Claude's response.
 
-# Nothing heavy to do here — the Memory Protocol in the skill
-# instructs the agent to call mem_session_summary before ending.
-# This hook exists as a placeholder for future heartbeat/tracking.
+ENGRAM_PORT="${ENGRAM_PORT:-7437}"
+ENGRAM_URL="http://127.0.0.1:${ENGRAM_PORT}"
+
+INPUT=$(cat)
+SESSION_ID=$(echo "$INPUT" | jq -r '.session_id // empty')
+
+if [ -z "$SESSION_ID" ]; then
+  exit 0
+fi
+
+curl -sf "${ENGRAM_URL}/sessions/${SESSION_ID}/end" \
+  -X POST \
+  -H "Content-Type: application/json" \
+  -d '{}' \
+  > /dev/null 2>&1
 
 exit 0


Alan-TheGentleman (commented):
LGTM

Alan-TheGentleman (commented):
Fix looks good — clean, mirrors session-start.sh, does exactly what it should.

One thing worth noting: if the server happens to be down when the session ends, this fails silently (which is fine for a hook). But the agent also can't call `mem_session_end` as a fallback because that tool isn't in the ToolSearch select list from #43. Not a blocker for this PR — just something to address in #43.

Also flagging that this only covers Claude Code. OpenCode, Gemini, and Codex don't have session-end hooks so their sessions still get `ended_at = NULL`. We should open a follow-up for that — the MCP tool exists, agents just need to be told to call it.

Approve from our side, good to merge.

PR #50, opened 2026-03-08
feat(cloudstore): add ErrNotFound sentinel and fix 404 responses

Tech-Code1:
## Summary

- Introduces `cloudstore.ErrNotFound` as a sentinel error alongside the existing `ErrProjectSyncPaused`, following the same pattern already established in the package.
- `GetObservation`, `GetChunk`, and `GetPrompt` now wrap `sql.ErrNoRows` with `ErrNotFound` using `%w`, so callers can use `errors.Is` to distinguish "not found" from real database failures.
- `writeStoreError` in the HTTP layer now maps `ErrNotFound` to **HTTP 404** instead of falling through to 500.
- `handlePullChunk` is simplified: the previous implementation had two duplicate branches both returning 404 and also checked `sql.ErrNoRows` directly (bypassing the wrapping done in `GetChunk`). It now delegates to `writeStoreError`.

## Motivation

Before this change, requesting a resource that does not exist (e.g. a chunk or observation by ID) returned **HTTP 500 Internal Server Error** instead of **404 Not Found**. Clients — including other AI agents using the MCP sync tools — could not distinguish a missing resource from an internal database failure, making error handling and retry logic incorrect.

## Test plan

- [x] New `errors_test.go` with unit tests for `writeStoreError` (no Docker required):
  - `ErrNotFound` → 404
  - Wrapped `ErrNotFound` via `fmt.Errorf("%w")` → 404 (verifies `errors.Is` chain)
  - DB connection error → 503
  - Generic error → 500
- [x] `isDBConnectionError` table-driven tests covering all recognized error strings
- [x] All existing `cloudserver` tests pass (`SKIP_DOCKER_TESTS=1`)
- [x] Full project builds without errors (`go build ./...`)

### DIFF
--- a/internal/cloud/cloudstore/cloudstore.go
+++ b/internal/cloud/cloudstore/cloudstore.go
@@ -224,6 +224,9 @@ func (cs *CloudStore) GetUserByUsername(username string) (*CloudUser, error) {
 		username,
 	).Scan(&u.ID, &u.Username, &u.Email, &u.PasswordHash, &u.APIKeyHash, &u.CreatedAt, &u.UpdatedAt)
 	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, fmt.Errorf("cloudstore: get user by username: %w", ErrNotFound)
+		}
 		return nil, fmt.Errorf("cloudstore: get user by username: %w", err)
 	}
 	return &u, nil
@@ -238,6 +241,9 @@ func (cs *CloudStore) GetUserByEmail(email string) (*CloudUser, error) {
 		email,
 	).Scan(&u.ID, &u.Username, &u.Email, &u.PasswordHash, &u.APIKeyHash, &u.CreatedAt, &u.UpdatedAt)
 	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, fmt.Errorf("cloudstore: get user by email: %w", ErrNotFound)
+		}
 		return nil, fmt.Errorf("cloudstore: get user by email: %w", err)
 	}
 	return &u, nil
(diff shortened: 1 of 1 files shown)


Alan-TheGentleman (commented):
Good stuff — the `ErrNotFound` sentinel is the right pattern and matches what we already do with `ErrProjectSyncPaused`. The `handlePullChunk` cleanup removing `database/sql` from the handler layer is clean.

A couple things before we merge:

1. **Cover the other Get methods** — you wrapped `ErrNotFound` in `GetObservation`, `GetPrompt`, and `GetChunk`, but `GetSession` (and any other single-row Get in cloudstore) should follow the same pattern. Otherwise we'll have inconsistent 404 behavior depending on which resource is missing. Can you do a pass on the remaining Get methods?

2. **Minor:** the fallback message in `handlePullChunk` changed from `"chunk not found"` to `"failed to fetch chunk"` — that's fine since `writeStoreError` catches `ErrNotFound` first, but just making sure it was intentional and not a leftover.

Once (1) is addressed this is good to go.

Tech-Code1:
Done — extended `ErrNotFound` to all remaining single-row Get methods: `GetUserByUsername`, `GetUserByEmail`, `GetUserByAPIKeyHash`, `GetUserByID`, and `GetSession`. All follow the same pattern now.

Re point (2): the fallback message change in `handlePullChunk` from `"chunk not found"` to `"failed to fetch chunk"` was intentional. Since `writeStoreError` catches `ErrNotFound` first and returns `"not found"`, the fallback only fires for genuine internal errors (DB down, scan failure, etc.) — so `"failed to fetch chunk"` is more accurate in that context.

PR #51, opened 2026-03-08
feat(cloudserver): add input validation for register and search endpoints

Tech-Code1:
## Summary

- `POST /auth/register` now validates username format/length, email structure, and password length before hitting the database.
- `GET /sync/search` now enforces a 500-character cap on the query string and clamps `limit` to 1–100.
- Validation lives in a new `validators.go` file — pure functions, no dependencies, easy to extend.

## Motivation

Both endpoints accepted arbitrary input with no bounds checking:

- A username of `""` or 200 characters was accepted.
- An email like `notanemail` passed through to Postgres, which would reject it with a cryptic constraint error.
- A password longer than 72 characters would be **silently truncated by bcrypt** — the user would believe they have a strong password, but only the first 72 bytes would be hashed.
- `GET /sync/search?limit=999999` would issue an unbounded Postgres FTS query.
- A 50,000-character `q` parameter would be passed directly to the FTS engine.

## Changes

| Endpoint | What is validated | Status on failure |
|----------|-------------------|-------------------|
| `POST /auth/register` | Username: 3–50 chars, `[a-zA-Z0-9_-]` only | 400 |
| `POST /auth/register` | Email: must have `@`, domain must contain `.` | 400 |
| `POST /auth/register` | Password: max 72 chars (bcrypt hard limit) | 400 |
| `GET /sync/search` | Query: max 500 chars | 400 |
| `GET /sync/search` | Limit: 1–100 | 400 |

The existing `auth.ErrWeakPassword` check (min 8 chars) in the auth layer is preserved — both validations coexist.

## Test plan

- [x] `TestValidateRegisterRequest` — 13 table-driven cases covering valid input and every rejection path
- [x] `TestValidateSearchParams` — 8 cases covering bounds and edge values
- [x] All existing `cloudserver` tests pass (`SKIP_DOCKER_TESTS=1`)
- [x] Full project builds without errors (`go build ./...`)

### DIFF
--- a/internal/cloud/cloudserver/validators.go
+++ b/internal/cloud/cloudserver/validators.go
@@ -26,23 +26,39 @@ func validateRegisterRequest(username, email, password string) error {
 		return fmt.Errorf("username must be at most %d characters", maxUsernameLen)
 	}
 	for _, r := range username {
-		if !unicode.IsLetter(r) && !unicode.IsDigit(r) && r != '_' && r != '-' {
-			return fmt.Errorf("username may only contain letters, digits, underscores, and hyphens")
+		if !unicode.IsLetter(r) && !unicode.IsDigit(r) && r != '_' && r != '-' && r != '.' {
+			return fmt.Errorf("username may only contain letters, digits, underscores, hyphens, and dots")
 		}
 	}
 
-	// Email — basic structure check: must have exactly one @, and a dot after it
+	// Email — basic structure check: must have @ with content before it.
+	// The domain check only requires a dot when the domain has more than one
+	// segment, so user@localhost is accepted for local/dev environments.
 	atIdx := strings.Index(email, "@")
 	if atIdx < 1 {
 		return fmt.Errorf("email must contain an @ sign")
 	}
 	domain := email[atIdx+1:]
-	if !strings.Contains(domain, ".") || strings.HasPrefix(domain, ".") || strings.HasSuffix(domain, ".") {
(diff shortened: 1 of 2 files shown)


Alan-TheGentleman (changes requested):
The validation layer is a solid idea — catching bad input before it hits Postgres is the right call, and the bcrypt 72-char guard is a nice detail.

A few things to fix before this can go in:

1. **This PR depends on #50** — the diff includes the `ErrNotFound` check in `writeStoreError` which doesn't exist on main yet. Please rebase onto #50 or wait until it's merged, then rebase onto main.

2. **Username charset is too restrictive** — right now it rejects dots and unicode characters, so usernames like `alan.b` or `señor_dev` would fail. For an open source project with global users, we should at minimum allow dots (`.`). Consider whether unicode letters should be allowed too (Go's `unicode.IsLetter` already handles this, you just need to keep it in the allowed set).

3. **Password min length not validated** — the comment says it "mirrors the 8-char floor in auth.ErrWeakPassword" but the function only checks the max (72). Either validate both min and max here (defense in depth), or drop the comment so it doesn't mislead. I'd prefer validating both since we're already doing input validation at this layer.

4. **Email rejects `user@localhost`** — this is valid in local/dev environments. Not a hard blocker, but if it's easy to relax (e.g. only require a dot if the domain has more than one segment), that'd be nice. Up to you on this one.

Fix (1), (2), and (3) and this is ready.

PR #110, opened 2026-03-21
docs: fix build-from-source instructions and add Claude CLI version note

### LINKED ISSUE #67: docs: clarify install-from-source version stamp and Claude Code CLI minimum version
## Problem statement

```
```

## Proposed approach

## Risks/tradeoffs

## Affected areas

- Docs (README.md only)

Kevinsillo:
## Summary

- **INSTALLATION.md**: Split the optional version-stamp step into separate, copy-pasteable code blocks (`go build` for local testing vs `go install` for system-wide install). Previously `go build -o engram` left the versioned binary in the current directory without installing it — users who followed the "optional" step ended up with a stale binary in `$GOPATH/bin`.
- **AGENT-SETUP.md & PLUGINS.md**: Add a note about the minimum Claude Code CLI version required (`>= 2.1.70`) for the marketplace plugin. Older versions fail with `Invalid schema: plugins.0.source: Invalid input`.

Closes #67

## Test plan

- [ ] Verify the `go install -ldflags` command works and places the versioned binary in `$GOPATH/bin`
- [ ] Verify the `go build -ldflags` command builds the binary in the current directory
- [ ] Confirm the Claude CLI version note renders correctly in both docs

### DIFF
--- a/docs/AGENT-SETUP.md
+++ b/docs/AGENT-SETUP.md
@@ -66,6 +66,11 @@ See [Plugins → OpenCode Plugin](PLUGINS.md#opencode-plugin) for details on wha
 
 > **Prerequisite**: Install the `engram` binary first (via [Homebrew](INSTALLATION.md#homebrew-macos--linux), [Windows binary](INSTALLATION.md#windows), [binary download](INSTALLATION.md#download-binary-all-platforms), or [source](INSTALLATION.md#install-from-source-macos--linux)). The plugin needs it for the MCP server and session tracking scripts.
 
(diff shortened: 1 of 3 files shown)


Alan-TheGentleman (approved):
Buenísimo, @Kevinsillo. Todo impecable.

La separación entre `go build` (local) y `go install` (system-wide) era necesaria, antes confundía y los usuarios terminaban con un binario suelto sin instalar. Ahora queda claro qué hace cada uno.

La nota de versión mínima del CLI (>= 2.1.70) con el mensaje de error exacto es un golazo para que nadie pierda tiempo debuggeando.

Dale para adelante.

Alan-TheGentleman:
Closing this because it is superseded by the newer path in #254 for the same approved issue. Keeping both open creates duplicate review paths for one fix. Please continue that work in #254.

PR #161, opened 2026-04-07
perf(claude-code): shrink SessionStart hook injection from ~10 KB to ~2 KB

todie:
Addresses #163 (partial — this PR captures the shell-only win; full fix lands with #162).

## Motivation

Every time a user starts a new Claude Code session (or a session is recovered from compaction), the engram plugin's `SessionStart` hook in [`plugin/claude-code/scripts/session-start.sh`](../blob/main/plugin/claude-code/scripts/session-start.sh) injects the contents of `stdout` into the agent's context window as `additionalContext`. On a busy project I benchmarked against (200+ observations across several months of work), that injection was **~9.8 KB** — roughly **2,500 tokens burned on every single session start**, before the user has typed a single word.

That's a meaningful ongoing cost:
- Claude Code users open many sessions per day (one per VSCode window, one per `claude` invocation, one per `/clear`, etc.). On my machine I routinely have 5–7 concurrent runtimes alive.
- The injected content is 80%+ redundant — either a verbatim copy of the `engram:memory` skill (which already auto-loads on demand) or inline 300-char previews of observations the agent rarely needs the body of at session start (titles alone are enough to jog memory).
- Token budgets matter more than ever now that agent memory systems are stacking on top of each other. Engram competing with its own hook for context window space is a bad look.

Breakdown of the original 9.8 KB:

1. **~1.8 KB** is a hardcoded `## Engram Persistent Memory — ACTIVE PROTOCOL` heredoc that exhaustively re-teaches the save/search/close protocol on every session. That protocol is **already shipped in `skills/memory/SKILL.md`** in this same plugin, and the skill auto-loads on demand via Claude Code's skill system. Teaching it eagerly + teaching it lazily via skill is pure duplication.
2. **~8 KB** is the rendered `/context` payload. The server-side `FormatContext` in [`internal/store/store.go`](../blob/main/internal/store/store.go) inlines up to 300 characters of **raw, unflattened** `obs.Content` into each observation bullet. Since `obs.Content` is often multi-line markdown (session summaries are whole documents), a single observation can easily spill across 5+ lines in the output. Combined with the default `MaxContextResults=20`, a busy project's `/context` response routinely hits 8 KB.

The motivating question: how do we cut this without losing anything the agent actually uses?

## Design

This PR does the minimum viable fix — **shell-only, single file**. A follow-up PR (stacked, coming next) does the cleaner server-side version; this PR lets users who can't easily upgrade their engram binary capture most of the win via a plugin update alone.

### Change 1: Shrink the `PROTOCOL` heredoc (35 lines → 9 lines)

The new heredoc keeps:
- One line announcing that engram memory is active
- The full list of core + admin tool names (so the agent can reference them without a ToolSearch round-trip)
- A pointer to the `engram:memory` skill for the full protocol

It drops the 30-line "PROACTIVE SAVE / SEARCH MEMORY / SESSION CLOSE" block, because that block is the exact content of `skills/memory/SKILL.md`. The skill system will pull it when the agent needs it.

### Change 2: Compact the `/context` response via `awk`

Since this PR doesn't touch the server, we post-process the server response in the hook. The awk pass:

1. **Concatenates each observation's continuation lines** into a single logical line. The server's raw output has multi-line observation bodies; we flatten them.
2. **Collapses whitespace** (`gsub(/[[:space:]]+/, " ", buf)`).
3. **Caps per-bullet length** at `$CTX_MAXLEN` chars (default 140) with a `…` ellipsis.
4. **Caps bullet count** at `$CTX_LIMIT` (default 8).
5. Passes headers, the sessions section, and the prompts section through unchanged.

Both caps are env-overridable (`ENGRAM_CONTEXT_LIMIT`, `ENGRAM_CONTEXT_MAXLEN`) so anyone who wants the old verbose behavior back can get it without editing the hook.

### Byte-compatibility & risk

- **Zero Go changes.** The engram binary is untouched. Users upgrading the plugin without upgrading their binary lose nothing.
- **Server contract unchanged.** `/context` still takes `project` and `scope`; response shape is the same `{context: string}`.
- **awk portability.** The script uses only POSIX awk features (`function`, `gsub`, `substr`, `[[:space:]]` class). Verified on gawk; should work on mawk and BSD awk.
- **Rollback is `git revert`.** Nothing to migrate, nothing to un-migrate.
- **UTF-8 nit.** `substr()` truncates at byte count, not rune count, so a bullet cut at byte 140 could emit a half-encoded multi-byte character before `…`. In practice observation bodies are overwhelmingly ASCII, and the downstream consumer is an LLM which tolerates malformed UTF-8. Not fixing inline; happy to if a reviewer cares.

## Measurement

Benchmarked on a real project with 200+ observations:

| | Before | After | Δ |
|---|---:|---:|---:|
| `/context` payload (rendered) | 7961 B | 1487 B | **−81%** |
| `PROTOCOL` heredoc | ~1800 B | ~450 B | **−75%** |
| **Total hook stdout** (`additionalContext`) | **~9800 B** | **~1900 B** | **−80%** |

Token savings scale with observation volume; empty/small projects see less dramatic numbers because the observations section is already short.

## Test plan

- [x] End-to-end smoke: `echo '{"session_id":"t","cwd":"/tmp"}' | bash plugin/claude-code/scripts/session-start.sh` produces well-formed output, correct byte count
- [x] awk pass handles multi-line observation bodies (session summaries with embedded markdown headers/lists) — verified against a real `/context` payload
- [x] Empty project still produces a minimal well-formed injection (just the PROTOCOL pointer)
- [x] `ENGRAM_CONTEXT_LIMIT=0` returns no observations; `ENGRAM_CONTEXT_MAXLEN=9999` returns full bullets — both behave as expected
- [ ] Reviewer to sanity-check on a fresh project

## Follow-up

I have a stacked PR ready that adds `limit=N` and `compact=1` query parameters to `GET /context` in the Go server, with a byte-compatible `FormatContext` wrapper that keeps all existing callers unchanged. That brings the end-to-end hook injection down from ~1.9 KB (this PR) to ~1.1 KB (this PR + the follow-up) and removes the need for the awk pass on matched-version deployments. I'll open it against this branch as soon as this lands — or against `main` directly if you'd prefer to review them independently.

### DIFF
--- a/plugin/claude-code/scripts/session-start.sh
+++ b/plugin/claude-code/scripts/session-start.sh
@@ -4,10 +4,21 @@
 # 1. Ensures the engram server is running
 # 2. Creates a session in engram
 # 3. Auto-imports git-synced chunks if .engram/manifest.json exists
-# 4. Injects Memory Protocol instructions + memory context
+# 4. Injects a minimal tool-availability pointer + compacted memory context
+#
+# Memory protocol (when/what to save, search, close) lives in the
+# `engram:memory` skill shipped with this plugin and is loaded on demand.
+# Re-injecting the full protocol on every SessionStart wastes ~1.8 KB of
+# context window per session, so this script only emits a short pointer.
 
 ENGRAM_PORT="${ENGRAM_PORT:-7437}"
 ENGRAM_URL="http://127.0.0.1:${ENGRAM_PORT}"
+
+# Tunables (override via env)
+#   ENGRAM_CONTEXT_LIMIT   — max observations to inject (default 8)
+#   ENGRAM_CONTEXT_MAXLEN  — max chars per observation line (default 140)
+CTX_LIMIT="${ENGRAM_CONTEXT_LIMIT:-8}"
+CTX_MAXLEN="${ENGRAM_CONTEXT_MAXLEN:-140}"
 
 # Load shared helpers
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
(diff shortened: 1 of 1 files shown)


Alan-TheGentleman (changes requested):
## Review — REQUEST CHANGES

La motivación es válida y el trabajo técnico es sólido, pero hay un problema crítico de compatibilidad y un problema conceptual con el truncamiento del contexto.

### CRITICAL: Pérdida de protocolo para agentes sin sistema de skills

El nuevo output dice "Full protocol lives in the engram:memory skill — load it on demand." Pero los agentes que NO son Claude Code (Codex, custom MCP clients, sub-agentes) no tienen sistema de skills. Resultado: las tools están disponibles pero el agente no sabe CUÁNDO usarlas.

**Solución**: Agregar ~3 líneas con el protocolo mínimo al heredoc:

```
Save proactively: call mem_save after decisions, bug fixes, discoveries, conventions.
Search first: call mem_search when user references past work or starts a new topic.
Close sessions: call mem_session_summary before ending (Goal, Discoveries, Accomplished, Next Steps, Relevant Files).
```

Costo: ~250 bytes. Seguís ahorrando ~80% de tokens. Beneficio: compatibilidad universal.

### HIGH: El truncamiento de previews rompe el flow de búsqueda

El flow de engram funciona así:
1. `mem_search` → resultados con previews
2. El agente lee los previews y decide cuál es relevante
3. `mem_get_observation` por ID → contenido completo

El preview en el `/context` del hook es el **primer filtro** que tiene el agente para decidir si una observación le sirve. Truncar a 140 chars le quita información para esa decisión. Consecuencias:
- Más calls a `mem_get_observation` para compensar → **más tokens en total** (ahorro neto potencialmente negativo)
- O peor: el agente IGNORA observaciones relevantes porque el título solo no alcanzó

Los previews de contenido (los 300 chars actuales) son parte integral del mecanismo de búsqueda. No son "exceso" — son el índice que el agente usa para decidir qué leer completo.

**Sugerencia**: Mantener los previews de contenido en el output del context. Si querés reducir tokens, usá `limit` para controlar la CANTIDAD de observations, no la CALIDAD de cada una.

### HIGH: Inconsistencia con post-compaction.sh

`post-compaction.sh` NO se toca y sigue con el heredoc completo (~1.8KB). Sesión normal = protocolo mínimo, post-compaction = protocolo full. Esto genera confusión en mantenimiento.

### MEDIUM: Truncamiento UTF-8

`substr()` trunca por bytes, no por runes. Caracteres multi-byte (emojis, acentos, CJK) pueden quedar rotos en el byte 139. Lo documentás en el PR body, que está bien, pero es un code smell.

### MEDIUM: Categorización silenciosa de tools

`mem_update`, `mem_session_start`, etc. pasan de "ToolSearch" a "Core tools". Es una MEJORA pero no está documentada en el PR body.

### LOW: CTX_MAXLEN=140 es agresivo

El `[type] **title**` consume 40-60 chars, dejando ~80 para contenido. Considerar 200 si se mantiene el truncamiento.

### Lo bueno
- El ahorro de tokens es real y la motivación es válida
- El fallback awk para binarios viejos es inteligente
- Las env vars CTX_LIMIT y CTX_MAXLEN dan configurabilidad
- La calidad del PR description es excepcional

### Resumen
El PR necesita: (1) agregar protocolo mínimo al heredoc para agentes sin skills, (2) reconsiderar el truncamiento de previews que rompe el flow de búsqueda de engram, (3) sincronizar post-compaction.sh. Con esos cambios, es un gran PR.

Alan-TheGentleman:

PR #162, opened 2026-04-07
feat(context): add limit & compact query params to GET /context (byte-compatible)

### LINKED ISSUE #163: perf: SessionStart hook injects ~10 KB (~2.5k tokens) into every Claude Code session
> **Heads up, fixes already in flight.** PRs #161 (shell-only) and #162 (server-side, stacked on #161) are open and together drop the injection from ~9.8 KB to ~1.1 KB (−88%) with byte-compatible defaults, full tests, and benchmark numbers. Filing this issue as the standalone \"why\" writeup so there's a durable place to point at — didn't want to just walk in, bitch about token costs, and leave without bringing the patch.

## TL;DR

On any project with a non-trivial observation history, the engram plugin's `SessionStart` hook at [`plugin/claude-code/scripts/session-start.sh`](../blob/main/plugin/claude-code/scripts/session-start.sh) injects roughly **9–10 KB** of markdown into Claude Code's `additionalContext` on every new session — about **2,500 tokens burned before the user has typed a single character**. ~80% of that is duplication or verbose formatting that the agent doesn't actually need at session start. This scales linearly with how often users open sessions (many times per day for power users), and it competes with the user's own work for context window space.

## Measurement

Taken on a real project with ~200 observations across several months, `personal` scope, mix of `session_summary` / `pattern` / `discovery` / `decision` / `bug` / `manual` types. Hook stdout measured with `wc -c` on the actual bytes fed to Claude Code.

```
total SessionStart hook stdout : ~9800 B    (~2500 tokens)
├── PROTOCOL heredoc            : ~1800 B   hardcoded in session-start.sh
└── /context payload            : ~8000 B   fetched from GET /context
    ├── Recent Sessions section : ~500 B
    ├── Recent Prompts section  : ~300 B
    └── Recent Observations     : ~7200 B   ← dominant cost
```

This is the injection cost **per session start**. Power users routinely keep 5–7 concurrent Claude Code runtimes alive (one per VSCode window, one per terminal `claude` invocation). Every `/clear` also re-triggers `SessionStart`. The compounded daily cost is substantial.

### 1. The PROTOCOL heredoc (~1.8 KB) is a verbatim duplicate of `skills/memory/SKILL.md`

[`plugin/claude-code/scripts/session-start.sh`](../blob/main/plugin/claude-code/scripts/session-start.sh) contains a 35-line heredoc that exhaustively re-teaches the save/search/close protocol on every session — \"PROACTIVE SAVE — do NOT wait for user to ask\", the self-check checklist, the \"SEARCH MEMORY when\" list, the session-close reminder, etc.

The **exact same content** ships as a Claude Code skill in this same plugin at [`skills/memory/SKILL.md`](../blob/main/plugin/claude-code/skills/memory/SKILL.md). Claude Code's skill system auto-loads skills on demand. So the agent learns the protocol twice: eagerly at session start (paid every time), and lazily via the skill when it actually needs the rules (paid only when needed).

Eager duplication pays the cost 100% of the time even though the lazy path covers all actual uses.

### 2. The `/context` endpoint inlines full observation bodies

[`internal/store/store.go`](../blob/main/internal/store/store.go)'s `FormatContext` renders each observation bullet as:

```
- [type] **title**: <first 300 chars of obs.Content>
```

Because `obs.Content` is often multi-line markdown (session summaries are entire documents with headers, lists, and paragraphs), a single observation can easily spill across 5+ lines in the rendered output. Combined with `MaxContextResults = 20`, a busy project's `/context` response routinely hits 7–8 KB.

On a real session, the rendered output looks like this (one observation, abridged):

```markdown
- [session_summary] **Session summary: ctodie**: ## Goal
Check for dangling sessions and runtime/process leaks across engram, Claude Code, docker, and dev servers — then clean them up.

## Discoveries
- **7 concurrent Claude Code runtimes** were alive at session start (one per VSCode window / terminal claude). Each spawns its own `engram mcp --too...
```

Multiply that by 20 observations and the section is massive. And for the agent's actual needs at session start — \"do I have prior context on this project, and what was I working on?\" — **titles alone are almost always sufficient**. The body is only needed when the agent decides to drill into a specific memory, which is exactly what `mem_search` / `mem_get_observation` are for.

## Why titles are enough at session start

The agent uses the injected context for exactly one thing: deciding whether to pull in more context before answering the user's first message. For that decision, it needs to know:

1. **Is there prior work on this project?** → counts of recent sessions (headers)
2. **What topics/decisions exist?** → observation titles
3. **Anything from the last few hours?** → timestamps + session summaries

Of those, only #2 and #3 touch the observations section, and both are answered by titles + timestamps. If the agent decides something in the list is relevant, it already has `mem_search` and `mem_get_observation` to pull the full body. Eagerly inlining 300 chars of body preview for all 20 observations is paying upfront for a fetch the agent usually doesn't need.

## Why this matters

- **Direct token cost.** ~2,500 tokens × (sessions opened per day) × (users) is a meaningful fraction of the total context window cost, and unlike work the agent actually does, this is pure overhead.
- **Context window scarcity.** Every byte the hook injects is a byte not available for the user's actual code, docs, or conversation. On long sessions this bites.
- **Cold-start perception.** The injection lands before the first user turn. On a slow pipe (or a session recovered from compaction) it's visible latency.
- **Scaling pressure.** The cost grows with observation count. Power users with mature engram databases are punished proportionally to how much they've invested in the tool.

## Proposed fixes (two PRs, stacked)

I have two PRs open that address the two sources independently:

- **#161** — `perf(claude-code): shrink SessionStart hook injection from ~10 KB to ~2 KB`
  Shell-only, single file. Shrinks the PROTOCOL heredoc to a 9-line pointer that lists the tool names and directs the agent to `skills/memory/SKILL.md` for the rules. Adds an `awk` post-processor to the hook that flattens multi-line observation bodies onto single lines, caps per-bullet length at `ENGRAM_CONTEXT_MAXLEN` (default 140), and caps bullet count at `ENGRAM_CONTEXT_LIMIT` (default 8). Both env-tunable. **Zero Go changes.** Lets users capture most of the win via a plugin update alone, without rebuilding the binary.

- **#162** — `feat(context): add limit & compact query params to GET /context` *(stacked on #161)*
  Server-side. Adds `limit=N` and `compact=1` query params to `GET /context`, implemented via a new `ContextOptions` struct and `FormatContextWithOptions` method. **Byte-compatible** with the old behavior: `FormatContext(project, scope)` stays as a thin wrapper over `FormatContextWithOptions(project, scope, ContextOptions{})`, so all ~15 existing callers and test fixtures are untouched. Empirically verified: `GET /context` with no new params returns byte-for-byte identical output to the pre-change binary. The hook is updated to pass `?limit=8&compact=1` on matched-version deployments; the awk fallback from #161 stays in place for mixed-version deployments.

## Measured impact

| stage | SessionStart injection |
|---|---:|
| today | ~9800 B |
| after #161 | ~1900 B (−81%) |
| after #161 + #162 | **~1151 B (−88%)** |

`/context` endpoint cost (server-side):

| request | old binary | new binary |
|---|---:|---:|
| `/context?project=X` (no new params) | 7949 B | **7949 B** (byte-identical) |
| `/context?project=X&limit=8&compact=1` | n/a | **728 B** (−91%) |

## Alternatives considered

1. **Drop the PROTOCOL heredoc entirely and rely solely on the skill.** Rejected — the heredoc's tool-list section (announcing which `mem_*` tools exist) is genuinely useful for the agent's first turn because it avoids a ToolSearch round-trip. The fix is to keep the tool list and drop the 30 lines of prescriptive \"when to save\" rules that the skill already covers.

2. **Lower `MaxContextResults` globally.** Rejected — that would affect every consumer of `RecentObservations`, not just the hook. The hook is the only caller that demands short output; other callers (CLI `engram context`, TUI, MCP `mem_context` tool) may legitimately want the full 20. The fix belongs at the render layer, not the data layer.

3. **Cache the `/context` response.** Rejected — doesn't address the underlying \"we're rendering too much\" problem, and cache invalidation on `mem_save` is non-trivial. Fixing the render is simpler and strictly better.

4. **Move the PROTOCOL to a one-time `UserPromptSubmit` injection on the first turn instead of `SessionStart`.** Considered but rejected as a separate concern — the `UserPromptSubmit` hook in [`plugin/claude-code/scripts/user-prompt-submit.sh`](../blob/main/plugin/claude-code/scripts/user-prompt-submit.sh) already does a one-time ToolSearch injection for the `mem_*` tools, and conflating the two would couple the tool-loading concern with the protocol-teaching concern. Separate refactor for another day.

## Risk & rollback

Both PRs are low-risk:

- **#161** has zero Go changes. Server contract unchanged. Rollback is `git revert` on a single shell file.
- **#162** preserves byte-compatible output when called with zero-value options (empirically verified). No database migration, no config format change. Rollback is `git revert` on the merge commit. Breakage would be caught by `TestFormatContextWithOptions`'s exact-string-equality assertion against `FormatContext`, which fails loudly on any semantic drift.

Full test suite passes on both branches (`go test ./...` → 748 → 749 passed).

## Environment

- OS: Linux (WSL2 on Windows 11)
- Go: 1.24.1
- engram: `1.12.0-beta.1.0.20260407055054-49dc372fe2e4` (built from this PR branch)
- Claude Code: desktop app, latest

Happy to split the issue or the PRs further if the scope feels too big to land as one change.

todie:
Closes #163 (together with #161).

## Stacking note

**This PR is stacked on top of #161** (perf: shrink SessionStart hook injection). Review #161 first. Because GitHub doesn't support cross-repo stacked PRs with a fork base, this PR's diff against `main` currently **includes #161's changes**. Once #161 lands, I'll rebase and force-push-with-lease, and the diff will shrink to just this PR's additions (5 files, +192/−17).

If you'd prefer, I can open this as a single combined PR instead — just let me know.

## Motivation

#161 solved the SessionStart token cost client-side, in the hook, using `awk`. That's the right shape for users who can't easily rebuild the engram binary, but it has drawbacks as a long-term solution:

1. **The knob lives far from the data.** The decision about how much context to render should be made in the Go code that knows the data model, not in a shell script that parses the rendered output. Every client (MCP, the CLI, the TUI, external agents) currently has to re-implement its own compaction if it wants short output.
2. **Compact truncation is fragile in shell.** The awk pass works but has edge cases — byte vs rune truncation, multi-line observation bodies with embedded `### ` headers that could confuse section tracking, etc. Doing it in Go is about 20 lines and covers all of them correctly.
3. **Shell compaction can't undo server-side work.** The server still serializes the full 300-char content preview for every observation on every request, even if the client immediately throws it away. It's wasted CPU and memory on the engram daemon.

This PR moves the knobs into the server via two new query parameters on `GET /context`, while preserving **byte-for-byte backward compatibility** for existing callers.

## Design: byte-compatible interface via zero-value options struct

The core requirement: **don't break any existing caller**. `FormatContext` is called from:

- `internal/server/server.go` (HTTP handler) — 1 site
- `internal/mcp/mcp.go` (MCP context tool) — 1 site
- `cmd/engram/main.go` (CLI `context` command) + test shim `storeFormatContext` — 2 sites
- `internal/store/store_test.go` — 8 direct test assertions
- `cmd/engram/main_extra_test.go` — 2 shim overrides

Changing the signature of `FormatContext(project, scope string)` would cascade to **~15 call sites**, many of them in tests that pin exact output strings. That's too much churn for a feature addition.

The pattern I chose — used elsewhere in this codebase (`SearchOptions`, `AddObservationParams`) — is the **options struct + wrapper** idiom:

```go
type ContextOptions struct {
    Limit   int  // 0 → use MaxContextResults default
    Compact bool // false → legacy behaviour (inline 300-char preview)
}

// FormatContext is a thin wrapper that forwards a zero-value ContextOptions
// to FormatContextWithOptions. This keeps every existing caller unchanged.
func (s *Store) FormatContext(project, scope string) (string, error) {
    return s.FormatContextWithOptions(project, scope, ContextOptions{})
}

func (s *Store) FormatContextWithOptions(project, scope string, opts ContextOptions) (string, error) {
    // ... new implementation honoring opts ...
}
```

**Byte-compatibility proof:** a zero-value `ContextOptions{}` hits the same code paths as the old `FormatContext` body. The `Limit <= 0` branch falls back to `s.cfg.MaxContextResults` (exactly what the old code used). The `!opts.Compact` branch renders `- [%s] **%s**: %s\n` with `truncate(obs.Content, 300)` (exactly the old format string). No conditionals flip, no new allocations in the default path.

**Empirical verification:** I built both the old binary (`main`) and this branch, pointed them at the same 200+ observation database, and compared the responses:

```
old binary, /context                 → 7949 B, ~1.0 ms
new binary, /context (no params)     → 7949 B, ~0.9 ms   ← byte-for-byte identical
new binary, /context?limit=8&compact=1 → 728 B, ~0.8 ms   ← −91%
```

The default-params path is a **byte-identical drop-in** replacement for the old behavior.

## Query parameters

| param | values | effect |
|---|---|---|
| `limit` | positive integer | Cap the observations section at N bullets. `0`, negative, missing, or unparseable → falls back to `MaxContextResults` (default 20). Sessions (5) and prompts (10) are never affected. |
| `compact` | `strconv.ParseBool`: `1`, `0`, `t`, `f`, `T`, `F`, `true`, `false`, `True`, `False`, `TRUE`, `FALSE` | Render observation bullets as `- [type] **title**` only, dropping the `: <300 chars of body>` preview. Unknown values silently fall back to `false` (legacy behavior). |

**Caveat on `compact=`:** I originally documented \"`yes`\" as accepted because I assumed `ParseBool` followed common-sense rules. It doesn't — only the 12 values above. The second commit in this PR fixes the misleading inline comment. I considered accepting `yes`/`no` via a custom parser but decided consistency with the Go stdlib was more valuable than convenience, and silent fallback to `false` is safer than erroring out on unknown values (bad clients get the old behavior, not a 400).

## Hook integration (matched-version fast path + mixed-version fallback)

`plugin/claude-code/scripts/session-start.sh` is updated to pass `?limit=${ENGRAM_CONTEXT_LIMIT}&compact=1`. On a matched-version deployment (new plugin + new binary), the server does the compaction and the awk pass from #161 becomes a near no-op. On a mixed-version deployment (new plugin + old binary), the old binary silently ignores the unknown query params, returns the full response, and the awk fallback from #161 compacts it client-side. **Both cases produce a correctly compacted output.**

This is the main reason I kept the awk pass in the hook rather than removing it in this PR.

## Measurement methodology

- **Database:** my real working `engram.db` with 200+ observations across several months (`ctodie` project), all in `personal` scope. Mix of `session_summary` (multi-line markdown), `pattern`, `discovery`, `decision`, `manual`, `bug` types.
- **Setup:** built both binaries from the same git state (old = `main`, new = this branch), ran each in turn on port 7438 to avoid stepping on the live daemon's port. Hit `/context` 3 times per variant with `curl -w \"time_total=%{time_total}s size=%{size_download}\\n\"` to smooth out cold-start jitter.
- **Latencies are loopback.** Don't read too much into the micro-differences; the real signal is the payload size delta. The \"fast\" path is ~10% faster mainly because it serializes less data, not because of any algorithmic win.

| binary | request | payload | median latency |
|---|---|---:|---:|
| old (`main`) | `/context?project=ctodie` | 7949 B | ~1.0 ms |
| new (this PR) | `/context?project=ctodie` (default params) | **7949 B** | ~0.9 ms |
| new (this PR) | `/context?project=ctodie&limit=8&compact=1` | **728 B** (−91%) | ~0.8 ms |

End-to-end SessionStart hook injection, real project:

| | bytes |
|---|---:|
| original (before #161) | ~9800 |
| after #161 alone (shell awk) | ~1900 (−81%) |
| **after #161 + this PR (server-side)** | **~1151 (−88%)** |

## Risk analysis

**What could break?**

1. **Silent semantic drift in `FormatContext`.** Mitigated by `TestFormatContextWithOptions` asserting `defaultCtx == legacyCtx` as exact string equality. Any future refactor that touches `FormatContextWithOptions` without updating this test will flag a drift immediately.
2. **Query-param injection.** Both params are parsed with `strconv.Atoi` / `strconv.ParseBool` before any store interaction. Unparseable values silently fall back to defaults, never reach the store, and never affect the SQL query (the store-level `limit` passed to `RecentObservations` is either a validated positive int or the config default).
3. **Existing test fixtures breaking.** None did. `go test ./...` → 748 passed → 749 passed (one new test) → 750 passed (one more assertion in the e2e suite). All existing tests untouched.
4. **MCP clients expecting the old format.** MCP clients go through `internal/mcp/mcp.go` which still calls `FormatContext(project, scope)` — i.e. the backward-compat wrapper. Their output is unchanged byte-for-byte.

**What's the rollback?**

- `git revert <merge-commit>`. No database migration, no config file format change, no on-disk state change. The new `ContextOptions` struct and `FormatContextWithOptions` method disappear with the revert; nothing outside the store package depends on them.

## Tests

- **New:** `internal/store/store_test.go::TestFormatContextWithOptions`. Covers:
  - `ContextOptions{}` → byte-equal to `FormatContext` (backward compat)
  - `Compact: true` → drops content preview, keeps titles
  - `Limit: 2` → exactly 2 observation bullets (using `strings.Count`)
  - `Limit: 3, Compact: true` → both knobs compose correctly
  - `Limit: 0` → falls back to default (byte-equal to legacy)
- **Extended:** `internal/server/server_e2e_test.go`'s `/context` test now also hits `?limit=1&compact=1` and asserts the compact payload is strictly smaller than the default and contains no `: <body>` segments.
- **Full suite:** `go test ./...` → 748 passed in 10 packages (was 747 before the new test; a second e2e assertion brings it to 749 on this branch).

## Files changed (this PR's additions only, excluding #161)

```
internal/server/server.go                   | 25 ++++++--
internal/server/server_e2e_test.go          | 20 ++++++
internal/store/store.go                     | 43 ++++++++++++-
internal/store/store_test.go                | 96 +++++++++++++++++++++++++++++
plugin/claude-code/scripts/session-start.sh | 25 +++++---
5 files changed, 192 insertions(+), 17 deletions(-)
```

## Commits

1. `feat(context): add limit + compact query params to GET /context` — the main change
2. `docs(context): fix compact= param comment — ParseBool rejects 'yes'` — drive-by doc fix after I benchmarked the live binary and caught the inaccuracy

Happy to squash if you prefer a single commit.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -471,10 +471,24 @@ func (s *Server) handleImport(w http.ResponseWriter, r *http.Request) {
 // ─── Context ─────────────────────────────────────────────────────────────────
 
 func (s *Server) handleContext(w http.ResponseWriter, r *http.Request) {
-	project := r.URL.Query().Get("project")
-	scope := r.URL.Query().Get("scope")
+	q := r.URL.Query()
+	project := q.Get("project")
+	scope := q.Get("scope")
 
-	context, err := s.store.FormatContext(project, scope)
+	opts := store.ContextOptions{}
+	if raw := q.Get("limit"); raw != "" {
+		if n, err := strconv.Atoi(raw); err == nil && n > 0 {
+			opts.Limit = n
+		}
+	}
+	if raw := q.Get("compact"); raw != "" {
+		// Accept "1", "true", "yes" (case-insensitive); anything else is false.
+		if b, err := strconv.ParseBool(raw); err == nil {
+			opts.Compact = b
+		}
+	}
+
+	context, err := s.store.FormatContextWithOptions(project, scope, opts)
 	if err != nil {
 		jsonError(w, http.StatusInternalServerError, err.Error())
 		return
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -482,7 +482,10 @@ func (s *Server) handleContext(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 	if raw := q.Get("compact"); raw != "" {
-		// Accept "1", "true", "yes" (case-insensitive); anything else is false.
+		// Accepts any value strconv.ParseBool understands: "1", "t", "T",
+		// "TRUE", "true", "True" (and their false counterparts). Unknown
+		// values (e.g. "yes") silently leave Compact at its zero value so
+		// callers that expected full content don't get surprised.
 		if b, err := strconv.ParseBool(raw); err == nil {
 			opts.Compact = b
 		}
--- a/internal/server/server_e2e_test.go
+++ b/internal/server/server_e2e_test.go
@@ -382,6 +382,26 @@ func TestCoreReadHandlersAndHelpersE2E(t *testing.T) {
 		t.Fatalf("expected formatted context output")
 	}
 
+	// ── limit + compact query params (added in feat/context-limit-compact).
+	// compact=1 drops the ": <content>" preview from observation bullets,
+	// and limit=1 caps the observations section. The resulting payload must
+	// be strictly smaller than the default.
+	compactResp, err := client.Get(ts.URL + "/context?project=engram&scope=project&limit=1&compact=1")
+	if err != nil {
+		t.Fatalf("context compact: %v", err)
+	}
+	if compactResp.StatusCode != http.StatusOK {
+		t.Fatalf("expected 200 context compact, got %d", compactResp.StatusCode)
+	}
+	compactData := decodeJSON[map[string]string](t, compactResp)
+	if len(compactData["context"]) >= len(contextData["context"]) {
+		t.Fatalf("compact context (%d) should be smaller than default (%d)",
+			len(compactData["context"]), len(contextData["context"]))
+	}
+	if strings.Contains(compactData["context"], "- [") && strings.Contains(compactData["context"], "**: ") {
+		t.Fatalf("compact context should not include ': <body>' content preview, got:\n%s", compactData["context"])
+	}
+
 	statsResp, err := client.Get(ts.URL + "/stats")
 	if err != nil {
 		t.Fatalf("stats: %v", err)
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -1610,13 +1610,46 @@ func (s *Store) Stats() (*Stats, error) {
 
 // ─── Context Formatting ─────────────────────────────────────────────────────
 
+// ContextOptions tunes the output of FormatContextWithOptions.
+//
+// A zero-value ContextOptions (Limit==0, Compact==false) reproduces the
+// legacy FormatContext behaviour exactly, so it is safe to pass through
+// from handlers that do not care about the new knobs.
+type ContextOptions struct {
+	// Limit caps the number of observations rendered into the context
+	// string. Zero means "use the store-wide MaxContextResults default".
+	// Sessions and prompts are not affected — their counts are fixed at
+	// 5 and 10 respectively, matching the legacy behaviour.
+	Limit int
+
+	// Compact drops the inline content preview on each observation bullet,
+	// rendering `- [type] **title**` instead of `- [type] **title**: <300
+	// chars of body>`. On a busy project this cuts the observation section
+	// by ~80%. Sessions and prompts are not affected.
+	Compact bool
+}
+
+// FormatContext is a thin wrapper around FormatContextWithOptions that uses
+// default options, preserving the pre-ContextOptions call signature so
+// existing callers and tests keep working unchanged.
 func (s *Store) FormatContext(project, scope string) (string, error) {
+	return s.FormatContextWithOptions(project, scope, ContextOptions{})
+}
+
+// FormatContextWithOptions renders the "## Memory from Previous Sessions"
+// markdown block for the given project/scope, honoring the supplied
+// ContextOptions for observation limit and compact rendering.
+func (s *Store) FormatContextWithOptions(project, scope string, opts ContextOptions) (string, error) {
 	sessions, err := s.RecentSessions(project, 5)
 	if err != nil {
 		return "", err
 	}
 
-	observations, err := s.RecentObservations(project, scope, s.cfg.MaxContextResults)
+	obsLimit := opts.Limit
+	if obsLimit <= 0 {
+		obsLimit = s.cfg.MaxContextResults
+	}
+	observations, err := s.RecentObservations(project, scope, obsLimit)
 	if err != nil {
 		return "", err
 	}
@@ -1657,8 +1690,12 @@ func (s *Store) FormatContext(project, scope string) (string, error) {
 	if len(observations) > 0 {
 		b.WriteString("### Recent Observations\n")
 		for _, obs := range observations {
-			fmt.Fprintf(&b, "- [%s] **%s**: %s\n",
-				obs.Type, obs.Title, truncate(obs.Content, 300))
+			if opts.Compact {
+				fmt.Fprintf(&b, "- [%s] **%s**\n", obs.Type, obs.Title)
+			} else {
+				fmt.Fprintf(&b, "- [%s] **%s**: %s\n",
+					obs.Type, obs.Title, truncate(obs.Content, 300))
+			}
 		}
 		b.WriteString("\n")
 	}
--- a/internal/store/store_test.go
+++ b/internal/store/store_test.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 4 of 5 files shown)


Alan-TheGentleman (approved):
## Review — APPROVED

Excelente PR. Backward compatibility verificada al 100%, bien testeada, y la implementación es idiomática.

### Backward Compatibility ✓
- `FormatContext(project, scope)` sigue existiendo como wrapper de `FormatContextWithOptions` con zero-value options
- MCP clients siguen llamando `FormatContext()` directamente — output byte-por-byte idéntico
- `GET /context` sin query params → flujo idéntico al handler anterior
- El test de `defaultCtx == legacyCtx` como string equality es un guardrail excelente

### Compact mode ✓
- Elimina solo el preview de contenido, mantiene type + título + sesiones + prompts
- Suficiente para que un agente sepa QUÉ existe y haga `mem_get_observation` on-demand

### Edge cases ✓
- `limit=0`, `limit=-1`, `limit=abc` → todos caen al default correctamente
- Double safety net en `RecentObservations` — redundante pero seguro
- `limit` + `compact` combinados → testeado

### Observaciones menores (non-blocking)
1. **Test e2e de compact**: la assertion busca `**: ` como substring global — podría matchear un título de observation. Mejor verificar línea por línea.
2. **`compact=yes` cae silenciosamente a false**: `strconv.ParseBool` no acepta "yes"/"on". El fallback es safe (retorna full context) pero puede confundir.
3. **Squash commits** antes de merge — el commit 1 es de PR #161.

Fantástico laburo. El patrón de options struct + wrapper es consistente con `SearchOptions` del mismo codebase. Se puede mergear con confianza.

Alan-TheGentleman:

PR #192, opened 2026-04-16
fix(setup): repair Claude MCP tool discovery

### LINKED ISSUE #190: [Bug] MCP tools not found on session start causing ~5 minute timeout per occurrence
### 📋 Pre-flight Checks

### 📝 Bug Description

When starting a Claude Code session with engram enabled, Claude Code attempts to use engram's MCP tools (mem_search, mem_get_observation) before the MCP server has fully initialized. This causes a ~5 minute timeout per occurrence. The issue happens on the first message of a session and randomly on subsequent messages, suggesting a race condition between Claude Code's tool invocation and the engram MCP server readiness.

### 🔄 Steps to Reproduce

1. Install engram as part of the gentle-ai stack
2. Open a Claude Code session (claude --debug)
3. Send any prompt
4. Observe ~5 minute delay before Claude Code responds
5. Repeat — the delay may not occur on every message, but recurs unpredictably within the same session

### ✅ Expected Behavior

Claude Code should wait for the engram MCP server to be fully ready before attempting to invoke its tools, or gracefully retry with a short delay instead of timing out after 5 minutes.

### ❌ Actual Behavior

Error: Tool mcp__plugin_engram_engram__mem_search not found
Error: Tool mcp__plugin_engram_engram__mem_get_observation not found

Each error triggers a ~5 minute timeout before Claude Code continues. The issue occurs on the first message and randomly on later ones within the same session.
Debug log also shows a secondary issue:

RipgrepTimeoutError: Ripgrep search timed out after 20 seconds.

### Operating System

Windows

### Engram Version

1.12.0

### Agent / Client

Claude Code

### 📋 Relevant Logs

```shell
2026-04-15T11:19:45.390Z [ERROR] Error: Error: Tool mcp__plugin_engram_engram__mem_search not found
2026-04-15T11:19:45.391Z [ERROR] Error: Error: Tool mcp__plugin_engram_engram__mem_search not found
2026-04-15T11:19:45.392Z [ERROR] Error: Error: Tool mcp__plugin_engram_engram__mem_get_observation not found
2026-04-15T11:19:45.392Z [ERROR] Error: Error: Tool mcp__plugin_engram_engram__mem_get_observation not found
2026-04-15T11:19:45.463Z [ERROR] RipgrepTimeoutError: Ripgrep search timed out after 20 seconds.
```

### 💡 Additional Context

Running claude mcp list confirms engram is connected (✓ Connected), so the MCP server does start successfully — the issue is a timing/race condition where Claude Code invokes tools before the server is ready. The 5-minute wait matches the default MCP tool timeout. The problem also occurred on a colleague's machine on the same corporate network, suggesting it is not environment-specific.

jgarrone82:
Fixes #190

## Summary

This reworks the original readiness-delay approach into a deterministic Claude Code setup repair.

## Changes

- Removes the fixed MCP startup delay approach. No `--init-delay`, no `time.Sleep`, and no stderr readiness contract.
- Derives Claude Code MCP permission names from the actual `agent` tool profile via `mcp.ResolveTools("agent")`.
- Adds both current durable user-level server id permissions (`mcp__engram__...`) and legacy plugin-scoped permissions (`mcp__plugin_engram_engram__...`).
- Updates the Claude memory skill fallback so it no longer hardcodes stale plugin-scoped ToolSearch names.
- Updates setup docs to describe re-running `engram setup claude-code` as the repair path.

## Validation

- `go test ./cmd/engram ./internal/setup ./internal/mcp`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ For authenticated mode, upgrade flow, dashboard behavior, reason codes, and full
 |---------|-------------|
 | `engram setup [agent]` | Install agent integration |
 | `engram serve [port]` | Start HTTP API (default: 7437) |
-| `engram mcp` | Start MCP server (stdio) |
+| `engram mcp [--tools=PROFILE]` | Start MCP server (stdio transport) |
 | `engram tui` | Launch terminal UI |
 | `engram search <query>` | Search memories |
 | `engram save <title> <msg>` | Save a memory |
--- a/cmd/engram/main_extra_test.go
+++ b/cmd/engram/main_extra_test.go
@@ -3617,5 +3617,4 @@ func TestCmdMCP(t *testing.T) {
 		_, stderr, recovered := captureOutputAndRecover(t, func() { cmdMCP(cfg) })
 		assertFatal(t, stderr, recovered, "stdio failed")
 	})
-
 }
--- a/cmd/engram/main_extra_test.go
+++ b/cmd/engram/main_extra_test.go
@@ -3617,4 +3617,5 @@ func TestCmdMCP(t *testing.T) {
 		_, stderr, recovered := captureOutputAndRecover(t, func() { cmdMCP(cfg) })
 		assertFatal(t, stderr, recovered, "stdio failed")
 	})
+
 }
--- a/docs/AGENT-SETUP.md
+++ b/docs/AGENT-SETUP.md
@@ -97,7 +97,7 @@ That's it. The plugin registers the MCP server, hooks, and Memory Protocol skill
 engram setup claude-code
 ```
 
-During setup, you'll be asked whether to add engram tools to `~/.claude/settings.json` permissions allowlist — this prevents Claude Code from prompting for confirmation on every memory operation.
+During setup, you'll be asked whether to add engram's agent-profile MCP tools to `~/.claude/settings.json` `permissions.allow`. The setup writes entries for both the durable user-level MCP server id (`mcp__engram__...`) and the plugin-scoped server id used by older Claude Code plugin installs, so re-running setup repairs stale or incomplete allowlists without adding startup delay.
 
 **Option C: Bare MCP** — all 17 tools by default, no session management:
 
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -20,7 +20,10 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strings"
+
+	"github.com/Gentleman-Programming/engram/internal/mcp"
 )
 
 var (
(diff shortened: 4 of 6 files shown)


Alan-TheGentleman (changes requested):
Requesting changes because this PR is not merge-ready yet.

Blockers:

1. It does not compile
- The new readiness subtests in `cmd/engram/main_extra_test.go` declare `stdout` and do not use it.
- I applied the PR patch in a temp copy and ran `go test ./cmd/engram ./internal/mcp -count=1`.
- Result: `cmd/engram` fails to build, while `internal/mcp` passes.

2. `--init-delay=<duration>` skips the next flag
- In `cmd/engram/main.go`, the `--init-delay=` branch increments `i`.
- That is only valid for the separated form (`--init-delay 1s`), not the equals form.
- Example: `engram mcp --init-delay=1s --project=testproj` will skip `--project`. Same risk with `--tools`.

3. This is not yet a real readiness handshake
- `internal/mcp/mcp.go` shows tools are registered before `serveMCP()` is called.
- The current change adds a sleep plus a stderr log, but does not wire any supported client/setup path in this repo to wait for that signal.
- `internal/setup/setup.go` still generates `engram mcp --tools=agent` with no readiness consumer.

Suggested fixes:
- Remove the unused variables so the PR compiles.
- Remove the extra `i++` from the `--init-delay=<duration>` branch.
- Add a regression test combining `--init-delay` with `--project` / `--tools`.
- Either wire an actual consumer for the readiness signal or add a reproducible test proving the server-side-only change resolves issue #190.

jgarrone82:
Hi @reviewer, thanks for the detailed review. All four blockers have been addressed in the latest push (312c007). Here's how each point was resolved:
1. Compilation error — unused stdout variables
Fixed. Removed the unused stdout bindings from all three readiness subtests, replacing them with the discard pattern:
_, stderr, recovered := captureOutputAndRecover(t, func() { cmdMCP(cfg) })
Files: cmd/engram/main_extra_test.go (3 occurrences).
---
2. --init-delay=<duration> skips the next flag
Fixed. Removed the erroneous i++ from the --init-delay= (equals-form) branch. The increment is only valid for the separated form (--init-delay 1s), where the next element is the value. In the equals form, the value is self-contained, so incrementing i incorrectly consumed the following flag.
Before:
} else if strings.HasPrefix(os.Args[i], "--init-delay=") {
    if d, err := time.ParseDuration(...); err == nil {
        initDelay = d
    }
    i++ // ← removed
}
cmd/engram/main.go.
---
3. Regression test combining --init-delay with --project / --tools
Added. New subtest "--init-delay with equals does not skip subsequent flags" exercises the exact scenario you called out:
engram mcp --init-delay=1s --project=testproj --tools=agent
The test verifies that:
- newMCPServerWithConfig receives a non-nil allowlist (proving --tools=agent was parsed)
- The ready signal contains delay=1s (proving --init-delay=1s was parsed)
- The ready signal contains project="testproj" (proving --project=testproj was NOT skipped)
cmd/engram/main_extra_test.go.
---
4. Reproducible test proving the server-side change resolves issue #190
Added. New subtest "initialization delay precedes serving" uses time.Now() to measure that serveMCP is invoked after the delay has elapsed:
start := time.Now()
serveMCP = func(...) error {
    serveStart = time.Now() // record when serving actually begins
    return nil
}
// ... invoke cmdMCP with --init-delay=100ms ...
if serveStart.Before(start.Add(90 * time.Millisecond)) {
    t.Fatalf("serveMCP called too early")
}
This demonstrates the exact ordering required to resolve the race condition:
1. Server created, tools registered
2. time.Sleep(initDelay) executes
3. Ready signal emitted to stderr
4. Only then serveMCP starts accepting connections
Since the actual consumers (Claude Code plugin, OpenCode plugin, etc.) live outside this repo, this server-side test provides the reproducible proof that the delay correctly prevents clients from hitting tools before initialization is complete.
cmd/engram/main_extra_test.go.

Alan-TheGentleman:
I pushed a rework of this PR to avoid the fixed startup delay. The branch now repairs Claude Code MCP tool discovery directly: setup derives permissions from the actual `agent` tool profile, writes both current `mcp__engram__...` and legacy plugin-scoped permission names, and removes the stale hardcoded ToolSearch fallback from the Claude memory skill. I also rebased it onto `main`, updated the PR title/body, and verified `go test ./cmd/engram ./internal/setup ./internal/mcp` plus `git diff --check`. No checks are currently reported by GitHub for this branch, but it is now mergeable.

Alan-TheGentleman (approved):
Scoped, tested, linked to the approved issue. CI and PR validation are green. Approving.

PR #210, opened 2026-04-21
fix(store): escape forward slashes in FTS5 sanitize and normalize topic_key queries

### LINKED ISSUE #208: bug(store): FTS5 search fails with '/' in topic keys like 'sdd/summa-kit/spec'
## Problem

When searching for observations using a topic_key that contains a forward slash (e.g., \sdd/summa-kit/spec\), the FTS5 search path fails because \/\ is treated as an FTS5 query operator, causing a \&quot;malformed FTS5 query\&quot; error.

Additionally, the topic_key direct lookup path does not normalize the query, so raw topic keys stored with normalized format (e.g., \sdd-summa-kit-spec\) won't match queries using the human-readable format with slashes.

## Expected Behavior

\mem_search\ with query \sdd/summa-kit/spec\ should find observations with topic_key \sdd-summa-kit-spec\.

## Root Cause

1. \sanitizeFTS\ (store.go) does not escape forward slashes before passing to FTS5 MATCH
2. The topic_key direct lookup path (line ~1576) uses the raw query instead of normalizing it

## Proposed Fix

1. Escape \/\ to \\\/\ in \sanitizeFTS\
2. Apply \
ormalizeTopicKey()\ to the query before the topic_key direct lookup

jorgehara:
## Summary

- Escape `/` to `\/` in `sanitizeFTS` to prevent FTS5 "malformed query" errors when searching topic keys containing slashes (e.g., `sdd/summa-kit/spec`)
- Normalize query through `normalizeTopicKey()` before the topic_key direct lookup path so human-readable slash-separated queries match stored hyphen-separated topic keys

## Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | Escape `/` in sanitizeFTS + normalize topic_key queries |

## Test Plan

- [x] All existing store tests pass (`go test ./internal/store/...`)
- [x] Search tests specifically pass (`go test ./internal/store/... -run Search`)

## Checklist

- [x] Linked an approved issue (Closes #208)

### DIFF
--- a/cmd/engram/main.go
+++ b/cmd/engram/main.go
@@ -26,6 +26,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/Gentleman-Programming/engram/internal/agents"
 	"github.com/Gentleman-Programming/engram/internal/claudecode"
 	"github.com/Gentleman-Programming/engram/internal/mcp"
 	"github.com/Gentleman-Programming/engram/internal/obsidian"
@@ -178,6 +179,8 @@ func main() {
 		cmdClaudeCodeSync(cfg)
 	case "projects":
 		cmdProjects(cfg)
+	case "agents":
+		cmdAgents()
 	case "setup":
 		cmdSetup()
 	case "version", "--version", "-v":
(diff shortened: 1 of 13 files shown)


Alan-TheGentleman:

PR #254, opened 2026-04-26
docs: clarify install-from-source versioning and Claude Code requirement

### LINKED ISSUE #67: docs: clarify install-from-source version stamp and Claude Code CLI minimum version

idrojone:
Two documentation gaps addressed:

1. Clarified that go install sets version to dev. Provided the correct go build -ldflags command for Windows and macOS/Linux to get the proper version stamp.
2. Added a requirement note for Claude Code CLI >= 2.1.70 to support marketplace installation via git-subdir.

Closes #67

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,9 @@ brew install gentleman-programming/tap/engram
 
 Windows, Linux, and other install methods → [docs/INSTALLATION.md](docs/INSTALLATION.md)
 
+> [!TIP]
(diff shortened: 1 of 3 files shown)


Alan-TheGentleman (changes requested):
This has the right approved issue link, so the process is close. I am not going to merge it until the PR has `type:docs`, is rebased onto `main`, and all required checks pass. Update those pieces and I will re-review.

Alan-TheGentleman:
Closing as superseded by #410, a newer and cleaner take on the same install-from-source versioning docs for issue #67. Continuing there.

PR #255, opened 2026-04-26
fix(lifecycle): implement session end for OpenCode, Gemini CLI, and Codex

### LINKED ISSUE #48: Session end lifecycle missing for OpenCode, Gemini CLI, and Codex
## Context

PR #47 fixes session-stop for Claude Code by calling `POST /sessions/{id}/end` from the hook. But the other three agents have no equivalent — their sessions always have `ended_at = NULL`.

## Current state

| Agent | Session start | Session end |
|---|---|---|
| Claude Code | `session-start.sh` hook → `POST /sessions` | `session-stop.sh` hook → `POST /sessions/{id}/end` (PR #47) |
| OpenCode | `engram.ts` plugin → `POST /sessions` | Nothing |
| Gemini CLI | MCP `mem_session_start` via protocol | Nothing |
| Codex | MCP `mem_session_start` via protocol | Nothing |

## What's already available

The MCP tool `mem_session_end` exists and works. Agents just need to be told to call it.

## Proposed fix per agent

- **OpenCode**: Add a session-end event handler in `engram.ts` that calls `POST /sessions/{id}/end` (same pattern as the session creation handler)
- **Gemini CLI**: Add instruction in the system prompt telling the agent to call `mem_session_end` before closing
- **Codex**: Add instruction in the memory config telling the agent to call `mem_session_end` before closing

## Why it matters

Without `ended_at`, session duration is unknown, the session list grows unbounded with no way to tell which sessions are active vs finished, and any future cleanup/retention logic can't distinguish stale sessions from ongoing ones.

idrojone:
Implements session lifecycle termination for all supported agents to prevent orphaned sessions in the cloud dashboard.

- **OpenCode**: The plugin now calls POST /sessions/{id}/end directly when the session.deleted event fires.
- **Gemini CLI & Codex**: Added an explicit instruction to call mem_session_end in the Memory Protocol injected into their system prompts.
- **OpenCode (graceful)**: Also added the mem_session_end instruction to its MEMORY_INSTRUCTIONS.

Closes #48

### DIFF
--- a/internal/setup/plugins/opencode/engram.ts
+++ b/internal/setup/plugins/opencode/engram.ts
@@ -109,7 +109,9 @@ Before ending a session or saying "done" / "listo" / "that's it", you MUST:
 ## Relevant Files
 - path/to/file — [what it does or what changed]
 
-This is NOT optional. If you skip this, the next session starts blind.
+2. Call \`mem_session_end\` to formally close the session.
+
+This is NOT optional. If you skip this, the next session starts blind and leaves orphaned sessions.
 
 ### AFTER COMPACTION
 
@@ -313,6 +315,9 @@ export const Engram: Plugin = async (ctx) => {
         const info = (event.properties as any)?.info
         const sessionId = info?.id
         if (sessionId) {
+          // Tell Engram the session ended
+          await engramFetch(`/sessions/${sessionId}/end`, { method: "POST" })
+          
           toolCounts.delete(sessionId)
           knownSessions.delete(sessionId)
           subAgentSessions.delete(sessionId)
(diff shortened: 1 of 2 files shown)


Alan-TheGentleman (changes requested):
This has the right approved issue link, so the process is close. I am not going to merge it until the PR has exactly one `type:*` label, is rebased onto `main`, and all required checks pass. Update those pieces and I will re-review the code on its merits.

Alan-TheGentleman (changes requested):
Picking this back up. #48 is approved and I added the `type:bug` label, so the remaining blocker is mechanical: rebase onto `main` to clear the stale base and get CI green. Push the rebase and I will re-review and merge.

PR #266, opened 2026-04-27
feat(store): inject implicit graph-linking metadata via topic_key siblings

### LINKED ISSUE #243: feat(store): inject implicit graph-linking metadata in search results via topic_key siblings
### Pre-flight Checks

### Problem Description / Descripción del Problema

**English:** When `mem_search` returns results, each observation is isolated — the agent has no awareness of related observations that share the same `topic_key`. For example, if an agent finds 'JWT Middleware' (topic: `auth/jwt`), it doesn't know about 'Token Rotation Bug' or 'Session Fix' that share the same topic. This forces additional manual searches to build relational context.

**Español:** Cuando `mem_search` retorna resultados, cada observación está aislada — el agente no tiene awareness de observaciones relacionadas que comparten el mismo `topic_key`. Esto fuerza búsquedas manuales adicionales para construir contexto relacional.

### Proposed Solution / Solución Propuesta

**English:** Post-process search results in two layers:
1. **Store layer**: New `RelatedByTopicKeys()` function that queries siblings by `topic_key` using the existing `idx_obs_topic` composite index (efficient, no schema migration)
2. **MCP layer**: Inject related metadata as enriched text in `handleSearch` response:
   `Related (topic 'auth/jwt'): ID 4 (JWT Middleware), ID 12 (Token Rotation Bug)`

Limits: max 5 siblings per topic_key. Excludes soft-deleted observations. Filters by project + scope. No query executed when results have no topic_keys.

**Español:** Post-procesamiento en dos capas: (1) nueva función `RelatedByTopicKeys()` en store usando el índice compuesto `idx_obs_topic` existente, y (2) inyección de metadata relacional como texto enriquecido en `handleSearch`. Máximo 5 hermanos por topic_key, filtra por project+scope, excluye soft-deleted.

### Affected Area
Store (new query function), MCP Server (search response enrichment)

### Additional Context
- Leverages existing `idx_obs_topic(topic_key, project, scope, updated_at DESC)` index
- No schema migration needed
- New `RelatedObservation` struct: `{ID int64, Title string}`
- Separation of concerns: store returns pure data, handler serializes to text

Snakeblack:
## Summary / Resumen

Adds implicit graph-linking to `mem_search`: alongside FTS5 lexical results, the response now appends sibling observations that share `topic_key` with any returned hit. This gives agents the relational neighbourhood of a memory without an embedding model — staying true to Engram's *single binary, zero dependencies, local-first* philosophy.

### Store Layer
- `RelatedByTopicKeys(keys, project, scope, excludeIDs) (map[string][]RelatedObservation, error)`: returns up to 5 sibling rows per `topic_key`, filtered by `project + scope`, excluding the IDs already in the search hit set and rows where `deleted_at IS NOT NULL`. Uses the existing `idx_obs_topic(topic_key, project, scope, updated_at DESC)` composite index.
- `InsertObservationForTest(...)`: low-level test helper that bypasses `AddObservation`'s `topic_key` UPSERT path so tests can seed multiple sibling rows sharing `(topic_key, project, scope)`.

### MCP Layer
- `handleSearch` groups returned topic_keys by `(project, scope)` so the lookup respects the same partition the search ran in. After listing the primary results, it appends:
  ```
  ---
  Related Context (from graph):
  Related (topic "auth/jwt"): ID 4 (Title), ID 12 (Title)
  ```
- The existing direct `topic_key` route (Rank=-1000) is untouched.

### Tests
- Store (5): siblings returned, deleted excluded, project+scope respected, 5-per-key limit, empty-keys short-circuit.
- MCP (2): `Related Context` block formatted correctly when results carry `topic_key`; absent when no result has one.

## Migration impact
- No DB schema change. No new columns.
- Reuses existing `topic_key`, `project`, `scope` columns and `idx_obs_topic` index.
- Backwards compatible — only enriches the `mem_search` response text.

Closes #243


Alan-TheGentleman:

Issue #263, opened 2026-04-27
engram update with gentle-ai


QADovah:
### 📋 Pre-flight Checks

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/engram/issues) and this is not a duplicate
- [x] I understand this issue needs `status:approved` before a PR can be opened

### 📝 Bug Description

updated gentle-ai
with gentle-ai command.

After that.

Download file:
https://github.com/Gentleman-Programming/engram/releases/tag/v1.14.1 (windows amd)

Installed (cmd window open and closed).
Run engram tui from terminal.

Can't update because of this:
   Could not check for updates: GitHub API returned 403 Forbidden. Set GH_TOKEN or GITHUB_TOKEN to reduce rate limits.

- engram version 
returns
engram 1.12.1-0.20260412141513-64bf163d0b9b

### 🔄 Steps to Reproduce

1. 
2. 
3. 

### ✅ Expected Behavior

update engram version to 1.14.

### ❌ Actual Behavior

Could not check for updates: GitHub API returned 403 Forbidden. Set GH_TOKEN or GITHUB_TOKEN to reduce rate limits.

### Operating System

macOS

### Engram Version

1.14 installer

### Agent / Client

Claude Code

### 📋 Relevant Logs

```shell

```

Alan-TheGentleman:
Closing this. The failure is GitHub API rate-limiting on unauthenticated requests, which is environment-side rather than an engram bug. Set a `GH_TOKEN` in your environment and the update check will authenticate. Reopen if it still fails with a token set.

PR #273, opened 2026-04-28
fix(mcp): run checkForUpdates async for mcp and serve subcommands

### LINKED ISSUE #269: bug(mcp): checkForUpdates blocks MCP startup on every invocation
### 📋 Pre-flight Checks

### 📝 Bug Description

When `engram mcp` starts, it performs a synchronous HTTP request to the GitHub API (`checkForUpdates`) before initializing the MCP server. On Windows, the OS-level socket (`WSARecv`) can hang beyond Go's 2s context deadline, delaying or permanently blocking `cmdMCP` from starting. Since Claude Code has a stream idle timeout of `idleDeadlineMs=300000` (5 minutes), a hung update check means the MCP server never responds and Claude Code waits the full 5 minutes before giving up.

### 🔄 Steps to Reproduce

1. Install engram and configure it as a Claude Code MCP plugin on Windows
2. Start Claude Code with `claude --debug`
3. Send any prompt that triggers an MCP tool call
4. Observe a delay of several seconds to minutes before MCP responds
5. In severe cases (firewall silently dropping GitHub API traffic), the MCP server never starts and Claude Code times out after 5 minutes

### ✅ Expected Behavior

The MCP server should start listening immediately, without waiting for a network request to complete. Update availability notifications are informational and should not block server initialization.

### ❌ Actual Behavior

The MCP server startup is gated behind a GitHub API call. If the call hangs at the OS level (beyond the Go context timeout), `cmdMCP` is never reached. Claude Code's stream idle timeout fires after 5 minutes.

Debug logs show MCP startup variance:
- Normal sessions: ~1775ms
- Affected sessions: ~6417ms (and in worst cases, full 5-minute timeout)

### Operating System

Windows 11

### Engram Version

1.14.5

### Agent / Client

Claude Code

### 📋 Relevant Logs

```shell
# Normal session — MCP starts in ~1775ms
# Affected session — MCP starts in ~6417ms or never responds

# Claude Code stream idle timeout fires:
# idleDeadlineMs=300000 → stream killed after 5 minutes of no MCP response
```

### 💡 Additional Context

- `checkForUpdates` uses `http.DefaultClient` with only a `context.WithTimeout(2s)`. On Windows, Go's context cancellation does not always force-close the underlying OS socket synchronously.
- The fix is to run `checkForUpdates` in a goroutine for `mcp` and `serve` subcommands, so the server starts immediately regardless of network conditions.
- Related: #263 (GitHub API 403 during update check — same call, different symptom).
- PR #273 was opened ahead of approval as part of a debugging investigation; please review issue first.

jgarrone82:
## Problem

`checkForUpdates` runs **synchronously** in `main()` before `cmdMCP` is called (`cmd/engram/main.go:534`). This means the MCP server does not start listening until the GitHub API call completes or times out.

```go
// main.go:534 — runs BEFORE cmdMCP
if result := checkForUpdates(version); ... {
    fmt.Fprintln(os.Stderr, result.Message)
}
cmdMCP(cfg) // ← never reached until the HTTP call above finishes
```

## Root cause

- Uses `http.DefaultClient` with no transport-level timeout (only a 2s `context.WithTimeout`)
- On Windows, DNS resolution and TCP operations can hang at the OS level beyond the context deadline — Go cancels the context but the underlying `WSARecv` call may not unblock immediately
- Claude Code has a stream idle timeout of `idleDeadlineMs=300000` (5 minutes). If MCP never starts responding, Claude Code waits the full 5 minutes then kills the connection

## Observed impact

- Measured MCP startup of **6417ms** vs **1775ms** baseline in debug logs (Windows 11)
- Suspected root cause of intermittent **5-minute hangs** reported by users on Windows

## Fix

Run the update check in a goroutine so `cmdMCP` starts immediately:

```go
go func() {
    if result := checkForUpdates(version); result.Status != versioncheck.StatusUpToDate && result.Message != "" {
        fmt.Fprintln(os.Stderr, result.Message)
    }
}()
cmdMCP(cfg)
```

Interactive subcommands (`save`, `search`, `tui`, etc.) keep the synchronous check so the update message appears before command output.

## Environment

- Windows 11, Claude Code with Engram MCP plugin
- Related: #263 (GitHub API 403 — same `checkForUpdates` call)

Closes #269

### DIFF
--- a/cmd/engram/main.go
+++ b/cmd/engram/main.go
@@ -531,9 +531,18 @@ func main() {
 	}
 
 	// Check for updates on every invocation.
-	if result := checkForUpdates(version); result.Status != versioncheck.StatusUpToDate && result.Message != "" {
-		fmt.Fprintln(os.Stderr, result.Message)
-		fmt.Fprintln(os.Stderr)
+	// For long-running server subcommands (mcp, serve) run asynchronously so the
+	// server starts immediately without blocking on a network request.
+	printUpdate := func() {
+		if result := checkForUpdates(version); result.Status != versioncheck.StatusUpToDate && result.Message != "" {
+			fmt.Fprintln(os.Stderr, result.Message)
+			fmt.Fprintln(os.Stderr)
+		}
+	}
+	if os.Args[1] == "mcp" || os.Args[1] == "serve" {
+		go printUpdate()
+	} else {
+		printUpdate()
 	}
 
 	cfg, cfgErr := store.DefaultConfig()
--- a/cmd/engram/main_test.go
+++ b/cmd/engram/main_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/Gentleman-Programming/engram/internal/mcp"
 	"github.com/Gentleman-Programming/engram/internal/obsidian"
 	"github.com/Gentleman-Programming/engram/internal/setup"
+	engramsrv "github.com/Gentleman-Programming/engram/internal/server"
 	"github.com/Gentleman-Programming/engram/internal/store"
 	engramsync "github.com/Gentleman-Programming/engram/internal/sync"
 	versioncheck "github.com/Gentleman-Programming/engram/internal/version"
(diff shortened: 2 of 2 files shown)


Alan-TheGentleman:
Thanks for chasing #269. The direction is right, but I am closing this PR rather than merging the stale branch as-is because I am rebuilding the startup fix on current `main` with the rest of the issue batch. I will reference this PR in the replacement work so the original context is not lost. Closing the PR, not rejecting the underlying fix.

PR #281, opened 2026-04-29
feat(claude-code-plugin): honor ENGRAM_URL env var in hooks

johntrandall:
## Summary

Restore `ENGRAM_URL` env var override in the Claude Code plugin hooks. Default behavior unchanged.

## Motivation

Some users run engram as a centralized server (Synology, k8s, dev container) shared by multiple Macs / containers. Currently the hooks hardcode `127.0.0.1:7437` so these users have to fork and patch on every release.

This change:
1. Adds `${ENGRAM_URL:-http://127.0.0.1:${ENGRAM_PORT}}` fallback — same default
2. Skips the auto-spawn `engram serve` block in `session-start.sh` when `ENGRAM_URL` was overridden (the user is pointing at a remote server, not a local binary, so spawning a local `engram serve` would be wrong)

## Files changed

- `plugin/claude-code/scripts/session-start.sh` — env override + skip auto-spawn when overridden
- `plugin/claude-code/scripts/post-compaction.sh` — env override
- `plugin/claude-code/scripts/user-prompt-submit.sh` — env override
- `plugin/claude-code/scripts/session-stop.sh` — env override
- `plugin/claude-code/scripts/subagent-stop.sh` — env override
- `docs/INSTALLATION.md` — document `ENGRAM_URL`
- `DOCS.md` — document `ENGRAM_URL`

Total: 7 files, +17 / -9.

## Test plan

- [ ] Default install (no `ENGRAM_URL`): hooks behave identically to before — `session-start.sh` still auto-spawns `engram serve` if localhost is unreachable
- [ ] `ENGRAM_URL=http://192.168.1.10:7437 claude`: hooks talk to remote server; no local `engram serve` spawn attempted
- [ ] `ENGRAM_URL` empty / unset: same as default

## Compatibility

Backward compatible. Existing users who don't set `ENGRAM_URL` see no change.

## Out of scope

The OpenCode plugin (`plugin/opencode/engram.ts`) has the same hardcoded URL pattern. Happy to do a follow-up PR for that if this approach is accepted.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -200,6 +200,7 @@ Engram is local-first: local SQLite is authoritative; cloud features are optiona
 |---|---|---|
 | `ENGRAM_DATA_DIR` | Override data directory | `~/.engram` |
 | `ENGRAM_PORT` | Override HTTP server port | `7437` |
+| `ENGRAM_URL` | Override the URL Claude Code hooks talk to (multi-machine / remote backend setups). When set to a non-default value, the hooks will not auto-spawn a local `engram serve`. | `http://127.0.0.1:${ENGRAM_PORT}` |
 | `ENGRAM_PROJECT` | Override project name for MCP server | auto-detected via git |
 
 ### Cloud CLI (opt-in)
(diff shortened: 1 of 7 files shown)


Alan-TheGentleman:

PR #284, opened 2026-04-29
fix: resolve multi-repo cwd ambiguity + 3 related bugs (Closes #283)

### LINKED ISSUE #283: MCP server fails on multi-repo cwd + 3 related bugs (ambiguity, path-as-name, case-sensitivity, unsafe consolidate)
## Summary

Engram MCP server (`engram mcp`) keeps creating ambiguity errors and garbage projects when the cwd has multiple `.git` children, and several related issues compound the experience. After running for a few months on Windows my user store accumulated **9 garbage projects with absolute paths as project names** plus several unintended duplicates, while every `mem_save` from the MCP returned `ambiguous_project`.

This issue bundles 4 related bugs found while debugging on `engram 1.14.12` (Windows 11, MCP integrated with Claude Code).

---

## Bug 1 — `mem_save` fails with `ambiguous_project` when cwd has multiple `.git` children

**Severity**: High — blocks `mem_save` from MCP entirely whenever the user's home (or any parent directory) contains multiple repos.

**Repro**:
1. Have multiple git repos directly under `$HOME` (very common on Windows: e.g. `~/engram`, `~/agent-teams-lite`, `~/dotfiles`).
2. Start Engram MCP server with default args (no `--project`).
3. Call `mem_save` with title + content (no project param — schema doesn't accept one anyway).

**Expected**: save succeeds, project resolved to `general` (or last-used, or `--project` default).

**Actual**:
```json
{
  "available_projects": ["agent-teams-lite", "dotfiles", "engram", ...],
  "error_code": "ambiguous_project",
  "hint": "Use mem_current_project to inspect detection results, or cd into one of the listed repositories.",
  "message": "Cannot determine project: ambiguous project: multiple git repos found in cwd"
}
```

**Why it's blocking**: when Engram MCP is launched by Claude Code / OpenCode, the server's cwd is fixed (typically the user's home directory). The user cannot `cd` into a specific repo from inside the agent — the MCP process inherits the parent agent's cwd. So the suggested workaround in `hint` is impossible to apply.

The `mem_save` tool schema also doesn't accept `project` as an argument, so even passing it explicitly doesn't help.

**Suggested fix**: when ambiguity is detected, fall back to (in order):
1. The `--project NAME` flag if provided to the MCP server.
2. The most recently used project (last `last_activity_at`).
3. `general`.

Don't error out — the server has the info to make a sensible default.

---

## Bug 2 — When auto-detection fails, projects get created with the absolute path as the project name

**Severity**: Medium — produces garbage data over time. Workaround = manual SQL cleanup.

After several months of usage I had **9 garbage project entries** in my Engram store, all with Windows-style absolute paths as the project name and **0 observations** but lots of empty session rows. Sample shape (paths anonymized):

| Project name pattern | obs | sessions |
|---|---|---|
| `C:\Users\<user>` | 0 | 296 |
| `C:\Users\<user>\.opencode` | 0 | 53 |
| `C:\<workspace-A>` | 0 | 50 |
| `C:\Users\<user>\source\repos\<repo>` | 0 | 11 |
| `D:\<workspace-B>\<sub>\<sub>` | 0 | 17 |
| ...and 4 more similar | 0 | varied |

These were created in earlier 1.14.x versions where the MCP fell back to "use cwd as the project name" when detection failed. Newer versions error out instead (Bug 1), but the garbage from old versions remains.

**Cleanup query** (worked safely — 434 empty sessions deleted, 0 observations lost):
```sql
DELETE FROM sessions WHERE project LIKE 'C:\\%' AND project IN (
  SELECT project FROM sessions GROUP BY project HAVING COUNT(*) > 0
  AND project NOT IN (SELECT DISTINCT project FROM observations WHERE deleted_at IS NULL)
);
```

**Suggested fix**: validate project names. Reject any name that:
- Contains `\` or `/` (filesystem separators) — these are paths, not project names
- Is an absolute path pattern (`^[A-Z]:` on Windows, `^/` on Unix)

Fall back to `general` instead of using the path verbatim.

Optionally: `engram doctor cleanup-paths` command that detects and deletes these in one shot.

---

## Bug 3 — `mem_search` filter `project: "E3"` is lowercased internally and not found

**Severity**: Medium — case sensitivity inconsistency.

**Repro**:
1. Have a project saved with name `E3` (uppercase).
2. Call `mem_search(query: "...", project: "E3")` from MCP.

**Expected**: returns observations under project `E3`.

**Actual**:
```json
{
  "available_projects": [..., "E3", ...],
  "error_code": "unknown_project",
  "hint": "Use one of the available_projects values, or omit project to auto-detect.",
  "message": "Project \"e3\" not found in store"
}
```

The error message shows the input was lowercased to `"e3"` before lookup, even though `available_projects` lists `E3` (uppercase) in the same response.

**Suggested fix**: case-insensitive lookup against the `available_projects` set, OR preserve case as-is. Currently behavior is inconsistent (server preserves case in storage but downcases in input filter).

This bug is also present in the SQLite backend if you query directly: `SELECT * FROM observations WHERE project = 'e3'` returns nothing (case-sensitive collation), but the user filter that arrives via MCP transforms the input — so it's specifically an input normalization issue at the MCP boundary.

---

## Bug 4 — `engram projects consolidate --all` fuzzy-matching groups unrelated projects

**Severity**: High — runs the risk of catastrophic data merge.

**Repro**:
```bash
engram projects consolidate --all --dry-run
```

**Expected**: groups of projects with truly similar names (typos, casing variants, kebab vs snake case).

**Actual**: returns 4 groups, but **Group 2 contains 46 completely unrelated projects** with no shared substring or naming pattern, and proposes merging all of them into `general`. Sample shape (project names anonymized — what matters is the count and that they are clearly distinct projects):

```
Group 2:
    [1]  <internal-tool-A>                47 obs
    [2]  <absolute-path-B>                 0 obs
    [10] <product-C>                       1 obs
    [11] <product-C-variant>               1 obs
    [12] <presentation-D>                 13 obs
    [14] <mcp-E>                           4 obs
    [16] <plugin-F>                        6 obs
    [17] <daemon-G>                        1 obs
    [22] <training-H>                     10 obs
    [23] <runtime-I>                      12 obs
    [24] <plugin-J>                       90 obs
  → [27] general                         262 obs
    [29] <industrial-K>                   46 obs
    [35] <agent-L>                       119 obs
    [38] <user-home>                     229 obs
    [40] <opencode-plugin-M>              69 obs
    ...46 entries total — every one of these is a distinct project...
  Suggested canonical: "general"
  [dry-run] Would merge into "general"
```

These are **fully distinct projects** with their own observation history (4 to 229 obs each). If a user runs `engram projects consolidate --all` (without `--dry-run`) trusting the algorithm, they would lose project separation across **most of their store**.

**Suggested fix**: tighten the similarity threshold significantly. A reasonable heuristic:
- Levenshtein distance ≤ 3 on the canonical (lowercased, dehyphenated) form, OR
- Substring containment with both strings >50% of each other, OR
- Only group projects that share a normalized prefix of ≥ 4 characters.

The current algorithm seems to use something much looser (maybe just "ends with vowel-consonant" or "shares 1+ token") which is far too permissive.

For reference, **Group 1** in the same report is correct (`course-branding` ↔ `course-presentation`), so the algorithm has the right shape, just calibrated way too loose for Group 2.

**Suggested additional safeguard**: add an interactive `--confirm-each` flag that requires explicit y/N per group. The current `--dry-run` is good but a user who skips dry-run could lose data instantly.

---

## Environment

- OS: Windows 11 Pro 10.0.22631
- Engram: 1.14.12 (also reproduced on 1.14.4)
- Shell: Git Bash + PowerShell 7
- Integration: Claude Code (MCP via `~/.claude/mcp/engram.json`)
- Multi-repo home directory: 3+ git repos directly under `$HOME` (and growing)

## Impact summary

These 4 bugs together produce a degrading user experience over time:
1. Bug 1 prevents normal `mem_save` from MCP → users do SQL workarounds → data ends up inconsistently saved
2. Bug 2 silently accumulates garbage entries that pollute the project list
3. Bug 3 surfaces as confusing "Project not found" when the project clearly exists
4. Bug 4 turns a "cleanup" tool into a data-loss footgun

I'd be glad to provide more diagnostics or test fixes — happy to PR the validation logic for Bug 2 if you'd like.

Thanks for the great tool 🙏

Maicololiveras:
## Summary

Fixes the 4 related bugs reported in #283:

1. **`mem_save` blocks with `ambiguous_project`** when cwd has multiple `.git` children (very common when MCP server starts from `$HOME`).
2. **Path-like project names** can slip into the store (legacy `C:\Users\…` artifacts).
3. **`mem_search` with `project="E3"`** reports "Project 'e3' not found" even though the store has a row under `E3` (case-sensitivity inconsistency for legacy data).
4. **`engram projects consolidate --all`** groups dozens of unrelated projects into a single component because every project transitively shares `$HOME` as a "directory fingerprint" → unsafe merge proposed.

Closes #283

### Bug #1 — `mem_save` ambiguity fallback

- **New flag** `engram mcp --default-project=NAME` configures a project to fall back to when cwd-based detection returns `ErrAmbiguousProject`.
- New package-level setter `mcp.SetDefaultProjectFallback(name)` in `internal/mcp/mcp.go`. The fallback is **opt-in**: empty string preserves the historical "fail fast" contract.
- When the fallback fires, the response envelope keeps the original `available_projects` and adds `Source: "default_fallback"` plus a `Warning` so callers can see what happened.
- Help text updated.

**Files**: `internal/mcp/mcp.go`, `cmd/engram/main.go` (flag parsing + help text).

### Bug #2 — Reject path-like project names

- `NormalizeProject` now detects filesystem paths (Windows drive paths, UNC, Unix absolute, anything with `\` or `/`) and reduces them to the basename with a warning.
- Pure helper `sanitizePathLike` extracted for testability.
- Edge case: a drive root (`C:\`) yields empty basename → recursion bottoms out at `"unknown"` instead of leaving an empty project name.

**Files**: `internal/store/store.go`.

### Bug #3 — Case-insensitive project resolution

- New `Store.ResolveProjectName(input)` method: tries case-sensitive match first (fast path), falls back to `COLLATE NOCASE` and returns the canonical stored casing. Distinct casings (e.g. both `E3` and `e3` in store) remain individually addressable — exact-case match wins.
- `resolveReadProject` in `internal/mcp/mcp.go` now uses `ResolveProjectName` instead of `ProjectExists`, with two attempts: normalized form first, then the raw user override (in case the user passed exact stored casing that survives normalization).
- The returned `DetectionResult.Project` is the canonical stored name, not the normalized input — so downstream queries match exact rows.

**Files**: `internal/store/store.go`, `internal/mcp/mcp.go`.

### Bug #4 — Tighten `consolidate` similarity grouping

- `groupSimilarProjects` is now a thin wrapper over `groupSimilarProjectsWithMode(projects, strictSimilarity bool)`.
- New constant `maxSharedProjectsForDirMatch = 3`: when a directory is touched by more than 3 distinct projects, it is treated as too noisy to be a fingerprint and is **skipped** in the union-find. This is what eliminates the catastrophic transitive cluster where every project sharing `$HOME` got merged into one component.
- New flag **`engram projects consolidate --strict-similarity`**: disables shared-directory union entirely. Only name-similarity edges are formed. Recommended for stores where the user knows many projects share noisy parent paths.
- Help text updated.

**Files**: `cmd/engram/main.go`.

## Tests added

- `TestNormalizeProject_PathLike` (8 cases — Windows drive paths, UNC, Unix absolute, relative paths with separators, drive root edge case, plain-name passthrough).
- `TestResolveProjectName_CaseInsensitive` (legacy `E3` row inserted directly via SQL, covers exact match, lowercase fallback, unknown, empty).
- `TestResolveProjectName_PrefersExactCaseWhenBothExist` (asserts addressability of distinct casings).
- `TestResolveWriteProject_AmbiguousFallback` (fallback to `general` when `--default-project` is set, asserts envelope fields).
- `TestSetDefaultProjectFallback_NormalizesAndTrims` (confirms input normalization).
- `TestGroupSimilarProjects_NoisyDirectoryNotMerged` (5 unrelated projects sharing `$HOME` → 0 groups).
- `TestGroupSimilarProjects_LegitimateRenameStillMerges` (2 projects with unique shared dir → 1 group, canonical chosen by obs count).
- `TestGroupSimilarProjectsWithMode_StrictDisablesDirUnion` (strict mode skips dir union even for legitimate renames).

`TestResolveWriteProject_AmbiguousError` was updated to explicitly clear the fallback so it continues to assert the historical error path.

## Backward compatibility

- **No breaking changes**. All new behavior is opt-in:
  - `--default-project` is empty by default → ambiguous cwd still errors as before.
  - `--strict-similarity` is opt-in → consolidate behavior unchanged for users who do not pass it.
  - Path sanitization in `NormalizeProject` only activates when the input contains path separators — names without separators (the overwhelming majority of inputs) are unchanged.
  - `ResolveProjectName` is a new method; existing `ProjectExists` is untouched. The MCP layer now prefers the new method but case-sensitive matches still work identically.

## Test plan

- [x] `go build ./...` — clean.
- [x] `go test ./internal/project/ ./internal/store/ ./internal/mcp/ ./cmd/engram/ -count=1` — all passing locally (existing + new tests).
- [ ] Maintainer to confirm CI passes.

## Notes

- Issue #283 is currently `status:needs-review`. This PR is opened ahead of approval as a courtesy starting point so the maintainer can review the proposed shape of the fixes alongside the bug report. Happy to revise scope, split into smaller PRs, or hold until approval — whatever works best for the project.
- Two of the bugs (#1, #4) materially reduce risk for users with multi-repo home directories or large project stores accumulated over months. The fixes are conservative (opt-in flags + numeric thresholds) so the blast radius is small even if a behavior nuance was missed.

🙏 Thanks for the great tool.

### DIFF
--- a/cmd/engram/main.go
+++ b/cmd/engram/main.go
@@ -806,13 +806,27 @@ func tryStartAutosync(ctx context.Context, s *store.Store, cfg store.Config) (au
 }
 
 func cmdMCP(cfg store.Config) {
-	// Parse --tools flag. Project is always auto-detected from cwd at call time (JR2-4).
+	// Parse flags:
+	//   --tools=PROFILE         (filter exposed tools)
+	//   --default-project NAME  (fallback project name when cwd is ambiguous)
+	// Project is always auto-detected from cwd at call time (JR2-4); the
+	// --default-project flag only kicks in when detection returns
+	// ErrAmbiguousProject (cwd has multiple git repos as children — common on
+	// home directories for MCP clients like Claude Code / OpenCode).
 	toolsFilter := ""
+	defaultProject := ""
 	for i := 2; i < len(os.Args); i++ {
-		if strings.HasPrefix(os.Args[i], "--tools=") {
-			toolsFilter = strings.TrimPrefix(os.Args[i], "--tools=")
-		} else if os.Args[i] == "--tools" && i+1 < len(os.Args) {
+		arg := os.Args[i]
+		switch {
+		case strings.HasPrefix(arg, "--tools="):
+			toolsFilter = strings.TrimPrefix(arg, "--tools=")
+		case arg == "--tools" && i+1 < len(os.Args):
 			toolsFilter = os.Args[i+1]
+			i++
+		case strings.HasPrefix(arg, "--default-project="):
+			defaultProject = strings.TrimPrefix(arg, "--default-project=")
+		case arg == "--default-project" && i+1 < len(os.Args):
+			defaultProject = os.Args[i+1]
 			i++
 		}
 	}
@@ -822,6 +836,10 @@ func cmdMCP(cfg store.Config) {
 		fatal(err)
 	}
 	defer s.Close()
+
+	if defaultProject != "" {
+		mcp.SetDefaultProjectFallback(defaultProject)
+	}
 
 	mcpCfg := mcp.MCPConfig{}
 	allowlist := resolveMCPTools(toolsFilter)
@@ -1667,7 +1685,25 @@ type projectGroup struct {
 
 // groupSimilarProjects groups projects by name similarity and shared directories.
 // Uses a simple union-find approach.
+// maxSharedProjectsForDirMatch caps how many distinct projects may share a
+// single directory before that directory is treated as too "noisy" to be a
+// meaningful fingerprint. Without this cap, a directory like the user's home
+// (which dozens of unrelated projects' sessions transit through) would
+// transitively unite every project into a single component via union-find.
+//
+// 3 is a deliberately conservative number: a real rename or move shows up as
+// 1–2 project rows pointing to the same directory, never more.
+const maxSharedProjectsForDirMatch = 3
+
 func groupSimilarProjects(projects []store.ProjectStats) []projectGroup {
+	return groupSimilarProjectsWithMode(projects, false)
+}
+
+// groupSimilarProjectsWithMode groups projects with optional strict-similarity
+// behavior. When strictSimilarity=true, only name-based similarity is used —
+// shared-directory union is skipped entirely. This is safer for stores where
+// many projects share noisy parent directories (e.g. $HOME).
+func groupSimilarProjectsWithMode(projects []store.ProjectStats, strictSimilarity bool) []projectGroup {
 	n := len(projects)
 	if n == 0 {
 		return nil
@@ -1711,18 +1747,38 @@ func groupSimilarProjects(projects []store.ProjectStats) []projectGroup {
 		}
 	}
 
-	// Group by shared directory
-	dirToProjects := make(map[string][]int)
-	for i, p := range projects {
-		for _, dir := range p.Directories {
-			if dir != "" {
-				dirToProjects[dir] = append(dirToProjects[dir], i)
+	// Group by shared directory — opt-out via strictSimilarity, and also
+	// skip directories that are touched by more than maxSharedProjectsForDirMatch
+	// distinct projects. A directory shared by 5+ projects is almost always
+	// $HOME or some other non-fingerprint location, not a rename signal — and
+	// blindly unioning over it produces the catastrophic transitive cluster
+	// where, for example, gis ↔ jarvis ↔ forge get merged into "general".
+	if !strictSimilarity {
+		dirToProjects := make(map[string][]int)
+		for i, p := range projects {
+			for _, dir := range p.Directories {
+				if dir != "" {
+					dirToProjects[dir] = append(dirToProjects[dir], i)
+				}
 			}
 		}
-	}
-	for _, idxs := range dirToProjects {
-		for k := 1; k < len(idxs); k++ {
-			union(idxs[0], idxs[k])
+		for _, idxs := range dirToProjects {
+			// Distinct count (a project may appear multiple times in idxs if it
+			// recorded the same directory under multiple sessions).
+			seen := make(map[int]bool, len(idxs))
+			distinct := 0
+			for _, idx := range idxs {
+				if !seen[idx] {
+					seen[idx] = true
+					distinct++
+				}
+			}
+			if distinct > maxSharedProjectsForDirMatch {
+				continue // noisy parent path — skip to avoid transitive cluster
+			}
+			for k := 1; k < len(idxs); k++ {
+				union(idxs[0], idxs[k])
+			}
 		}
 	}
 
@@ -1766,12 +1822,15 @@ func groupSimilarProjects(projects []store.ProjectStats) []projectGroup {
 func cmdProjectsConsolidate(cfg store.Config) {
 	doAll := false
 	dryRun := false
+	strictSimilarity := false
 	for i := 3; i < len(os.Args); i++ {
 		switch os.Args[i] {
 		case "--all":
 			doAll = true
 		case "--dry-run":
 			dryRun = true
+		case "--strict-similarity":
+			strictSimilarity = true
 		}
 	}
 
@@ -1917,7 +1976,7 @@ func cmdProjectsConsolidate(cfg store.Config) {
 		fatal(err)
 	}
 
-	groups := groupSimilarProjects(projects)
+	groups := groupSimilarProjectsWithMode(projects, strictSimilarity)
 
 	if len(groups) == 0 {
 		fmt.Println("No similar project name groups found.")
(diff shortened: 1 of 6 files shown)


Maicololiveras:
**CI environment note for the maintainer reviewing this PR:**

Confirmed locally that `cmd/engram` test suite has a **pre-existing 10-minute hang on Windows** that is independent of this PR. Verified by:

1. Checking out `origin/main` (commit `5f3329f`, no PR changes applied) and running `go test ./cmd/engram/ -timeout=120s -count=1`.
2. Test suite hung at the timeout. Stack trace shows goroutines blocked on `chan receive` for >1 minute, all rooted at `cmd/engram/main.go:734` → `cmdServe.func1()` (line 735):

```
goroutine 1261 [chan receive, 1 minutes]:
github.com/Gentleman-Programming/engram/cmd/engram.cmdServe.func1()
	C:/Users/maicolj/engram/cmd/engram/main.go:735 +0x2f
created by github.com/Gentleman-Programming/engram/cmd/engram.cmdServe in goroutine 1268
	C:/Users/maicolj/engram/cmd/engram/main.go:734 +0x436
```

The test packages exercised by this PR (`internal/project`, `internal/store`, `internal/mcp`) all pass cleanly:

```
ok  	github.com/Gentleman-Programming/engram/internal/project	5.459s
ok  	github.com/Gentleman-Programming/engram/internal/mcp	19.327s
```

(`internal/store` reports two pre-existing Windows TempDir/SQLite-locking flakes unrelated to this PR's changes — both functions create their own temp DBs so it's likely the same Windows file-locking issue around test cleanup.)

The `cmd/engram` tests added by this PR (`TestGroupSimilarProjects_*`) pass when run in isolation:

```
go test ./cmd/engram/ -run "TestGroupSimilarProjects" -count=1 -v
=== RUN   TestGroupSimilarProjects_NoisyDirectoryNotMerged
--- PASS: TestGroupSimilarProjects_NoisyDirectoryNotMerged (0.00s)
=== RUN   TestGroupSimilarProjects_LegitimateRenameStillMerges
--- PASS: TestGroupSimilarProjects_LegitimateRenameStillMerges (0.00s)
=== RUN   TestGroupSimilarProjectsWithMode_StrictDisablesDirUnion
--- PASS: TestGroupSimilarProjectsWithMode_StrictDisablesDirUnion (0.00s)
```

CI on Linux likely won't hit the same hang — but flagging it in case it surfaces and the reviewer wonders if the PR introduced it. Happy to dig into the cmdServe goroutine leak as a follow-up issue if it'd be useful.

Thanks again 🙏

Alan-TheGentleman:

PR #289, opened 2026-04-30
feat(mcp): honor directory arg on remaining write tools

johntrandall:
## Summary

Extends the `directory`-parameter pattern that landed for `mem_session_start` in commits e092095 and 5f3329f (2026-04-29) to the remaining six write tools: `mem_save`, `mem_save_prompt`, `mem_capture_passive`, `mem_session_summary`, `mem_session_end`, and `mem_update`.

Each write handler now accepts an optional `directory` argument. When non-empty, project is resolved via `DetectProjectFull` on that path (with a `SourceDirBasename` safety-net fallback to the existing `resolveWriteProject()` that already exists in `resolveSessionStartProject`). When empty or missing, behavior is unchanged — REQ-308's "auto-detect from server CWD" remains the default.

The previously-introduced `resolveSessionStartProject` is renamed to `resolveWriteProjectWithDirectory` and shared by all six write handlers; the session-start call site is updated. Schemas for each tool gain a `directory` property with a description noting its purpose for remote MCP deployments.

## Why

When engram MCP runs as a remote HTTP server — for example behind `mcp-proxy` on a Synology NAS, or any setup where stdio transport is replaced with HTTP — the server's CWD has no relation to the client's project context. All write tools today auto-detect project from `os.Getwd()`, which returns the container's WORKDIR (typically `/data` or `/`). The result: every observation lands at `project="/"` instead of the actual project the client is working in, polluting the project list and breaking memory-by-project semantics.

`mem_session_start` already accepts `directory` to fix this for one tool. This PR extends the same pattern to the other six writes so the same hook-injected directory can route everything correctly.

This complements PR #281 (`feat: honor ENGRAM_URL env var in hooks`), which addresses the URL half of the same architectural mismatch — that PR lets clients reach a remote MCP server; this one lets the remote server know which project the client is in.

## Design choices

- **Per-call granularity, not a server-wide flag.** A single MCP server instance may be receiving calls from different clients in different projects, so a `--cwd` flag (as in PR #284) isn't the right fit. The `directory` argument matches the granularity that's already correct for `mem_session_start`.
- **Backward compatible.** Empty/missing `directory` preserves the REQ-308 contract exactly. No client that doesn't pass `directory` sees any behavior change.
- **Same fallback semantics as `mem_session_start`.** When the explicit directory has no `.git` anywhere up the tree (`SourceDirBasename`), the helper falls back to `resolveWriteProject()` rather than minting a project from a basename of an arbitrary client path. This mirrors the safety net introduced in 5f3329f.
- **Read tools left alone.** `mem_search`, `mem_context`, etc. already accept a `project` override and are out of scope. `mem_judge`, `mem_delete`, `mem_stats`, and `mem_current_project` aren't writes that create observations, so they're untouched.

## Test plan

- [x] `go build ./...` — clean
- [x] `go test ./...` — full suite passes (all 20 packages)
- [x] New tests added for each modified handler covering:
  - Empty `directory` → behaves as before (uses `os.Getwd()`)
  - Explicit `directory` pointing to a git repo → routes to that repo's project
  - Explicit `directory` with no git → falls back to `resolveWriteProject()` (matches `resolveSessionStartProject`'s `SourceDirBasename` fallback)
- [x] `TestResolveWriteProjectWithDirectory` exercises the renamed helper directly
- [x] `TestWriteToolsHaveDirectoryArg` asserts every write tool's schema advertises `directory`
- [x] Existing `TestSessionStartWithExplicitDirectory*` tests continue to pass after the helper rename

## Files changed

- `internal/mcp/mcp.go` — rename helper, wire `directory` through six handlers, add schema entries
- `internal/mcp/mcp_test.go` — new tests for each handler

### DIFF
--- a/internal/mcp/mcp.go
+++ b/internal/mcp/mcp.go
@@ -325,6 +325,9 @@ Examples:
 				mcp.WithString("topic_key",
 					mcp.Description("Optional topic identifier for upserts (e.g. architecture/auth-model). Reuses and updates the latest observation in same project+scope."),
 				),
+				mcp.WithString("directory",
+					mcp.Description("Working directory for project resolution. Defaults to server CWD (REQ-308 contract). Set explicitly when using a remote MCP server where the server CWD differs from the client's project context."),
+				),
 			),
 			queuedWriteHandler(writeQueue, handleSave(s, cfg, activity)),
 		)
@@ -359,6 +362,9 @@ Examples:
 				),
 				mcp.WithString("topic_key",
 					mcp.Description("New topic key (normalized internally)"),
+				),
+				mcp.WithString("directory",
+					mcp.Description("Working directory for project resolution. Defaults to server CWD (REQ-308 contract). Set explicitly when using a remote MCP server where the server CWD differs from the client's project context."),
 				),
 			),
 			queuedWriteHandler(writeQueue, handleUpdate(s)),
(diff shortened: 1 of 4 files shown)


Alan-TheGentleman:

PR #335, opened 2026-05-05
test(project): regression coverage for #312 — HOME-like config does not leak

### LINKED ISSUE #312: MCP server hardcodes ~/.engram/config.json project_name; --project flag, ENGRAM_PROJECT, ENGRAM_DATA_DIR, HOME all ignored
## Description

When running engram as an MCP server (e.g. `engram mcp --tools=agent --project=project-c`), the `project_name` value from `~/.engram/config.json` **always wins** over every documented mechanism for selecting a different project.

This breaks the typical multi-agent setup where each agent has its own engram MCP server pointing at its own logical project (e.g. `engram-b` → `project-b`, `engram-c` → `project-c`) but they share a single host. All writes end up in whatever project_name is set in `~/.engram/config.json`, regardless of how each MCP server is configured.

This appears related to (but is more severe than) #56 (closed), #202, #262, and is not addressed by the recently merged #307.

## Environment

- engram **v1.15.4** (latest at time of report)
- OS: Debian (VPS)
- Runtime: OpenClaw 2026.4.29 spawning engram MCP servers per agent
- Multi-agent setup: 3 agents (`agent-a`, `agent-b`, `agent-c`) each with own engram MCP

## Reproduction (purely engram CLI, no openclaw needed)

```bash
# 1. Set up shared config that "hijacks" project
mkdir -p ~/.engram
echo '{"project_name": "project-a"}' > ~/.engram/config.json

# 2. Try to run engram MCP for a DIFFERENT project using every override mechanism
echo '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"mem_current_project","arguments":{}}}' \
  | ENGRAM_DATA_DIR=/some/other/dir \
    ENGRAM_PROJECT=project-c \
    HOME=/tmp/fake-home \
    /usr/local/bin/engram mcp --tools=agent --project=project-c
```

## Expected

`mem_current_project` returns `project-c` (because `--project` flag was passed).

Per docs:
- `--project` flag: \"Override detected project name (default: git remote → cwd)\"
- `ENGRAM_PROJECT`: \"Override auto-detected project name for MCP server\"
- `ENGRAM_DATA_DIR`: \"Override data directory (default: ~/.engram)\"

## Actual

```json
{
  \"available_projects\": null,
  \"cwd\": \"/root\",
  \"project\": \"project-a\",
  \"project_path\": \"/root\",
  \"project_source\": \"config\"
}
```

`project_source: \"config\"` confirms engram is reading `~/.engram/config.json` and ignoring **all four** override mechanisms (`--project`, `ENGRAM_PROJECT`, `ENGRAM_DATA_DIR`, `HOME`).

## What I tried (all failed)

| Approach | Result |
|---|---|
| `--project=X` CLI flag in MCP args | Ignored, returns config.json value |
| `ENGRAM_PROJECT=X` env var on MCP process | Ignored |
| `ENGRAM_DATA_DIR=/different/path` env var | Ignored — still reads `~/.engram/config.json` |
| `HOME=/tmp/fake` env var | Ignored — still reads `/root/.engram/config.json` |
| Pass `project=X` parameter inside `mem_save` MCP call | Ignored |

Note: the **CLI** mode (`engram save --project X`) DOES respect `--project`. Only the **MCP** mode is affected.

## Why this matters (real-world impact)

In multi-agent deployments where each agent gets its own engram MCP scoped to a logical project, this bug means **all agents write to whichever single project is in `config.json`**, destroying compartmentalization. There is no way to opt out short of running each engram MCP from a separate `chroot` or container, which defeats the purpose of having a single binary with `--project`.

Concrete case: 3 chat-driven agents each with own MCP config pointing at different projects. ALL writes silently go to one shared project. Agents can detect the discrepancy in their responses (e.g., \"saved but in wrong project\") but cannot fix it.

## Suggested resolution priority

`--project` CLI flag SHOULD have higher precedence than `~/.engram/config.json`. Suggested precedence (highest to lowest):

1. `--project` CLI flag
2. `ENGRAM_PROJECT` env var
3. `~/.engram/config.json` `project_name`
4. Auto-detect (cwd / git remote)

This matches expectations from the help text and the documented purpose of each mechanism (\"Override\").

## Related issues / PRs

- #56 ENGRAM_DATA_DIR Isn't Honored (CLOSED — partially fixed for cloud.json only)
- #262 Support explicit project/cwd targeting for MCP memory tools
- #202 First-class data-directory selection
- #307 (merged) Recover ambiguous project writes — does not address this
- #213 (open) Remove DefaultProject from reads
- #289 (open) Honor directory arg on remaining write tools

Happy to provide additional repro details or test a fix.

ricardoarz-dev:
## Summary

Adds two regression tests in \`internal/project/detect_test.go\` that pin the contract \`.engram/config.json\` is a project/repo lock — never inherited from arbitrary ancestors. They cover the multi-agent shape reported in #312, where each agent's MCP server runs from its own cwd under a shared \`\$HOME\` that contains \`~/.engram/config.json\`.

Tests-only PR — no behavior change. The actual fix that prevents \`~/.engram/config.json\` from contaminating arbitrary cwds is already in main (\`detectFromConfig\` only reads at the git-root or the input dir, never walking to ancestors). These tests blind-side regressions of that behavior.

Closes #312 if maintainers agree the issue is resolved by the existing fix + this regression coverage. Otherwise happy to keep iterating.

## Why these specific tests

\`TestDetectProjectFull_DoesNotInheritParentConfigOutsideGitRepo\` already covers the immediate-parent + plain-child case. The two shapes added here match what #312 actually reproduces in real multi-agent deployments:

1. **\`TestDetectProjectFull_HomeLikeConfigDoesNotLeakIntoNestedGitRepo\`** — config at a HOME-like ancestor, cwd inside a git repo several levels below (e.g. \`\$HOME/repos/agent-b/src/pkg\`). Asserts \`SourceGitRoot\` wins over the ancestor config; project resolves to the repo basename, not the HOME-level \`project_name\`.

2. **\`TestDetectProjectFull_HomeLikeConfigDoesNotLeakIntoDeeplyNestedNonGitCwd\`** — config at a HOME-like ancestor, cwd several levels below outside any git tree. Asserts \`SourceDirBasename\` of the leaf cwd, not the ancestor's \`project_name\`.

Both shapes are described in #312's reproduction (\"3 agents (\`agent-a\`, \`agent-b\`, \`agent-c\`) each with own engram MCP\" sharing \`/root/.engram/config.json\`).

## Test plan

- [x] \`go test ./internal/project/\` — passes (5.7s)
- [x] \`go build ./...\` — clean
- [x] New tests fail when \`detectFromConfig\` is altered to walk ancestors (verified locally by temporarily widening the lookup; both new tests fail with \`HOME-like ancestor config leaked\`, as expected)
- [x] Existing \`TestDetectProjectFull_*\` tests continue to pass

## Files changed

- \`internal/project/detect_test.go\` — +94 lines (two new test functions, no edits to existing tests)

## Notes

- Scope kept tight per Alan's note on #312 (\"Keep the implementation tight and include regression coverage where behavior changes\"). No production code touched.
- Followup option (out of scope here): a parallel test in \`internal/mcp/mcp_test.go\` that exercises the same shape end-to-end via \`resolveWriteProject()\`. Happy to add in a follow-up if useful — left out of this PR to keep the diff minimal.

### DIFF
--- a/internal/project/detect_test.go
+++ b/internal/project/detect_test.go
@@ -581,6 +581,100 @@ func TestDetectProject_MatchesFull(t *testing.T) {
 	}
 }
 
+// TestDetectProjectFull_HomeLikeConfigDoesNotLeakIntoNestedGitRepo asserts
+// that an ancestor `.engram/config.json` (e.g. one sitting at $HOME, the
+// scenario reported in #312) does NOT contaminate project detection for a
+// cwd that is inside a nested git repository several levels below.
+//
+// The repo's own root is the project lock; ancestor configs above it must be
+// ignored. This guards the multi-agent setup where each agent runs from its
+// own repo under a shared $HOME — every agent must resolve to its own repo
+// project, not the HOME-level project_name.
+func TestDetectProjectFull_HomeLikeConfigDoesNotLeakIntoNestedGitRepo(t *testing.T) {
+	homeLike := t.TempDir()
+	configDir := filepath.Join(homeLike, ".engram")
+	if err := os.MkdirAll(configDir, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(configDir, "config.json"), []byte(`{"project_name":"home-lock"}`), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	// Nested git repo a few levels below the HOME-like dir, with no config of
+	// its own. cwd is a subdirectory inside the repo (mirrors typical agent
+	// workflow: cwd is somewhere inside the repo tree, not necessarily root).
+	repoRoot := filepath.Join(homeLike, "repos", "agent-b")
+	if err := os.MkdirAll(repoRoot, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	initGit(t, repoRoot)
+	cwd := filepath.Join(repoRoot, "src", "pkg")
+	if err := os.MkdirAll(cwd, 0o755); err != nil {
+		t.Fatal(err)
+	}
+
+	res := DetectProjectFull(cwd)
+
+	if res.Error != nil {
+		t.Fatalf("unexpected detection error: %v", res.Error)
+	}
+	if res.Project == "home-lock" {
+		t.Fatalf("HOME-like ancestor config leaked into nested repo cwd: project=%q source=%q", res.Project, res.Source)
+	}
+	if res.Source != SourceGitRoot {
+		t.Fatalf("expected SourceGitRoot for nested repo without remote, got source=%q project=%q", res.Source, res.Project)
+	}
+	if res.Project != "agent-b" {
+		t.Fatalf("expected repo basename %q, got %q", "agent-b", res.Project)
+	}
+	gotPath, _ := filepath.EvalSymlinks(res.Path)
+	wantPath, _ := filepath.EvalSymlinks(repoRoot)
+	if gotPath != wantPath {
+		t.Fatalf("expected path %q, got %q", wantPath, gotPath)
+	}
+}
+
+// TestDetectProjectFull_HomeLikeConfigDoesNotLeakIntoDeeplyNestedNonGitCwd
+// asserts that an ancestor `.engram/config.json` does not contaminate a
+// deeply-nested cwd that is not inside a git repo. This is the second half
+// of the #312 scenario: agents that operate outside any git tree must still
+// resolve to their own cwd basename, not to the HOME-level project_name.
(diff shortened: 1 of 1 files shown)


Alan-TheGentleman:
Thanks for the regression coverage proposal. The underlying project override / HOME-like config leakage work was resolved by PR #378, which closed #312 along with related project override fixes.

Closing this PR as superseded by the merged implementation. If there is still a distinct uncovered regression, please open a fresh issue describing that narrower gap.

PR #336, opened 2026-05-05
fix(server): add GET /observations endpoint for plugin compatibility

### LINKED ISSUE #288: Plugin save-nudge silently broken: GET /observations returns 405, hook never warns
### 📋 Pre-flight Checks

## Summary

  The Claude Code plugin's `user-prompt-submit.sh` hook attempts to detect when the user has gone too long without saving memory and inject a "save reminder" system
  message. It's broken: the HTTP request it depends on returns 405 and the hook silently fails, so the nudge never fires. Users (myself included) assume the nudge logic is
  working when in fact it never has.

  ## Versions

  - Engram binary: **1.14.13** (installed via `brew upgrade gentleman-programming/tap/engram`)
  - Plugin (marketplace HEAD): **`5f3329f`** (`fix: harden mcp and cloud recovery paths`)
  - Plugin script `user-prompt-submit.sh`: unchanged in upstream main since first commit
  - OS: macOS 14.8.4 (arm64)
  - Claude Code: latest

### 🔄 Steps to Reproduce

  ```bash
  # Start the server
  engram serve &

  # Hit the endpoint the plugin uses
  curl -i "http://127.0.0.1:7437/observations?project=anything&limit=1&sort=created_at:desc"

  Actual response:

  HTTP/1.1 405 Method Not Allowed
  Allow: POST
  Content-Type: text/plain; charset=utf-8

  Method Not Allowed

  The server only accepts POST /observations (which is the create endpoint, requires session_id, title, content).

  Why this is silent

  plugin/claude-code/scripts/user-prompt-submit.sh:95-97:

  LAST_SAVE_JSON=$(curl -sf \
    "${ENGRAM_URL}/observations?project=${ENCODED_PROJECT}&limit=1&sort=created_at:desc" \
    --max-time 0.2 2>/dev/null)

  - -sf makes curl exit non-zero on 4xx/5xx without printing anything
  - --max-time 0.2 makes any latency look like a network failure
  - 2>/dev/null swallows the rest

  The script then checks if [ -z "$LAST_SAVE_JSON" ] and returns early with no nudge. From the user's perspective, the absence of a nudge is indistinguishable from "I
  haven't crossed the 15-minute threshold yet" — so nobody notices the bug.


### ✅ Expected Behavior


  Either:
  1. Server exposes a read-only GET /observations endpoint (most idiomatic — REST listing). The script as written would just work.
  2. Plugin script switches to a different mechanism (e.g., shell out to engram search --json or engram context --json) that the server already supports.

  Suggested fix (option 1, server-side)

  Add a GET /observations handler in the HTTP server, accepting project, limit, sort query params, returning [{...}, ...] JSON. Keep the existing POST /observations for
  create.

  Happy to PR if the project owner confirms which direction is preferred.

  Side observation

  /health always reports "version":"0.1.0" regardless of the actual binary version (tested with 1.11.0 and 1.14.13). Looks hardcoded — might be worth surfacing the real
  binary version for diagnostics.


### ❌ Actual Behavior

  The endpoint rejects GET and only allows POST (which is the create endpoint, not list):

  ```
  HTTP/1.1 405 Method Not Allowed
  Allow: POST
  Content-Type: text/plain; charset=utf-8

  Method Not Allowed
  ```

  In `plugin/claude-code/scripts/user-prompt-submit.sh:95-97`:

  ```bash
  LAST_SAVE_JSON=$(curl -sf \
    "${ENGRAM_URL}/observations?project=${ENCODED_PROJECT}&limit=1&sort=created_at:desc" \
    --max-time 0.2 2>/dev/null)
  ```

  `-sf` makes curl exit non-zero on 4xx/5xx without printing anything, and `2>/dev/null` swallows the rest. The script then checks `if [ -z "$LAST_SAVE_JSON" ]` and returns
   early. From the user's perspective, no nudge ever fires — and the absence is indistinguishable from "the threshold hasn't been crossed yet."

### Operating System

macOS

### Engram Version

1.14.13

### Agent / Client

Claude Code

### 📋 Relevant Logs

```shell

```

### 💡 Additional Context

_No response_

Wild-sergunys:
## What
Added GET /observations endpoint that reuses handleRecentObservations handler.

## Why
The save-nudge plugin sends GET /observations?project=...&limit=1&sort=created_at:desc
but the server only had POST /observations, returning 405 Method Not Allowed.
The plugin silently failed because curl -sf suppresses error output.

## How
- Registered new route `GET /observations` pointing to `handleRecentObservations`
- Updated DOCS.md with the new endpoint
- Added test coverage

## Testing
- `main`: `curl -s -w "\nHTTP_CODE: %{http_code}\n" "..."` -> **405 Method Not Allowed**  
- This branch: same curl -> **200 OK**
- All existing server tests pass
- New test `TestHandleGetObservationsWithSort` passes

Closes #288

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -133,6 +133,7 @@ Engram is local-first: local SQLite is authoritative; cloud features are optiona
 
 - `POST /observations` — Add observation. Body: `{session_id, type, title, content, tool_name?, project?, scope?, topic_key?}`
 - `GET /observations/recent` — Recent observations. Query: `?project=X&scope=project|personal&limit=N`
+- `GET /observations` — List observations. Query: `?project=X&limit=N&sort=created_at:desc`. Same behaviour as /recent; accepts sort parameter for plugin compatibility
 - `GET /observations/{id}` — Get single observation by ID
 - `PATCH /observations/{id}` — Update fields. Body: `{title?, content?, type?, project?, scope?, topic_key?}`
 - `DELETE /observations/{id}` — Delete observation (`?hard=true` for hard delete, soft delete by default)
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -147,6 +147,7 @@ func (s *Server) routes() {
 	s.mux.HandleFunc("POST /observations", s.handleAddObservation)
 	s.mux.HandleFunc("POST /observations/passive", s.handlePassiveCapture)
 	s.mux.HandleFunc("GET /observations/recent", s.handleRecentObservations)
+	s.mux.HandleFunc("GET /observations", s.handleRecentObservations)
 	s.mux.HandleFunc("PATCH /observations/{id}", s.handleUpdateObservation)
 	s.mux.HandleFunc("DELETE /observations/{id}", s.handleDeleteObservation)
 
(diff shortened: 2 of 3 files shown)


Alan-TheGentleman:
Thanks for the contribution. This has since been implemented and merged in PR #377, which fixed #288 by adding the `GET /observations` compatibility endpoint with docs and tests.

Closing this PR as superseded by the merged implementation.

PR #390, opened 2026-05-18
docs(mcp): align server instructions with Claude Code plugin behavior

fedeiglesias:
## Summary
- Realigns `serverInstructions` and the two Claude Code hooks (`session-start.sh`, `post-compaction.sh`) so they no longer promise "always available — use without ToolSearch" for core memory tools.
- Documents the underlying Claude Code behavior: tools from plugin-loaded MCP servers (prefix `mcp__plugin_*`) are currently treated as deferred regardless of the server-emitted `defer_loading` flag.
- The plugin already declares core tools with `defer_loading: false`; this PR only changes user-facing copy so the contract matches what Claude Code actually does today.

## Repro of the underlying behavior
1. Probe the MCP server's `tools/list` directly — `mem_get_observation` and the other 6 core tools are emitted without `defer_loading` (which is `false` per the JSON contract; verified with raw stdio probe to `engram mcp --tools=agent`).
2. In a fresh Claude Code session, calling `mcp__plugin_engram_engram__mem_get_observation` fails with `id is required` — the tool schema is not loaded.
3. After `ToolSearch(query: "select:mcp__plugin_engram_engram__mem_get_observation")`, the same call succeeds and stays callable for the rest of the session.

The root fix lives in the Claude Code harness, not in Engram. This PR only makes the user-facing docs honest while that's pending upstream.

## Test plan
- [ ] `TestCoreToolsAreNotDeferred` / `TestNonCoreToolsAreDeferred` — annotations not touched, expected unchanged.
- [ ] `TestServerInstructionsConstantIsNonEmpty` — keywords preserved (`mem_save`, `mem_search`, `mem_context`, `mem_session_summary`).
- [ ] `TestServerInstructions_ConflictSurfacingBlock` — phrases preserved (`judgment_required`, `candidates[]`, `mem_judge`, `0.7`, `supersedes`, `conflicts_with`, `architecture`, `conversationally`, `evidence`).
- [ ] Hook scripts dry-run produce the new `MEMORY TOOLS` block (verified locally with `bash session-start.sh < <(echo '{"session_id":"x","cwd":"."}')`).

### DIFF
--- a/internal/mcp/mcp.go
+++ b/internal/mcp/mcp.go
@@ -167,11 +167,15 @@ func NewServer(s *store.Store) *server.MCPServer {
 }
 
 // serverInstructions tells MCP clients when to use Engram's tools.
-// 7 core tools are eager (always in context). The rest are deferred
-// and require ToolSearch to load.
+// Tools are split into two pools via the defer_loading annotation:
+//   - eager  (defer_loading=false): clients should keep them in context
+//   - deferred (defer_loading=true): clients load them via Tool Search on demand
+// NOTE: Claude Code currently ignores defer_loading for tools that arrive
+// through a plugin (prefix mcp__plugin_*) and treats all of them as deferred.
+// The CLAUDE CODE PLUGIN USERS block below documents the workaround.
 const serverInstructions = `Engram provides persistent memory that survives across sessions and compactions.
 
-CORE TOOLS (always available — use without ToolSearch):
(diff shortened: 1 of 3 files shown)


Alan-TheGentleman:
Closing this because it skips the issue-first process. No PR merges here without a linked `status:approved` issue, and this one has none. The server-instruction alignment may well be worth doing, so open an issue describing the mismatch, get it approved, then re-file with `Closes #N`.

PR #407, opened 2026-05-23
fix(mcp): apply process-level project override in handleSessionSummary

### LINKED ISSUE #403: bug(mcp): mem_session_summary and mem_save ignore explicit project parameter, always write to git-root project
## 📝 Bug Description

When working on a project whose directory is **different from the Claude Code working directory**, `mem_session_summary` and `mem_save` ignore the explicit `project` parameter passed in the tool call and always save to the project detected from the git root of the **current working directory** (`cwd`).

This causes all memories and session summaries to be saved under the wrong project, silently mixing data between unrelated projects.

**Root cause identified in source code:**

`handleSessionSummary` (and several other write handlers) call `resolveWriteProject()` directly, which only uses `os.Getwd()`. They do **not** call `resolveWriteProjectWithProcessOverride(cfg.DefaultProject)`, which would honor the process-level override.

The comment in the code says:
```go
// project field intentionally not read — auto-detect only (REQ-308 write-tool contract)
detRes, err := resolveWriteProject()
```

This correctly ignores the per-call `project` argument from the AI agent (REQ-308). However, it also accidentally ignores `cfg.DefaultProject` — the **trusted process-level override** set at startup via `ENGRAM_PROJECT` or `engram mcp --project`. These are two different authorities and should be treated differently.

The fix is a one-line change in `handleSessionSummary`:
```go
// Before:
detRes, err := resolveWriteProject()

// After:
detRes, err := resolveWriteProjectWithProcessOverride(cfg.DefaultProject)
```

`resolveWriteProjectWithProcessOverride` already exists in the codebase and is used correctly by other handlers (e.g. `handleSave`). The comment should also be updated to clarify that process-level overrides ARE honored, while per-call `project` arguments remain ignored per REQ-308.

## 🔄 Steps to Reproduce

1. Open Claude Code from directory `/Users/me/project-a` (a git repo)
2. Configure engram MCP in Claude Code settings
3. Start working on a different project at `/Users/me/project-b`
4. Call `mem_save` or `mem_session_summary` with `project: "project-b"` explicitly
5. Observe that the response envelope shows `"project": "project-a"` (the git root of the cwd, not the requested project)
6. All memories are saved under `project-a` regardless of the `project` parameter

**Verified in code:**
- `internal/mcp/mcp.go` → `handleSessionSummary` uses `resolveWriteProject()` (line ~1604)
- `internal/mcp/mcp.go` → `handleSave` uses `resolveSaveWriteProjectWithProcessOverride(..., cfg.DefaultProject)` ✅
- The inconsistency means `mem_save` and `mem_session_summary` behave differently depending on whether `cfg.DefaultProject` is set

## ✅ Expected Behavior

When `mem_session_summary` is called with `project: "claudestat"` from a Claude Code session whose `cwd` is `/Users/me/nutritrack`, the summary should be saved under project `"claudestat"`, not `"nutritrack"`.

More generally: `mem_session_summary` should honor `cfg.DefaultProject` (the process-level override) the same way `mem_save` does.

## ❌ Actual Behavior

The tool call response always shows:
```json
{
  "project": "nutritrack",
  "project_path": "/Users/me/nutritrack",
  "project_source": "git_root",
  "result": "Session summary saved for project \"nutritrack\""
}
```

Even when called with `project: "claudestat"` explicitly. The `project` parameter in the response is `"nutritrack"` because `handleSessionSummary` ignores both `cfg.DefaultProject` and the per-call `project` argument.

## 🖥️ Environment

- **OS**: macOS
- **Engram Version**: 1.15.8
- **Agent / Client**: Claude Code

## 📋 Relevant Logs

Tool call result from Claude Code:
```
plugin:engram:engram - Save Session Summary (MCP)
{
  "project": "nutritrack",
  "project_path": "/Users/db/nutritrack",
  "project_source": "git_root",
  "result": "Session summary saved for project \"nutritrack\""
}
```
Called with explicit `project: "claudestat"` parameter.

## 💡 Additional Context

- Related to issue #387 which identifies the same root cause but was not submitted with the issue template (no labels assigned)
- The infrastructure for the fix already exists: `resolveWriteProjectWithProcessOverride(cfg.DefaultProject)` is implemented and used by other write handlers
- The branch `fix/mcp-project-override` in the upstream repo adds this infrastructure but does not apply it to `handleSessionSummary`
- I am ready to submit a PR with the one-line fix + test as soon as this issue is approved

DeibyGS:
## Problem

Closes #403

`mem_session_summary` calls `resolveWriteProject()` directly, which ignores the process-level project override set via `--project` flag or `ENGRAM_PROJECT` env var (`cfg.DefaultProject`).

All other write-tool handlers in this file use `resolveWriteProjectWithProcessOverride(cfg.DefaultProject)`, but `handleSessionSummary` was calling the bare variant — so even when the MCP server is started as `engram mcp --project claudestat`, `mem_session_summary` would still auto-detect from cwd and write to the wrong project.

## Fix

One-line change: replace `resolveWriteProject()` with `resolveWriteProjectWithProcessOverride(cfg.DefaultProject)` in `handleSessionSummary`.

This aligns the handler with the pattern used by every other write-tool and respects REQ-308 (no per-call `project` parameter is added back).

## Testing

`go test ./internal/mcp/...` passes with no changes to existing tests.

### DIFF
--- a/internal/mcp/mcp.go
+++ b/internal/mcp/mcp.go
@@ -1600,8 +1600,8 @@ func handleSessionSummary(s *store.Store, cfg MCPConfig, activity *SessionActivi
 		sessionID, _ := req.GetArguments()["session_id"].(string)
 		// project field intentionally not read — auto-detect only (REQ-308 write-tool contract)
 
-		// Auto-detect project from cwd; fail fast on ambiguous (REQ-308, REQ-309)
-		detRes, err := resolveWriteProject()
+		// Auto-detect project from cwd; process-level override (--project / ENGRAM_PROJECT) takes precedence.
+		detRes, err := resolveWriteProjectWithProcessOverride(cfg.DefaultProject)
 		if err != nil {
 			return writeProjectErrorResult(nil, "", detRes, err), nil
 		}


Alan-TheGentleman:
Closing this because the linked issue #403 has no approval yet, and no PR merges here without a `status:approved` issue. The one-line `resolveWriteProject` to process-override fix looks correct, so get #403 approved first, then reopen this PR. I am flagging #403 for triage.

Issue #352, opened 2026-05-10
feat(store): mem_search returns 0 when no observation contains all query tokens


Basparin:
### 📋 Pre-flight Checks

- [X] I have searched [existing issues](https://github.com/Gentleman-Programming/engram/issues) and this is not a duplicate
- [X] I understand this issue needs `status:approved` before a PR can be opened

### 🔍 Problem Description

`mem_search` returns 0 results for multi-token queries when no single observation contains every token, even if observations exist for each token individually. The failure is silent: callers see "no results" and conclude the data is missing, sometimes saving duplicates of memos that already exist.

### 💡 Proposed Solution

Add a `match_mode` option (`"all" | "any"`) to `Store.Search()` and propagate to the `mem_search` MCP tool. The OR branch can reuse the join-with-OR logic already present in `sanitizeFTSCandidates` (`internal/store/relations.go`), which `FindCandidates` uses internally for the same kind of relaxed match.

Example usage:

```
mem_search(query: "auth compliance session", match_mode: "any")
```

### 📦 Affected Area

Store (database, queries)

### 🔄 Alternatives Considered

- Auto-relax fallback: retry as OR when AND returns 0, with a flag in the response so callers can detect the relaxation. Hides the param at API level.
- Client-side query validation: fragile, every consumer must implement and maintain it independently.
- Related work: #257 (trigram + fuzzy) targets a different failure mode (typo/substring); #168 (semantic memory enhancement) is broader scope (embeddings).

### 📎 Additional Context

Synthetic repro (fresh database, observations with partial overlap):

```
mem_save(title: "Auth session middleware", content: "")
mem_save(title: "Compliance audit notes",   content: "session policy")
mem_save(title: "OAuth tokens",              content: "auth and compliance")

mem_search("auth")        -> 2 hits
mem_search("compliance")  -> 2 hits
mem_search("session")     -> 2 hits
mem_search("auth compliance session") -> 0 hits
```

Each observation contains 2 of the 3 query tokens; the AND query still drops everything because no single observation contains all three. Root cause at `internal/store/store.go:5874-5884`: `sanitizeFTS()` wraps each token in quotes joined by spaces, and FTS5 reads spaces between quoted tokens as implicit AND.

**Stopgap mitigation (until `match_mode` lands)**

A minimal client-side pre-tool advisory catches the silent collapse before it hits the store. Wired as a Claude Code `PreToolUse` hook on `mcp__plugin_engram_engram__mem_search`:

```python
[...]

PR #448, opened 2026-05-29
fix(store): match ANY term in multi-word FTS queries (OR, not implicit AND)

joniiiAi:
## Problem

`sanitizeFTS` wraps each query word in quotes and joins them with a **space**. FTS5 treats a space between terms as an **implicit AND**, so a multi-word query only matches documents that contain **every** term.

For natural-language / multi-word searches this means near-zero recall — the query returns 0 hits unless one single document happens to contain all the words.

Reproducible at the SQL level on any populated DB:

```sql
-- implicit AND (current behaviour): every term required
SELECT count(*) FROM observations_fts
WHERE observations_fts MATCH '"email" "pipeline" "stalled" "leads"';   -- 0

-- OR: matches any term
SELECT count(*) FROM observations_fts
WHERE observations_fts MATCH '"email" OR "pipeline" OR "stalled" OR "leads"'; -- many
```

In practice `mem_search` (and the HTTP `/search` + context-injection hook, which share this helper) return nothing for typical multi-word queries while the data is clearly present and findable by single-term search.

## Fix

Join the quoted terms with `OR` instead of a space.

Relevance is preserved by the callers' existing `ORDER BY fts.rank` (bm25): documents matching **more** — and rarer — terms still rank highest. So the previous "best" matches stay on top, and partial matches simply become reachable instead of being dropped entirely. Each term stays quoted, so the original special-char protection is unchanged.

Diff is 3 lines of logic (plus an updated doc comment).

## Alternative considered

A hybrid "AND first, fall back to OR if 0 results" would also work and keeps strict-AND precision when it happens to match. I went with plain `OR` + bm25 because it's a one-line change, matches how FTS engines are normally queried, and the rank ordering already delivers the precision benefit (all-term matches float to the top) without a second query.

## Tests

- `TestSanitizeFTS` — table-driven unit test for the OR join, quote-stripping, empty/blank, single term.
- `TestSearchMatchesAnyTermInMultiWordQuery` — regression test: a query whose terms are only partially present now returns the document (0 results before this change).

```
ok  github.com/Gentleman-Programming/engram/internal/store
```

Verified locally end-to-end: rebuilt the binary, and `mem_search` / `engram search` / the `/search` HTTP endpoint now return ranked results for previously-empty multi-word queries.

### DIFF
--- a/internal/store/sanitize_fts_test.go
+++ b/internal/store/sanitize_fts_test.go
@@ -1,0 +1,64 @@
+package store
+
+import "testing"
+
+// TestSanitizeFTS verifies multi-word queries are joined with OR (not an
+// implicit AND), while each term stays quoted to survive FTS5 special chars.
+func TestSanitizeFTS(t *testing.T) {
+	cases := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{"empty", "", ""},
+		{"blank", "   ", ""},
+		{"single", "auth", `"auth"`},
+		{"multi", "fix auth bug", `"fix" OR "auth" OR "bug"`},
+		{"strips existing quotes", `"fix" auth`, `"fix" OR "auth"`},
+		{"collapses whitespace", "a   b", `"a" OR "b"`},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if got := sanitizeFTS(tc.in); got != tc.want {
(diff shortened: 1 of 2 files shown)


Alan-TheGentleman:
Closing this as superseded by #352 and #475. The recall issue is real, but changing multi-word search to OR globally is broader than we want. The `match_mode` path keeps the existing default while allowing broader recall explicitly, so future work should continue there.

PR #463, opened 2026-06-04
fix(sync): defer orphan observations on pull, reject sessions without directory on push

captainsparrow10:
## Problem

A cloud sync deadlock: an `observation` reached the cloud **without its parent `session`**, so every client pulling it hit `FOREIGN KEY constraint failed (787)` and the entire pull stream stalled (cursor stuck, no new memories from the team).

Root cause (two independent gaps):
1. **Server** (`/sync/mutations/push`) does not validate session payloads — a legacy session without `directory` is blocked client-side and never pushed, but its child observation is pushed and accepted → orphan in the cloud. The `directory` / referential checks only existed on the legacy `/sync/push` (chunk) path.
2. **Client** (`ApplyPulledMutation` / `ApplyPulledChunk`) only deferred FK-misses for `relation`; for `observation`/`session`/`prompt` it failed hard and halted the pull. The `sync_apply_deferred` mechanism existed but was hardcoded to relations.

## Fix

**Client — pull resilience**
- Defer legacy entities (observation/session/prompt) on a SQLite FK-miss (787) to `sync_apply_deferred` and advance the cursor, instead of failing hard. Applied to both `ApplyPulledMutation` (autosync) and `ApplyPulledChunk` (initial sync).
- `sync_apply_deferred` gains an `op` column (idempotent migration) so a deferred **delete** replays as a delete, not an upsert.
- Deferred-row delete moved inside the apply tx (crash safety).

**Server — prevent orphans at origin**
- `/sync/mutations/push` now rejects a session **upsert** without `directory`. Delete variants (`deleted` / `hard_delete` / `deleted_at` / `op=delete`) are exempt.

## Tests
- `TestApplyPulledObservation_DefersOnSessionFKMiss` / `_ReplayAfterSessionArrives`
- `TestApplyPulledChunk_DefersOrphanObservation`
- `TestReplayDeferred_PreservesDeleteOp`
- `TestHandleMutationPush_SessionMissingDirectory_Returns400` + delete-variant exemptions (`hard_delete`, `deleted_at`, `op=delete`)
- Updated `TestApplyPulledChunkIsAtomicAndRetrySafe` to use a genuinely unrecoverable error (FK-miss is now deferrable, not fatal).

`go build ./...`, `go vet`, and `go test ./internal/store/ ./internal/cloud/cloudserver/` all green. (`cmd/engram` failures are pre-existing — require network/token.)

### DIFF
--- a/internal/cloud/autosync/manager.go
+++ b/internal/cloud/autosync/manager.go
@@ -589,9 +589,11 @@ func (m *Manager) pull(ctx context.Context) error {
 				OccurredAt: rm.OccurredAt,
 			}
 			// Phase E: per-entity error policy (design §9).
-			// ApplyPulledMutation handles relation FK misses internally by writing
-			// to sync_apply_deferred and returning nil — the cursor advances normally.
-			// All other errors (legacy entities, decode errors) propagate and halt the pull.
+			// ApplyPulledMutation handles FK misses by writing to sync_apply_deferred
+			// and returning nil — the cursor advances normally. This covers both
+			// relation FK misses and legacy entities (observation/session/prompt) whose
+			// parent session is not yet present locally; both defer instead of halting.
+			// Decode errors and other unrecoverable failures propagate and halt the pull.
 			if err := m.store.ApplyPulledMutation(m.cfg.TargetKey, localMut); err != nil {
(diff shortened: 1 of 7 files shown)


captainsparrow10:
Closing — opened by mistake; this fix was intended for internal use, not as an upstream contribution.

PR #465, opened 2026-06-04
fix(search): OR+prefix FTS query + title-weighted bm25 to fix natural-language recall

omarhash95:
## Problem

`Search()` runs `mem_search` through `sanitizeFTS`, which quotes each term and joins them with **spaces** — i.e. FTS5 **implicit AND**. Every term must match, so any natural-language / paraphrased / multi-term query returns nothing even when the answer is stored verbatim. Examples that returned `No memories found` on v1.16.1 against a populated DB:

- `"how do I permanently remove a saved memory from engram"`
- `"should we adopt X for better memory recall"`
- `"How long have Mel ..."` (where the stored text says `Melanie`)

The longer / more natural the query, the more certain the miss. (I measured ~0% recall on a multi-question retrieval slice with raw questions; short keyword queries worked, natural questions did not.)

## Change

Two small, localized edits in `internal/store/store.go`:

1. **`sanitizeFTS`** now emits `"t1"* OR "t2"* OR ...` instead of space-joined AND:
   - `OR` restores recall (any term may match),
   - prefix (`*`) makes `mel` match `melanie`, `delet` match `deleted`,
   - internal quotes are escaped (`""`).
2. **`Search()` ORDER BY** uses weighted `bm25(observations_fts, 10,2,1,1,1,8)` (title 10×, topic_key 8×, content 2×) so the OR expansion still surfaces the most on-topic observation first rather than any doc that merely shares a common term.

Existing `internal/store` tests pass.

## Result (same DB, before → after)

| query | before | after |
|---|---|---|
| "should we adopt X for better memory recall" | 0 | correct obs at #1 |
| "what did we learn about retrieval quality being poor" | 0 | correct obs at #1 |
| "how do I permanently remove a saved memory" | 0 | relevant obs returned |

Precise keyword queries still rank tightly because bm25 orders multi-term matches highest.

## Tradeoff + happy to adjust

Defaulting to `OR` favors **recall over precision**, which is a behavior change. If you'd prefer to preserve strict-AND precision, two options I'm glad to implement instead:
- **AND-first, OR-fallback**: run the current implicit-AND query, and only expand to OR when it returns fewer than N results — best of both.
- **Config flag** to opt into the OR behavior.

I can also add unit tests for `sanitizeFTS` and a `Search` recall test if you'd like. Let me know which direction you prefer.

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -3039,7 +3039,12 @@ func (s *Store) Search(query string, opts SearchOptions) ([]SearchResult, error)
 		args = append(args, normalizeScope(opts.Scope))
 	}
 
-	sqlQ += " ORDER BY fts.rank LIMIT ?"
+	// Weighted bm25: favor matches in `title` (10×) and `topic_key` (8×) over
+	// `content` (2×) so the OR-recall expansion in sanitizeFTS still surfaces the
+	// most on-topic observation first instead of any doc that merely shares a
+	// common term. Column order matches the FTS5 schema:
+	// title, content, tool_name, type, project, topic_key. (regtime/fts-tuning)
+	sqlQ += " ORDER BY bm25(observations_fts, 10.0, 2.0, 1.0, 1.0, 1.0, 8.0) LIMIT ?"
 	args = append(args, limit)
 
 	rows, err := s.queryItHook(s.db, sqlQ, args...)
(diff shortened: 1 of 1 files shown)


Alan-TheGentleman:
Closing this because the search-recall problem is already being handled through #352 and #475 with a safer `match_mode` approach. Changing the default FTS behavior globally to OR plus prefix matching is too risky for the current contract. If more search tuning is needed after #475, we should do it as a follow-up issue with explicit behavior and ranking expectations.

Issue #257, opened 2026-04-27
feat(store): trigram + fuzzy search with weighted confidence scores


Giorgiosaud:
### Pre-flight Checks
- [x] I have searched existing issues and this is not a duplicate
- [x] I understand this issue needs status:approved before a PR can be opened

### Problem Description
The current `Search()` implementation wraps every term in quotes before passing to FTS5, producing exact phrase matches only. This means:
- `auth` does not find `authentication`
- `autentication` (typo) finds nothing
- Short or partial terms return empty results even when relevant memories exist

### Proposed Solution
Replace the single-pass exact-match search with a **3-pass weighted pipeline**:

**Pass 1 — FTS5 trigram** (score 0.6–1.0)
Rebuild `observations_fts` with `tokenize="trigram"`. Trigram natively handles prefix and substring matching. Score is normalized from FTS5 rank.

**Pass 2 — Prefix fallback** (score 0.5)
Runs if Pass 1 returns fewer than 3 results. Uses `term*` FTS5 prefix syntax.

**Pass 3 — Levenshtein fuzzy** (score 0.1–0.49)
Runs only if Pass 1 + Pass 2 return nothing. Go-side edit distance against a capped pool of 500 recent titles.

**Weighted average scoring**
Results appearing in multiple passes receive a weighted average score:
- FTS5 weight: 0.6 — Prefix weight: 0.3 — Fuzzy weight: 0.1

`SearchResult` gains a public `Score float64` field (0.0–1.0). Results sorted by score DESC.

**Migration**
An idempotent `migrateFTSTrigram()` rebuilds the FTS table in a transaction. Existing triggers are unchanged.

### Affected Area
Core search / store layer (`internal/store/store.go`)

### Alternatives Considered
- **Prefix-only (`term*`)**: fixes partial words but doesn't solve typos
- **BK-tree in-memory index**: faster fuzzy lookups at scale, but adds memory overhead and complexity — deferred unless Levenshtein proves too slow at the capped pool size

Alan-TheGentleman:
Useful idea, but it overlaps with the semantic-search proposal in #233 and the BM25 ranking in #241. Before a PR, we should define one search-improvement direction so these do not collide. Propose how trigram and fuzzy fit alongside (or instead of) those, with the scoring contract. Design-gated.

danielgap:
Suggesting we consolidate this into #233 and close it here, maintainer's call of course.

[...]

Issue #233, opened 2026-04-25
Add semantic search layer on top of FTS5 to improve mem_search recall


Basparin:
## Context

Empirical finding from Basparin/Lawen project, 900-observation DB, session 2026-04-25.

`mem_search` is currently a wrapper over SQLite + FTS5 (lexical/tokenized search). Natural-language queries with valid intent but disjoint tokens silently return 0 results despite the content existing.

## Reproducible evidence

Same DB, same content, ~5 minutes apart:

| Query | Hits | Content existed? |
|---|---|---|
| `workflow friction overuse optimization` | 0 | Yes — observations on hooks/bridge/CLAUDE.md |
| `tools rarely used skills inactive` | 0 | Yes — skill registry observation |
| `judgment-day ultrareview parallel review` | 0 | Yes — referenced in session summaries |
| `pulse statusline bridge` | 3 | **Same content as query #1, different tokens** |

## Impact

The value of the saved corpus is gated by the caller's ability to guess past-self's word choices. For agents (Claude orchestrators) writing across many sessions and weeks, this leads to silent recall failures — saved decisions become invisible to queries with semantically equivalent but lexically distinct phrasing.

## Proposal

Add an **optional** embeddings layer with **hybrid scoring**:
- FTS5 lexical match stays primary (deterministic, fast)
- Cosine similarity over embeddings augments — recovers synonymic / paraphrased queries
- Configurable weight (e.g. `--semantic-weight 0.3`) so users can tune lexical-vs-semantic mix
- Embedding model pluggable: local (`sentence-transformers`, ONNX) for offline, or API (Anthropic/OpenAI) for higher quality

## Acceptance criteria

Same observation recoverable via at least 3 distinct natural-language queries that share intent but no overlapping tokens.

## Backwards compatibility

Existing FTS5 behavior unchanged unless embeddings flag is enabled. Migration: re-embed existing observations on first opt-in run.

## Workaround documented (no upstream change)

For now Basparin's symbiosis repo carries the workaround: mandatory `topic_key` per evolving topic, multi-keyword retry before concluding absent. Tracking: https://github.com/Basparin/symbiosis/pull/13

## Related

Repo backend confirmed: `Persistent memory system for AI coding agents. Agent-agnostic Go binary with SQLite + FTS5` (README).

alexandervazquez98:
[...]

PR #475, opened 2026-06-10
feat(store,mcp): add match_mode option to Search and mem_search

### LINKED ISSUE #352: feat(store): mem_search returns 0 when no observation contains all query tokens

Basparin:
## 🔗 Linked Issue

Closes #352

---

## 🏷️ PR Type

- [ ] `type:bug` — Bug fix

---

## 📝 Summary

- Adds `match_mode: "all" | "any"` to `Store.Search()` and the `mem_search` MCP tool. Default (`""`/`"all"`) preserves today's FTS5 implicit-AND behavior exactly; `"any"` joins tokens with OR for broader recall, fixing the silent zero-result failure when no single observation contains every query token.
- The OR branch reuses the existing `sanitizeFTSCandidates` helper from `internal/store/relations.go` (already used by `FindCandidates`) — no new query-building logic.
- Invalid values error explicitly at both layers (store + MCP pre-validation) instead of silently falling back — this issue is precisely about silent failures, so the new knob refuses to fail silently itself.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | `MatchMode` field on `SearchOptions`; early validation in `Search()` (before the topic-key path); AND/OR branch at FTS query build |
| `internal/mcp/mcp.go` | `match_mode` param on the `mem_search` tool schema; pre-validation in `handleSearch` with a targeted error (no generic "Try simpler keywords" advice for a parameter mistake) |
| `internal/store/store_test.go` | 6 tests: default-AND preserved, explicit `"all"`, `"any"` recall (issue's synthetic repro: 3 observations with partial token overlap → 0 hits AND, 3 hits ANY), invalid value errors, single-token equivalence, empty-query behavior pinned |
| `internal/mcp/mcp_test.go` | 2 tests: `"any"` flows through the handler (asserts all 3 titles + count), invalid value surfaces as tool error without the misleading suffix |

Net: 4 files, +239/−10. HTTP server, CLI, and TUI intentionally untouched — the approved issue scopes store + MCP; propagating the flag there is a natural follow-up if wanted.

## 🧪 Test Plan

- [x] Unit tests pass locally: `go test ./...` (see note)
- [x] E2E tests pass locally: `go test -tags e2e ./internal/server/...` (see note)

Manual verification: reproduced the issue's exact synthetic scenario via the new store tests — `mem_search("auth compliance session")` → 0 hits with AND, 3 hits with `match_mode: "any"`.

Note (Windows environment): 16 tests across `cmd/engram` (cloud/serve/sync-status), `internal/setup`, `internal/sync`, and `internal/store` (`TestMigrate_Idempotent`, `TestNewErrorBranches`) fail locally on Windows. All 16 verified to fail identically on a clean checkout of `upstream/main` — environment issues unrelated to this change. The packages this PR touches (`internal/store` match_mode tests, `internal/mcp` full suite) pass.

---

## ✅ Contributor Checklist

- [x] I linked an approved issue above (`Closes #352`)
- [ ] I added exactly **one** `type:*` label to this PR — external contributors can't self-apply labels; requesting `type:feature`
- [x] I ran e2e tests locally: `go test -tags e2e ./internal/server/...`
- [x] Docs updated (if behavior changed) — checked: `mem_search` params aren't documented in README/docs, so there was nothing to update; the tool's self-describing MCP schema documents the new param

---

## 💬 Notes for Reviewers

- **Why error instead of silent fallback for invalid values**: this issue exists because zero-results failures were silent. A `match_mode: "or"` typo silently treated as `"all"` would reproduce the same class of bug one layer up.
- **Why `sanitizeFTSCandidates` reuse**: same package, already battle-tested by `FindCandidates`, identical quoting/injection surface as the existing AND path.
- Alternatives considered are in the issue body (auto-relax fallback, client-side validation); field evidence (149-query log, zero-hit rate scaling with token count) is in the issue comments.
- Related issues: #257 (trigram/fuzzy — different failure mode), #233 (semantic recall — design-gated). This PR is the syntactic-relaxation slice of that stack.

### DIFF
--- a/internal/mcp/mcp.go
+++ b/internal/mcp/mcp.go
@@ -280,6 +280,9 @@ func registerTools(srv *server.MCPServer, s *store.Store, cfg MCPConfig, allowli
 				mcp.WithString("scope",
 					mcp.Description("Filter by scope: project (default) or personal"),
 				),
+				mcp.WithString("match_mode",
+					mcp.Description("Token matching: \"all\" (default — every token must match, FTS5 AND) or \"any\" (any token matches — broader recall for multi-token queries). Any other value returns an error."),
+				),
 				mcp.WithNumber("limit",
 					mcp.Description("Max results (default: 10, max: 20)"),
 				),
@@ -903,8 +906,14 @@ func handleSearch(s *store.Store, cfg MCPConfig, activity *SessionActivity) serv
 		typ, _ := req.GetArguments()["type"].(string)
 		projectOverride, _ := req.GetArguments()["project"].(string)
 		scope, _ := req.GetArguments()["scope"].(string)
+		matchMode, _ := req.GetArguments()["match_mode"].(string)
 		allProjects := boolArg(req, "all_projects", false)
 		limit := intArg(req, "limit", 10)
+
+		// Validate match_mode before any project resolution or DB work.
+		if matchMode != "" && matchMode != "all" && matchMode != "any" {
+			return mcp.NewToolResultError(fmt.Sprintf("invalid match_mode %q: must be \"all\" or \"any\"", matchMode)), nil
+		}
 
 		// all_projects=true short-circuits project resolution: we search globally
 		// regardless of the project override or any auto-detected project. This
@@ -945,10 +954,11 @@ func handleSearch(s *store.Store, cfg MCPConfig, activity *SessionActivity) serv
 		activity.RecordToolCall(sessionID)
 
 		results, err := s.Search(query, store.SearchOptions{
-			Type:    typ,
-			Project: searchProject,
-			Scope:   scope,
-			Limit:   limit,
+			Type:      typ,
+			Project:   searchProject,
+			Scope:     scope,
+			Limit:     limit,
+			MatchMode: matchMode,
 		})
 		if err != nil {
 			return mcp.NewToolResultError(fmt.Sprintf("Search error: %s. Try simpler keywords.", err)), nil
--- a/internal/mcp/mcp_test.go
+++ b/internal/mcp/mcp_test.go
@@ -7088,3 +7088,90 @@ func TestHandleSearchLegacyMixedCaseProject(t *testing.T) {
 		t.Fatalf("expected search results for legacy project, got: %s", text)
 	}
 }
+
+// ─── match_mode MCP tests (issue #352) ──────────────────────────────────────
+
+// seedMCPMatchModeFixture creates a session and three observations with partial
+// token overlap — mirrors the store-level fixture.
+func seedMCPMatchModeFixture(t *testing.T, s *store.Store) {
+	t.Helper()
+	if err := s.CreateSession("s-mcp-matchmode", "engram", "/tmp"); err != nil {
+		t.Fatalf("create session: %v", err)
+	}
+	obs := []store.AddObservationParams{
+		{SessionID: "s-mcp-matchmode", Type: "decision", Title: "Auth session middleware", Content: "", Project: "engram", Scope: "project"},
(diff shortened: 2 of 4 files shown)


Alan-TheGentleman:
This is the right shape for #352: keep the current default and add explicit broader recall through `match_mode`. I added the `type:feature` label so the process gate can run. Next step is a code review after required checks are green.

Alan-TheGentleman (approved):
Deep review: SHIP. FTS5 match_mode is correct for both AND and ANY, no injection surface (all input goes through MATCH ? parameterized binding), backward compat is exact (empty/"all" unchanged), and build + store/mcp tests are green. Only low/pre-existing nits (interior-quote handling in sanitizeFTS is a separate follow-up). Branch is up to date with main.

PR #495, opened 2026-06-14
feat(store): derive memory confidence and blend it into search ranking

Makedayz:
## Problem

`Store.Search` orders results purely by FTS5/bm25 lexical relevance (`ORDER BY fts.rank`). The `duplicate_count`, `last_seen_at`, and `pinned` columns already exist on observations but have **no effect on retrieval** — a memory confirmed many times ranks identically to a one-off hit. And `mem_search` callers never see the evidence behind a result without a second `mem_get_observation` lookup.

## What this adds

A **derived** confidence signal (never stored, never asked of the model) used for ranking and surfacing:

- **`computeConfidence()`** — a `[0,1]` score from existing columns only: base `0.5`; `+0.08` per extra `duplicate_count` (capped `+0.40`); `+0.20` if `pinned`; `-0.30` when `review_after` marks the memory `needs_review`.
- **`rerankByConfidence()`** — blends min-max normalized lexical relevance (`0.7`) with confidence (`0.3`) for the FTS portion. `topic_key` direct hits (`topicRankBoost`) stay pinned at the front, untouched.
- **Candidate over-fetch** — when ranking is on, FTS fetches `limit * confidenceCandidateFactor` rows so a confirmed-but-lexically-weaker memory can surface instead of being cut by the bm25-only `LIMIT` before it is ever scored.
- **Surfacing** — `mem_search` now shows confidence label + score, `duplicate_count`, and `last_seen_at` in both the text and structured output. `SearchResult` gains computed `Confidence` / `ConfidenceLabel` fields.

## Design guarantees

- **No schema migration** — every input is an existing column.
- **No hallucination** — the number is computed from data, never estimated by the model.
- **Not bypassable** — lives in `Search()`, the data path of every `mem_search`.
- **Reversible** — gated behind `Config.ConfidenceRanking` (default `true`); legacy bm25-only path remains.

## Note on behavior

The confidence lift is proportional to how close lexical scores are: it breaks near-ties and gives a modest boost, but a wide relevance gap still wins (by design, given the `0.7/0.3` weights). With only 2 candidates min-max is degenerate (1.0 vs 0.0); the lift is meaningful within the realistic over-fetched pool.

## Tests

Adds `internal/store/confidence_test.go`: scoring cases, label tiers, re-rank lift within a graded pool, and topic-hit preservation. `gofmt`, `go vet`, `go build ./...` and `go test ./internal/store ./internal/mcp` all pass.

## Release Notes

* **New Features**
  * Search results now include confidence levels (confirmed, likely, tentative, unverified) based on observation history and patterns
  * Results can be reordered by confidence to prioritize more reliable matches (enabled by default)

* **Tests**
  * Added comprehensive test coverage for confidence scoring and result reordering logic

### DIFF
--- a/internal/mcp/mcp.go
+++ b/internal/mcp/mcp.go
@@ -1048,24 +1048,39 @@ func handleSearch(s *store.Store, cfg MCPConfig, activity *SessionActivity) serv
 			if r.State() == store.ObservationStateNeedsReview {
 				stateDisplay = " | state: needs_review"
 			}
-			fmt.Fprintf(&b, "[%d] #%d (%s) — %s\n    %s\n    %s%s | scope: %s%s\n",
+			// Surface derived confidence + the evidence behind it (part D), so the
+			// agent can judge trust at search time without a second lookup.
+			confidenceDisplay := fmt.Sprintf(" | confidence: %s (%.2f)", r.ConfidenceLabel, r.Confidence)
+			if r.DuplicateCount > 1 {
+				confidenceDisplay += fmt.Sprintf(" | confirmed %d×", r.DuplicateCount)
+			}
+			if r.LastSeenAt != nil && strings.TrimSpace(*r.LastSeenAt) != "" {
+				confidenceDisplay += fmt.Sprintf(" | last seen: %s", timeutil.FormatLocal(*r.LastSeenAt))
+			}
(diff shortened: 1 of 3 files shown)


Makedayz:
Closing — keeping this on my fork for now.

PR #494, opened 2026-06-14
fix(sync): export pre-existing relations missing from local chunks

forNerzul:
## 🔗 Linked Issue

Closes #353

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix
- [ ] `type:feature` — New feature
- [ ] `type:docs` — Documentation only
- [ ] `type:refactor` — Code refactoring (no behavior change)
- [ ] `type:chore` — Maintenance, dependencies, tooling
- [ ] `type:breaking-change` — Breaking change

---

## 📝 Summary

- Follow-up to #489. The local chunk sync writes relation mutations, but on **upgrade** it still dropped every pre-existing relation — the original #353 repro exported 0 of 35 relations.
- Filter relations by **chunk presence** instead of `lastChunkTime`: a relation is exported when it is absent from every prior chunk (brand-new *and* pre-existing) or was updated after the last chunk (re-judged relations still propagate).
- Round-trip restored: export 0 → 35, import into a fresh DB 0 → 35.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/sync/sync.go` | Replace `filterNewRelationMutations` (time-only) with `filterRelationMutationsForExport` (presence ∪ updated-since) plus `exportedRelationKeys`, which reads prior chunks via `Transport.ReadChunk` and collects already-exported relation keys. Missing chunks are skipped (idempotent re-export); corrupt chunks fail loudly. |
| `internal/sync/sync_test.go` | Add `…BackfillsRelationsCreatedBeforeLastChunk` (repro), `…ReexportsRelationUpdatedAfterLastChunk` (updated-since branch), `…FailsLoudlyOnCorruptPriorChunk`; fix `TestIncrementalRelationExport` so its prior chunk actually contains the relation it treats as already exported. |

## 🧪 Test Plan

- [x] Unit tests pass locally: `go test ./...`
- [ ] E2E tests pass locally: `go test -tags e2e ./internal/server/...` *(not run — change is confined to local chunk sync; full unit suite + `go vet` + `go build ./...` are green)*
- [x] Manually tested the affected functionality

Manual test: isolated copy of a real project's DB and `.engram` (nothing mutated). `engram sync` wrote a chunk with **35** relation mutations (was 0); importing it into a fresh DB restored `conflicts list` **Total: 35** (was 0).

---

## 🤖 Automated Checks

| Check | What it verifies | Status |
|-------|-----------------|--------|
| **Check Issue Reference** | PR body contains `Closes #N` | ⏳ |
[...]

Issue #353, opened 2026-05-11
fix(sync): local chunk sync silently drops memory_relations without warning


forNerzul:
### Pre-flight Checks
- [x] I have searched existing issues and this is not a duplicate
- [x] I understand this issue needs status:approved before a PR can be opened

### Bug Description
When running `engram sync --project <name>`, the local chunk written to `.engram/chunks/*.jsonl.gz` only contains `sessions`, `observations`, and `prompts`. Any `memory_relations` rows for the project are **silently omitted** from the chunk — no error, no warning, no log line. `engram sync --import` on a destination machine therefore restores observations without their semantic relation graph.

This violates the documented invariant in `docs/codebase/sync-and-cloud.md`:

> Business rule: **if sync is blocked, fail loudly and visibly. No silent drops.**

This is a **different code path** from #313, which addresses the cloud chunkcodec rejecting `relation`/`upsert` mutations loudly. The local sync path in `internal/sync/sync.go` doesn't reject anything — its chunk struct simply has no `Relations` field, and the export code never queries `memory_relations`.

### Steps to Reproduce
1. Create a project with at least one judged relation. Any flow that ends with `engram conflicts list --project <name>` showing one or more rows works.
2. Run local sync:
   ```
   cd <project-folder>
   engram sync --project <name>
   ```
3. Inspect the chunk:
   ```
   gzcat .engram/chunks/<id>.jsonl.gz | jq 'keys'
   ```
4. Confirm the DB still has the relations:
   ```
   engram conflicts list --project <name>
   ```

### Expected Behavior
Either of:

- **(a) Include relations in the chunk** — extend the chunk struct in `internal/sync/sync.go` with `Relations []store.Relation`, query the `memory_relations` table during export, and merge them on import. Matches the local-first source-of-truth model.
- **(b) Fail loudly if relations exist but cannot be exported** — at minimum, log a warning to stderr listing the count of skipped relations, per the documented "no silent drops" invariant. Acceptable as an interim guard until (a) lands.

### Actual Behavior
`engram sync` exits 0 and reports counts for sessions / observations / prompts only. The chunk is valid JSON but its keys are only `["observations", "prompts", "sessions"]`. The relations remain in `~/.engram/engram.db` and never make it into the versionable `.engram/chunks/` artifact, so they are unreachable from a fresh clone.

### Operating System
macOS

### Engram Version
1.15.10

### Agent / Client
Claude Code

### Shell
zsh

### Relevant Logs
```
[...]

Issue #313, opened 2026-05-04
fix(store,chunkcodec): cloud sync blocked by unsupported relation mutations


parraletz:
### Pre-flight Checks
- [x] I have searched existing issues and this is not a duplicate
- [x] I understand this issue needs status:approved before a PR can be opened

### Bug Description
The `memory_relations` feature (Phase 2) allows recording semantic relationships between observations. However, the cloud sync pipeline rejects `relation`/`upsert` mutations because the legacy-mutation evaluator and chunk canonicalizer do not recognize the `relation` entity.

### Steps to Reproduce
1. Create a project with pending relation mutations in `sync_mutations` (e.g., via `JudgeRelation` or `SaveRelation` while cloud sync is configured).
2. Run `engram sync --cloud --project <project>`.
3. Observe failure:
   - Before any upgrade repair: `upgrade_blocked_legacy_mutation_manual: unsupported legacy mutation "relation"/"upsert"`.
   - After repairing legacy mutations: `canonicalize cloud chunk: mutations[N]: unsupported mutation "relation"/"upsert"`.
   - Server push: `invalid push payload: mutations[N]: unsupported mutation "relation"/"upsert"`.

### Expected Behavior
- `engram cloud upgrade doctor` should classify relation mutations as `ready` or `repairable`, never `blocked` with manual action required.
- `engram sync --cloud` should export relation mutations in the canonical chunk format.
- The cloud server should accept and ingest relation mutations.

### Actual Behavior
Cloud sync is completely blocked for any project with pending relation mutations.

### Operating System
Linux (also affects macOS and Windows clients)

### Engram Version
1.15.4

### Agent / Client
OpenCode / Claude Code / any

### Relevant Logs
```
engram: cloud sync upgrade_blocked_legacy_mutation_manual: legacy mutation payloads require manual action before cloud sync for project "acme-challenge" (seq=7 entity=relation op=upsert): unsupported legacy mutation "relation"/"upsert"
```
```
canonicalize cloud chunk: mutations[6]: unsupported mutation "relation"/"upsert"
```

### Additional Context
- This affects both the client-side canonicalizer (`chunkcodec.go`) and the server-side chunk ingestion (`cloudserver.go` reuses the same chunk-codec).
- Local relation CRUD (`SaveRelation`, `JudgeRelation`) works fine; the issue is strictly in the cloud sync/export paths.

Alan-TheGentleman:
[...]

PR #497, opened 2026-06-14
fix(store): backfill relation sync mutations so relations reach cloud

### LINKED ISSUE #496: fix(store): relations missing a sync_mutations row never reach cloud (no relation backfill)
### 📋 Pre-flight Checks

### 📝 Bug Description

A relation can exist in the local `memory_relations` table with **no corresponding `sync_mutations` row**, and there is **no mechanism to ever create one**. Such relations are silently never replicated to Engram Cloud — even though the local chunk sync path (post-#489/#494) exports every non-orphaned relation directly from the table.

This is distinct from the already-resolved relation cloud issues, all of which operate on relation mutations that **already exist** in the journal:

- **#313 / #379** — cloud *rejected* enqueued relation mutations (chunkcodec/server). Fixed.
- **#383** — the legacy upgrade evaluator *blocked* enqueued relation mutations. Fixed (`SyncEntityRelation` is now whitelisted in `evaluateCloudUpgradeLegacyMutationTx`, `internal/store/store.go:1514`).
- **#340** — repairs *missing required fields* inside existing mutation payloads (session/observation/prompt). Does not create missing rows, and #383 itself notes it "does not call out the relation entity gap specifically".

None of these create a `sync_mutations` row for a relation that never had one. The cloud's own row-creation/repair mechanism, `backfillProjectSyncMutationsTx`, backfills sessions/observations/prompts but **never relations**.

### Root cause

Three confluent code paths, all verified against `engram 1.16.3`:

1. **No relation backfill.** `backfillProjectSyncMutationsTx` (`internal/store/store.go:4996`) calls only `backfillSessionSyncMutationsTx`, `backfillObservationSyncMutationsTx`, and `backfillPromptSyncMutationsTx`. `projectNeedsBackfill` (`store.go:5009`) only counts missing sessions/observations/prompts. There is **no** `backfillRelationSyncMutationsTx` anywhere in the repo. These run at startup (`New` → `repairEnrolledProjectSyncMutations`, `store.go:643`) and on enrollment/rename/merge — the exact "row exists in source table but not in journal" repair that relations are excluded from.

2. **`JudgeBySemantic` never enqueues.** `JudgeBySemantic` (`internal/store/relations.go:732`) inserts/updates `memory_relations` but never calls `enqueueSyncMutationTx`. The **only** relation enqueue in the codebase is `relations.go:687`, inside `JudgeRelation`. `JudgeBySemantic` is live: it backs `mem_compare` (`internal/server/server.go:1315`, `internal/mcp/mcp.go:2134`) and `ScanProject` (`relations.go:1400`). Every semantic/system verdict therefore lands in the table with no journal row.

3. **`JudgeRelation` skips the enqueue when the project is not enrolled.** `relations.go:653`: `if enrolled == 0 { return nil // not enrolled — no mutation enqueued }`. By contrast, `enqueueSyncMutationTx` (`store.go:5364`) does **not** gate on enrollment — observations/sessions/prompts are enqueued unconditionally and the backfill catches legacy ones. Relations have neither: judged-before-enrollment relations get no row, and enrolling later does not backfill them.

The net asymmetry: the **local** chunk export reads relations straight from `memory_relations` (query at `store.go:~3380`, used by `engram sync`), so local sync carries them regardless of how they were judged. The **cloud** path is journal-based and only ever carries what was enqueued — so the same table yields different replication outcomes for local vs cloud, and for `JudgeRelation` vs `JudgeBySemantic`.

### 🔄 Steps to Reproduce

**Path A — `JudgeBySemantic` (no upgrade required):**

1. Enroll a project in cloud sync.
2. Trigger a semantic verdict that is not `not_conflict` — e.g. an agent calls `mem_compare` on two observations, or `ScanProject` runs.
3. Inspect the journal:
   ```sql
   -- a row exists in memory_relations:
   SELECT sync_id, relation FROM memory_relations ORDER BY created_at DESC LIMIT 1;
   -- but no matching mutation was enqueued:
   SELECT COUNT(*) FROM sync_mutations WHERE entity='relation' AND entity_key='<that sync_id>';  -- 0
   ```
4. The relation is never pushed to cloud; `local` sync (`engram sync`) would export it.

**Path B — pre-enrollment `JudgeRelation`:**

1. With a project **not** enrolled, resolve a conflict candidate via `mem_judge` (→ `JudgeRelation`). `relations.go:653` returns without enqueuing.
2. Enroll the project. `repairEnrolledProjectSyncMutations` backfills sessions/observations/prompts but not the relation.
3. The relation is permanently absent from the cloud journal.

### ✅ Expected Behavior

- A relation present in `memory_relations` (non-orphaned) is eventually represented by a `sync_mutations` row and replicated to cloud, regardless of the judge path or enrollment ordering — matching how local chunk sync already treats relations and how the cloud already treats sessions/observations/prompts.
- Concretely: `backfillProjectSyncMutationsTx` / `projectNeedsBackfill` cover relations, and/or `JudgeBySemantic` enqueues, and/or `JudgeRelation` enqueues even when not-yet-enrolled (so backfill can pick it up later).

### ❌ Actual Behavior

Relations created via `JudgeBySemantic`, or via `JudgeRelation` before enrollment, have no `sync_mutations` row and no repair path. They never reach Engram Cloud and fail silently — there is no warning, and the local sync path replicates them, masking the loss until a cloud-only consumer is compared.

### Operating System

macOS

### Engram Version

1.16.3

### Agent / Client

Claude Code

### 📋 Relevant Logs

```text
# Verified by source inspection (engram 1.16.3):
internal/store/store.go:4996  backfillProjectSyncMutationsTx -> session/observation/prompt only
internal/store/store.go:5009  projectNeedsBackfill           -> counts session/observation/prompt only
internal/store/store.go:5364  enqueueSyncMutationTx          -> no enrollment gate (obs/sessions/prompts always enqueue)
internal/store/relations.go:653  JudgeRelation               -> "return nil // not enrolled — no mutation enqueued"
internal/store/relations.go:687  JudgeRelation               -> the ONLY relation enqueue in the codebase
internal/store/relations.go:732  JudgeBySemantic             -> inserts/updates memory_relations, never enqueues
grep -rn "backfillRelation" internal/ --include="*.go"  -> no match (no relation backfill exists)
```

### 💡 Additional Context

- Sibling of #353 (`fix(sync): local chunk sync silently drops memory_relations`) but for the **cloud** journal path. The local fix (PR #494) reads relations from the table on export; the cloud path cannot, because it syncs from the journal.
- Open question on intent: a Phase-1-era test documents `REQ-009: memory_relations are local-only in Phase 1` (`internal/mcp/mcp_conflict_loop_test.go:621`). Since `JudgeRelation` now enqueues relations and #313/#379/#383 explicitly enabled relation cloud sync, that constraint appears superseded — but please confirm whether full relation cloud replication is the intended end state before this is implemented.
- Suggested fix direction: add a `backfillRelationSyncMutationsTx` mirroring the observation backfill (scan `memory_relations` where `judgment_status != orphaned` and no matching `sync_mutations` row for the enrolled project), include relations in `projectNeedsBackfill`, and have `JudgeBySemantic` enqueue on the enrolled path (or rely on the new backfill). Add regression coverage analogous to the #353 export test (`internal/sync/sync_test.go:693`).
- Workaround today: none clean for cloud. `engram sync` (local chunk) does replicate these relations, so cross-machine continuity via local sync is unaffected; only cloud replication loses them.

Alan-TheGentleman:
Closes #496

## PR Type
- [x] `type:bug`

## Summary

Cloud-journal sibling of #353 (PR #494, which fixed the **local chunk** path). A relation in `memory_relations` could exist with **no** `sync_mutations` row, with no mechanism to ever create one, so it never replicated to Engram Cloud (journal-based) and failed silently. Three confluent gaps, all addressed:

1. **No relation backfill.** `backfillProjectSyncMutationsTx` only backfilled sessions/observations/prompts. Added `backfillRelationSyncMutationsTx`, mirroring `backfillObservationSyncMutationsTx` (two-phase collect-then-insert, non-orphaned only, `NOT EXISTS` guard against the cloud target key). Wired into `backfillProjectSyncMutationsTx`, so it runs on startup repair, enrollment, rename, and merge.
2. **`projectNeedsBackfill` ignored relations.** Added a 4th COUNT matching the backfill SELECT exactly, so the startup fast-path detects the gap.
3. **`JudgeBySemantic` never enqueued.** It backs `mem_compare` and `ScanProject`; every semantic verdict landed in `memory_relations` with no journal row. It now enqueues on the **enrolled** path, mirroring `JudgeRelation`s enrollment gate. Unenrolled stays a no-op and is caught by the backfill on later enrollment.

## REQ-009 intent (the open question in the issue)

Confirmed superseded. `#313/#379/#383` already enabled relation cloud sync (`JudgeRelation` already enqueued). The `TestConflictLoop_SyncRegression` / `assertNoRelationSyncMutations` guard still passes and is kept — but its comment was updated: it guards the **enrollment gate** (an *unenrolled* project must never enqueue), not a blanket "local-only" rule.

## Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | `backfillRelationSyncMutationsTx` (new), call from `backfillProjectSyncMutationsTx`, relation COUNT in `projectNeedsBackfill`, accurate doc comment on `CountRelationSyncMutations` |
| `internal/store/relations.go` | `JudgeBySemantic` enrollment-gated enqueue |
| `internal/store/relation_backfill_test.go` | New: backfill creates/excludes-orphaned/idempotent, `projectNeedsBackfill` cases, `JudgeBySemantic` enrolled/unenrolled/upsert, and `TestEnrollProject_BackfillsPreExistingRelations` (real pre-enrollment → EnrollProject → backfill end-to-end) |
| `internal/mcp/mcp_conflict_loop_test.go` | Updated stale REQ-009 comments to reflect the enrollment-gate invariant |

## Test Plan
- [x] `go build ./...`, `go vet ./...`
- [x] `go test ./...` — all packages green (Strict TDD: tests written red-first)
- [ ] Validated against the real cloud repro from #496 — **requested from the issue author below**

## Note for @forNerzul

You filed this and described the exact fix direction (mirror the observation/session/prompt backfill for relations) plus the REQ-009 intent question. This implements that. Since you have the real cloud-enrolled `cognicion-modular` setup that reproduces the silent drop, could you validate it against your scenario? Build from this branch:

```bash
git fetch origin fix/cloud-relation-backfill
git checkout fix/cloud-relation-backfill
go build -o ./engram ./cmd/engram
```

Then on an **enrolled** project: trigger a `JudgeBySemantic` verdict (`mem_compare` / `ScanProject`) or enroll a project that already has relations, and confirm a `relation` row appears in `sync_mutations` and reaches cloud (before this it was 0). Backfill path: `engram` startup on an enrolled project with pre-existing relations should now enqueue them.

If your own fix diverged in approach, flag it here and we will reconcile before merge.

## Release Notes

* **Bug Fixes**
  * Improved cloud sync backfill/repair for relations, including idempotent creation of missing relation upsert sync mutations.
  * Enforced enrollment-gated relation syncing: relations won’t enqueue sync while unenrolled, but are backfilled after enrollment.
  * Hardened cross-project validation and improved project resolution using session fallback; orphaned and pending/unjudged relations are excluded.

* **Tests**
  * Added and expanded automated coverage for relation sync backfill/repair, enrollment-gated enqueue behavior, payload project fallback, and cross-project guard scenarios.

### DIFF
--- a/internal/mcp/mcp_conflict_loop_test.go
+++ b/internal/mcp/mcp_conflict_loop_test.go
@@ -421,15 +421,21 @@ func TestConflictLoop_Orphaning(t *testing.T) {
 	}
 }
 
-// ─── G.4 — Sync regression test ─────────────────────────────────────────────
+// ─── G.4 — Sync enrollment-gate regression test ─────────────────────────────
 //
-// Verifies that memory_relations rows are NOT inserted into sync_mutations
-// (REQ-009: local-only relations in Phase 1). This is the regression guard.
+// Verifies that an UNENROLLED project never enqueues relation sync mutations.
+// The enqueue paths in JudgeRelation and JudgeBySemantic are guarded by an
+// enrollment check — this test is the regression guard for that gate.
 //
-// The assertion is entity-level, not count-level: no row in sync_mutations
-// should have entity = 'memory_relations' or 'relation' after any conflict
-// operation. Sessions and observations still produce their own sync mutations
-// as expected.
+// Note: relation cloud sync IS intentional for enrolled projects (#313/#379/
+// #383 enabled it; #496 extends it with backfill). This test uses an unenrolled
+// store (newMCPTestStore does not call EnrollProject), so the relation count
+// must remain zero — not because relations are local-only, but because the
+// enrollment gate must hold.
+//
+// The assertion is entity-level: no row in sync_mutations should have
+// entity = 'relation' after conflict operations on an unenrolled project.
+// Sessions and observations still produce their own sync mutations as expected.
 func TestConflictLoop_SyncRegression(t *testing.T) {
 	s := newMCPTestStore(t)
 	saveH := handleSave(s, MCPConfig{}, NewSessionActivity(10*time.Minute))
@@ -479,13 +485,14 @@ func TestConflictLoop_SyncRegression(t *testing.T) {
 			t.Fatalf("G.4 mem_judge: err=%v isError=%v", err, judgeRes.IsError)
 		}
 	} else {
-		t.Logf("G.4: no candidates returned after save B (FTS similarity below floor); judge step skipped — relation-entity assertion still covers REQ-009")
+		t.Logf("G.4: no candidates returned after save B (FTS similarity below floor); judge step skipped — enrollment-gate assertion still covers the unenrolled guard")
 	}
 
-	// ── Step 3: Assert no 'memory_relations' entity in sync_mutations. ────────
-	// REQ-009: memory_relations must not appear in the sync wire at all.
-	// Sessions ('session') and observations ('observation') are expected;
-	// any other entity type for relations would be a violation.
+	// ── Step 3: Assert no relation entity in sync_mutations (enrollment gate). ──
+	// This store is UNENROLLED — the enrollment gate in JudgeRelation /
+	// JudgeBySemantic must prevent any relation sync_mutations row from being
+	// written. Sessions ('session') and observations ('observation') are expected;
+	// a relation-entity row here would mean the enrollment guard was bypassed.
 	assertNoRelationSyncMutations(t, s)
 
 	// ── Step 4: Verify observation sync payloads exclude decay fields. ────────
@@ -619,9 +626,14 @@ func parseEnvelope(t *testing.T, label string, res *mcppkg.CallToolResult) map[s
 }
 
 // assertNoRelationSyncMutations verifies that NO sync_mutations rows reference
-// relation entities. REQ-009: memory_relations are local-only in Phase 1.
-// Sessions ('session') and observations ('observation') are expected entities;
-// any relation-entity row would be a regression.
+// relation entities on an UNENROLLED store. The enrollment gate in
+// JudgeRelation / JudgeBySemantic must prevent any relation row from being
+// written when the project is not enrolled. Sessions ('session') and
+// observations ('observation') are expected entities; a relation-entity row
+// in an unenrolled context means the enrollment guard was bypassed.
+//
+// Note: relation sync mutations ARE valid for enrolled projects (#313/#379/
+// #383/#496). This helper is intentionally scoped to the unenrolled test in G.4.
 func assertNoRelationSyncMutations(t *testing.T, s *store.Store) {
 	t.Helper()
 	count, err := s.CountRelationSyncMutations()
@@ -629,7 +641,7 @@ func assertNoRelationSyncMutations(t *testing.T, s *store.Store) {
 		t.Fatalf("G.4 CountRelationSyncMutations: %v", err)
 	}
 	if count > 0 {
-		t.Errorf("G.4 REQ-009 violated: found %d sync_mutations row(s) with relation entity — memory_relations must NOT sync in Phase 1", count)
+		t.Errorf("G.4 enrollment-gate violated: found %d sync_mutations row(s) with relation entity on unenrolled project — enrollment gate must prevent relation sync mutations", count)
 	}
 }
 
--- a/internal/store/relation_backfill_test.go
+++ b/internal/store/relation_backfill_test.go
@@ -639,3 +639,84 @@ func TestCrossProjectGuard_SessionFallback_AllowsSameSessionProject(t *testing.T
 		t.Error("JudgeBySemantic: expected non-empty relSyncID for same-session-project pair")
 	}
 }
+
+// ─── Test 8: JudgeRelation uses session-fallback for project derivation ────────
+
+// TestJudgeRelation_UsesSessionFallback_ForProject verifies that JudgeRelation
+// derives the Project for the enqueued sync mutation via the session-fallback
+// (coalesce(obs.project, session.project)), not from observations.project alone.
+//
+// Scenario: project P is enrolled; a session exists in P; two observations are
+// created in that session with BLANK observations.project (project lives only on
+// the session). A pending relation is saved between them. Calling JudgeRelation
+// must enqueue a sync_mutations row whose payload.project is "P", not "".
+//
+// Without the fix, srcProject="" → enrollCheckProject="" → enrolled=0 → return nil
+// (no mutation row). After the fix, the session-fallback resolves srcProject="P",
+// enrollCheckProject="P", enrolled=1, and the mutation is written immediately.
+func TestJudgeRelation_UsesSessionFallback_ForProject(t *testing.T) {
+	s := newTestStore(t)
+
+	// Create a session with an explicit project and enroll it.
+	if err := s.CreateSession("ses-jr-sf", "proj-jr-sf", "/tmp/jr-sf"); err != nil {
+		t.Fatalf("CreateSession: %v", err)
+	}
+	if err := s.EnrollProject("proj-jr-sf"); err != nil {
+		t.Fatalf("EnrollProject: %v", err)
+	}
+
+	// Create observations with BLANK observations.project — project lives only on
+	// the session. This simulates observations ingested before the project column
+	// was populated, which is the exact gap described in the CodeRabbit finding.
+	_, srcSyncID := addTestObsSession(t, s, "ses-jr-sf", "JR SF source obs", "decision", "", "project")
+	_, tgtSyncID := addTestObsSession(t, s, "ses-jr-sf", "JR SF target obs", "decision", "", "project")
+
+	// Save a pending relation (the precursor step before JudgeRelation).
+	relSyncID := newSyncID("rel-jr-sf")
+	if _, err := s.SaveRelation(SaveRelationParams{
+		SyncID:   relSyncID,
+		SourceID: srcSyncID,
+		TargetID: tgtSyncID,
+	}); err != nil {
+		t.Fatalf("SaveRelation: %v", err)
+	}
+
+	// Pre-condition: no mutation row yet (SaveRelation does not enqueue).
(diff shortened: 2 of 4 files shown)


Alan-TheGentleman:
@forNerzul concrete review request: this is the #496 fix ready for your eyes. GitHub will not let me add you as a formal reviewer (external/fork contributor), so flagging it here directly.

Could you review the diff and validate it against your cloud-enrolled `cognicion-modular` repro? Build from the branch:

```bash
git fetch origin fix/cloud-relation-backfill
git checkout fix/cloud-relation-backfill
go build -o ./engram ./cmd/engram
```

What to confirm:
1. On an **enrolled** project, a `JudgeBySemantic` verdict (`mem_compare` / `ScanProject`) now creates a `relation` row in `sync_mutations` (was 0).
2. **Backfill**: starting `engram` on an enrolled project that already had relations enqueues them (`backfillRelationSyncMutationsTx` via the enroll/startup-repair path).
3. Orphaned relations stay excluded.

If your own intended fix diverged in approach (you described mirroring the observation/session/prompt backfill — which is what this does), call it out and we will reconcile before merge. Trying to land this soon, so your validation is the last gate.

Alan-TheGentleman:
Addressed all CodeRabbit + Copilot findings in `e83d2d0`. Each was verified against the actual cloud validation before changing anything — and all were valid. The cloud path enforces the requirement at three layers: `chunkcodec.go:484-489` (client) and `cloudserver/mutations.go:303-333` reject relation mutations missing `marked_by_actor`/`marked_by_kind`, and `mutations.go:92-100` rejects `project == ""`.

- **Copilot (store.go:5064 / 5423) — backfill enqueued `pending` rows.** Fixed. Both `backfillRelationSyncMutationsTx` and the `projectNeedsBackfill` COUNT now use an identical predicate: `judgment_status NOT IN (orphaned, pending) AND marked_by_actor != "" AND marked_by_kind != ""`. This prevents enqueuing candidate rows that would 400 the whole cloud batch. Keeping the two predicates identical avoids fast-path desync.
- **Copilot (relations.go:827) — `JudgeBySemantic` project from observations only.** Fixed. It now resolves project via `coalesce(nullif(o.project,""), s.project, "")` with a session join, matching the backfill SELECT, so the enqueued payload never carries an empty project when the session has one.
- **CodeRabbit (relations.go:865) — missing REQ-011 warning.** Fixed. `JudgeBySemantic` now logs the same `project=""` WARNING breadcrumb as `JudgeRelation`.
- **Copilot (relation_backfill_test.go:49) — fixture lacked marked_by_*.** Fixed. New `insertJudgedRelationDirect` helper populates `marked_by_actor`/`marked_by_kind`; the bare `insertRelationDirect` is kept for orphaned/pending fixtures.

Added two regression tests: `TestBackfillRelationSyncMutations_SkipsPending` (pending rows are neither backfilled nor flagged by `projectNeedsBackfill`) and `TestJudgeBySemantic_UsesSessionFallback_ForProject` (enqueued payload carries the session-resolved project). `go build`, `go vet`, and `go test ./...` all green.

Alan-TheGentleman:
Both CodeRabbit findings addressed in `e1ed602` — verified against the code first, both were valid (the cross-project one was a regression I introduced when adding the session fallback).

- **Cross-project guard bypass (relations.go:701-714) — Major.** `validateCrossProjectGuard` compared `observations.project` only, while `JudgeBySemantic` now resolves project with a session fallback. So observations with blank `observations.project` but sessions in *different* projects slipped past the guard and got enqueued. Fixed at the source: the shared guard now uses the same `coalesce(nullif(o.project,""), s.project, "")` LEFT JOIN sessions form, so it catches cross-project relations regardless of whether `observations.project` is populated — and it fixes both `JudgeRelation` and `JudgeBySemantic` in one place rather than a duplicated inline check. REQ-011 "both empty allowed" semantics preserved.
- **REQ-011 warning placement (JudgeBySemantic) — Minor.** Moved the `project=""` WARNING log to *after* the `if enrolled == 0 { return }` gate so it only fires when the relation will actually be enqueued, matching `JudgeRelation`s ordering. No more false-positive breadcrumbs on the not-enrolled path.

Added `TestCrossProjectGuard_SessionFallback_RejectsDifferentSessionProjects` (fails before the guard fix, passes after; covers both judge paths) and `TestCrossProjectGuard_SessionFallback_AllowsSameSessionProject` (guard not over-tightened). `go build`, `go vet`, `go test ./...` all green.

Alan-TheGentleman:
Good catch — addressed in `9225c57`. Verified it was reachable: `JudgeRelation` derived `srcProject`/`tgtProject` from `observations.project` alone, so a relation judged on an enrolled project whose observations have blank project but whose session carries it resolved `enrollCheckProject=""` and returned without enqueuing — recreating the replication gap for the `JudgeRelation` path (only caught later by backfill-on-startup, not immediately).

Fixed by giving `JudgeRelation` the same session-fallback lookup already used by `JudgeBySemantic` and `validateCrossProjectGuard`:
```sql
SELECT coalesce(nullif(o.project,""), s.project, "") FROM observations o LEFT JOIN sessions s ON s.id = o.session_id WHERE o.sync_id = ?
```
Now `enrollCheckProject` and the payload `Project` carry the session-resolved project, so enrolled-project relations enqueue immediately and consistently across both judge paths. REQ-011 "source genuinely missing → project=\"\"" semantics preserved.

Added `TestJudgeRelation_UsesSessionFallback_ForProject` (fails before: 0 mutation rows; passes after: row exists with project resolved from the session). `go test ./...` all green.

PR #526, opened 2026-06-20
feat(store): replace FTS5 default ranking with weighted BM25

daniel-baf:
Closes #241

## PR Type

- [ ] `type:bug` — Bug fix
- [x] `type:feature` — New feature
- [ ] `type:docs` — Documentation only
- [ ] `type:refactor` — Code refactoring (no behavior change)
- [ ] `type:chore` — Maintenance, dependencies, tooling
- [ ] `type:breaking-change` — Breaking change

## Summary

- Replaced SQLite FTS5 default ranking (`fts.rank`) in `Store.Search` query with a weighted BM25 query: `bm25(observations_fts, 5.0, 1.0, 0.0, 0.0, 0.0, 3.0)`.
- Prioritizes matches in `title` (5.0 weight) and `topic_key` (3.0 weight) over lower-relevance columns like `tool_name` or `type`.
- Added unit test coverage validating that title matches rank higher than content matches.

## Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | Replace `fts.rank` with `bm25(...)` query and update `ORDER BY` clause |
| `internal/store/store_test.go` | Add `TestSearch_WeightedBM25Ranking` unit test |

## Test Plan

- [x] Unit tests pass locally: `go test ./...`
- [x] E2E tests pass locally: `go test -tags e2e ./internal/server/...`
- [x] Manually tested the affected functionality (verified via unit test)

## Contributor Checklist

- [x] I linked an approved issue above (`Closes #241`)
- [x] I added exactly **one** `type:*` label to this PR
- [x] I ran unit tests locally
- [x] I ran e2e tests locally
- [x] Docs updated (no behavior change needing doc changes)
- [x] Commits follow conventional commits format
- [x] No `Co-Authored-By` trailers in commits

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search result ranking so results are ordered using a more accurate BM25-based score, better prioritizing matches based on which observation fields contain the query.
* **Tests**
  * Added coverage to verify weighted BM25 ranking behavior, ensuring title matches are ranked ahead of content matches.

PR #537, opened 2026-06-23
fix(store): support CJK search with trigram FTS

### LINKED ISSUE #140: [Feature Request] Built-in CJK (Japanese/Chinese/Korean) support for FTS5 search
## Problem

The default FTS5 tokenizer (`unicode61`) cannot segment CJK (Chinese/Japanese/Korean) text into searchable tokens. This means `mem_search` returns 0 results for any CJK query, even when matching observations exist in the database.

**Environment:**
- engram v1.11.0 (macOS, Apple Silicon)
- Claude Code MCP plugin (`--tools=agent`)

**Steps to reproduce:**
1. Save an observation with Japanese text: `engram save -t "test" -c "サンドボックス修正テスト"`
2. Search for it: `engram search "サンドボックス"`
3. **Expected:** Returns the saved observation
4. **Actual:** Returns 0 results

## Root Cause

SQLite FTS5's `unicode61` tokenizer splits text on Unicode word boundaries, which works for space-delimited languages (English, etc.) but fails for CJK scripts that don't use spaces between words.

## Workaround (confirmed working)

Manually recreating the FTS tables with the `trigram` tokenizer resolves the issue:

```sql
-- Backup and recreate observations FTS
DROP TABLE IF EXISTS observations_fts;
CREATE VIRTUAL TABLE observations_fts USING fts5(
    title, content, tokenize='trigram'
);
INSERT INTO observations_fts(rowid, title, content)
    SELECT id, title, content FROM observations;

-- Same for prompts FTS
DROP TABLE IF EXISTS prompts_fts;
CREATE VIRTUAL TABLE prompts_fts USING fts5(
    content, tokenize='trigram'
);
INSERT INTO prompts_fts(rowid, content)
    SELECT id, content FROM prompts;
```

After this change, both Japanese and English searches work correctly.

## Proposal

Consider one of:

1. **Default to `trigram` tokenizer** — Works for all scripts (Latin, CJK, Cyrillic, etc.) with minimal performance trade-off for typical engram database sizes
2. **Auto-detect CJK content** — Switch to trigram tokenizer when CJK characters are detected in the database
3. **Configurable tokenizer** — Add a `--fts-tokenizer` flag to `engram setup` or a config option

Option 1 (default trigram) seems simplest and most robust, since engram databases are typically small (thousands of rows, not millions) where trigram overhead is negligible.

## Note

- `brew upgrade engram` resets the FTS tables to `unicode61`, requiring manual re-application of the trigram fix
- Related issue: #99 (FTS5 trigram tokenizer delete bug) — the trigram tokenizer may need additional testing for UPDATE/DELETE operations

daniel-baf:
## Linked Issue

Closes #140

---

## PR Type

- [x] `type:bug` — Bug fix
- [ ] `type:chore` — Maintenance/tooling

---

## Summary

- Switch local observation and prompt FTS5 tables to `tokenize='trigram'` so Japanese, Chinese, and Korean substring search works.
- Rebuild legacy/non-trigram FTS tables during migration (one-time `DROP` + reinsert), with a log line so large DBs know why startup paused.
- Preserve short-token searches such as `v2` with a bounded `LIKE` fallback **only when every query term is under three characters** (mixed queries like `go to prod` stay on the FTS/bm25 path).
- Keep standard FTS `'delete'`/`insert` triggers (same shape as `main`). Soft-delete trigger gating (`WHERE new.deleted_at IS NULL`) is intentionally deferred to a follow-up PR.

## Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | Trigram FTS schema/migration repair, short-token LIKE fallback (all-short only), preserves weighted `bm25(...)` on the FTS branch. |
| `internal/store/store_test.go` | CJK search, legacy migration, trigger update/delete, short-token + mixed-query regression tests. |
| `CHANGELOG.md` | Documents the memory search fix. |
| `docs/ARCHITECTURE.md` | Updates topic-key wording that assumed word-boundary tokenization. |
| `docs/TEAM-USAGE.md` | Clarifies that CJK substrings are searchable but search is not translation. |

## Test Plan

- [x] Targeted store tests: `go test ./internal/store -run 'TestSearch|Trigram|FTS'`
- [ ] Full unit suite locally: `go test ./...`
- [ ] E2E tests locally: `go test -tags e2e ./internal/server/...`

---

## Contributor Checklist

- [x] I linked an approved issue above (`Closes #140`)

---

## Notes for Reviewers

- Rebased/merged onto current `main` after #526 (bm25) and #586 (sanitizeFTS quotes).
- FTS branch preserves `bm25(observations_fts, 5.0, 1.0, 0.0, 0.0, 0.0, 3.0)`.
- Soft-delete FTS trigger gating will be a separate PR, as requested.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved substring search for Japanese, Chinese, and Korean across CLI, MCP, HTTP, and TUI using trigram-based FTS5 indexing.
  * Restored reliable search for short tokens (e.g., `v2`) via a bounded `LIKE` fallback when queries are very short.
  * Strengthened migration to consistently rebuild and repair trigram search indexes and related triggers.
* **Documentation**
  * Clarified shared-memory language strategy: FTS5 supports same-script substring searching, but isn’t a translation layer.
* **Tests**
  * Added end-to-end coverage for CJK substring matching, short-token fallback behavior, and migration/trigger correctness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,11 @@ Breaking changes are always marked with a `type:breaking-change` label and docum
 
 - **fix(cloud):** make chunk and mutation push payload limits configurable with `ENGRAM_CLOUD_MAX_PUSH_BYTES` while preserving the 8 MiB default.
 
+### Memory search
+
+- **fix(store):** use SQLite FTS5 trigram indexes for observations and prompts so Japanese, Chinese, and Korean substring searches work through CLI, MCP, HTTP, and TUI search.
+- **fix(store):** preserve short-token searches such as `v2` with a bounded `LIKE` fallback when every query term is under three characters (trigram indexes only match terms with at least three characters).
+
 ### Cloud user token management (`cloud-user-token-management`)
 
 - **feat(cloud):** add principal/human-user/token/project-grant/audit storage foundation (`cloud_principals`, `cloud_human_users`, `cloud_principal_tokens`, `cloud_project_grants`, `cloud_auth_audit_log`) alongside existing sync tables.
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -126,15 +126,15 @@ Examples:
 - `pattern/error-handling-convention`
 - `config/ci-environment`
 
-**Why this format?** SQLite FTS5 tokenises on word boundaries. Lowercase kebab-case ensures the key fragments are individually searchable and do not create unexpected FTS5 token splits.
+**Why this format?** Lowercase kebab-case keeps topic fragments readable, predictable, and searchable across Engram's SQLite FTS5 index.
 
 **Anti-patterns to avoid:**
 
 | Anti-pattern | Problem | Correct form |
 |---|---|---|
-| `authModel` | camelCase breaks FTS5 tokenisation | `architecture/auth-model` |
+| `authModel` | camelCase is harder to scan and match consistently | `architecture/auth-model` |
 | `auth model` | spaces create accidental multi-token keys | `architecture/auth-model` |
-| `ARCHITECTURE/AUTH` | uppercase is inconsistent with FTS5 normalisation | `architecture/auth-model` |
+| `ARCHITECTURE/AUTH` | uppercase is inconsistent with Engram's topic-key normalisation | `architecture/auth-model` |
 | `auth/model/v2/final` | more than 2 levels — use `v2` in the description | `architecture/auth-model-v2` |
 | `bugfix` | no slash — looks like a family with no description | `bug/auth-nil-panic` |
 
--- a/docs/TEAM-USAGE.md
+++ b/docs/TEAM-USAGE.md
@@ -76,7 +76,7 @@ worry when you see them on tables with >100k rows.
 
 ## Language strategy for shared memory
 
-FTS5, the SQLite full-text search engine Engram uses, is language-agnostic but not multilingual. A search in English will not match an observation saved in Spanish, and vice versa. For a globally distributed team, this matters: if every developer saves project-scope memories in their native language, the shared knowledge base fragments and search stops working as a team tool.
+FTS5, the SQLite full-text search engine Engram uses, is language-agnostic but not a translation layer. Engram indexes observations and prompts with trigram search, so Japanese, Chinese, and Korean substrings are searchable, but a search in English still will not match an observation saved only in Spanish, Russian, Japanese, or another language. For a globally distributed team, this matters: if every developer saves project-scope memories in their native language, the shared knowledge base fragments and search stops working as a team tool.
 
 ### Convention
 
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -959,9 +959,6 @@ func (s *Store) migrate() error {
 	if err := s.addColumnIfNotExists("user_prompts", "sync_id", "TEXT"); err != nil {
 		return err
 	}
-	if err := s.ensureFTSTrigramTables(); err != nil {
-		return err
-	}
 
 	if _, err := s.execHook(s.db, `
 		CREATE INDEX IF NOT EXISTS idx_obs_scope ON observations(scope);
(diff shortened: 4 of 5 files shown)


daniel-baf:
Reviewer note: this PR overlaps with #526 (`feat(store): replace FTS5 default ranking with weighted BM25`) in `Store.Search`.

If #526 merges first, this PR should be rebased and should preserve the weighted BM25 expression in the FTS branch. If this PR merges first, #526 should rebase and apply BM25 only inside the FTS branch, not the short-token LIKE fallback branch. I left the detailed merge-order note on #526: https://github.com/Gentleman-Programming/engram/pull/526#issuecomment-4775607961

Alan-TheGentleman (changes requested):
The CJK problem is real and the test coverage here is the best of the current FTS PRs — legacy migration, trigger repair, update/delete, and short-token regression are all covered.

Three things before this can land:

- [ ] **The description doesn't match the diff.** It says "replace FTS delete-command triggers with rebuild triggers", but the diff keeps the `'delete'` command triggers and only adds `WHEN` guards — `rebuild` appears only in two error strings. So the stated mitigation for the trigram delete bug isn't actually implemented. Either implement it or correct the description.
- [ ] **The LIKE fallback degrades English, not just CJK.** `shouldUseTrigramLikeFallback` returns true if *any* term is under 3 runes, so a query like `go to prod` drops the whole search to `LIKE '%…%'` over a six-column concatenation — unindexed full scan, `0.0 as rank`, ordered by `updated_at` instead of relevance. Please narrow it to per-term, or only when *all* terms are short.
- [ ] **Split the soft-delete trigger fix** (`WHERE new.deleted_at IS NULL`) into its own PR — it's a separate correctness issue from CJK tokenization.
- [ ] Rebase on `main` to resolve conflicts.

On merge order: please rebase **after** #586 and #526 land. Note that #586 is still required even with this PR — your LIKE fallback uses `searchTerms()` and bypasses `sanitizeFTS`, but `foo"bar` is 7 runes so it still takes the FTS path and still crashes. When you rebase over #526, preserve the `bm25(...)` expression in the FTS branch.

Also worth flagging: `ensureObservationsFTSTrigram` does a `DROP TABLE` + reinsert for every existing DB. No data loss since these are external-content tables, but it's a silent blocking O(N) rebuild inside `migrate()` — consider logging it so users on large DBs know why startup paused.

daniel-vantumst:
Rebased onto latest main (linear history now, no merge commits) and addressed the review. Summary of where each item landed:

Blocking items
- [x] Description ↔ diff mismatch — collapsed the messy branch history into two clean commits; the CJK commit message now matches the diff exactly. Triggers keep main's standard 'delete'/insert shape (no WHEN guards); rebuild only appears in the one-time migration log/error strings.
- [x] LIKE fallback degrading English — shouldUseTrigramLikeFallback now returns true only when every query term is under 3 runes. go to prod → prod is 4 → stays on the FTS/bm25 path with real ranking.
- [x] Split soft-delete trigger gating — the WHERE new.deleted_at IS NULL gating is out of this PR, deferred to a follow-up (noted in the trigger code).
- [x] Rebase — done as a true rebase onto current main, not a merge.

Merge-order notes
- Rebased on top of #586 and #526. bm25(observations_fts, 5.0, 1.0, 0.0, 0.0, 0.0, 3.0) preserved on the FTS branch, and the FTS path goes through sanitizeFTS so the foo"bar case is covered.
- [x] The O(N) DROP+reinsert rebuild inside migrate() now logs a one-time line so large-DB users know why startup paused.

Also fixed (regression not in the original review)
While validating, TestNewMigratesLegacyUserPromptsSyncIDSchema failed on this branch but passed on main — the trigram migration corrupted the DB (database disk image is malformed) on legacy DBs. Root cause: FTS triggers were being created before the migration backfill UPDATEs, so those UPDATEs fired an external-content 'delete' against a still-empty index. Fixed by creating the triggers after the backfills (mirrors main's ordering), and wrapped both FTS rebuilds in a transaction so a mid-sequence failure can't leave an empty FTS table.

Full internal/store suite passes. Force-pushed after the rebase (commits must appear later).

Still to come (separate PR): the soft-delete trigger gating (old.deleted_at IS NULL), as agreed.

dnlrsls:
Thank you, Daniel, for the thoughtful CJK search work and for the effort through rebases and review.

Our investigation found that the documented manual workaround creates a \prompts_fts\ schema shape that the migration cannot safely upgrade. Maintainers will continue with a maintainer-owned root fix, so we are closing this PR.

We appreciate your contribution and the care you put into it.

PR #540, opened 2026-06-24
fix(store,mcp): OR-match FTS5 queries and surface recent session activity

totakid420:
## Summary

Two retrieval bugs make `mem_search` / `mem_context` unreliable for the way agents actually use them. Both are fixed here with regression tests. The two fixes are independent — happy to split into separate PRs if you prefer.

### 1. `sanitizeFTS` used an implicit AND between terms

`sanitizeFTS` wrapped each term in quotes and joined them with a bare space. FTS5 treats `"a" "b" "c"` (space-separated, no operator) as an **implicit AND**, so a multi-word natural-language query required *every* term to appear in a single observation. Descriptive queries of 5–9 words returned nothing.

```
Search("engram obsidian bridge sessions not saved")  →  0 results (pre-fix)
```

**Fix:** join terms with `OR`. The BM25 rank (`ORDER BY fts.rank`, already present in both `Store.Search` and `Store.SearchPrompts`) ranks observations matching more terms higher, so recall improves without losing relevance ordering. Two hardening changes ride along:
- embedded double-quotes are stripped (not just trimmed from the ends), so a term like `a"b` can no longer produce an unbalanced FTS5 string (`"a"b"` → syntax error);
- empty / whitespace-only queries are guarded at both call sites so they never reach an empty `MATCH` (which FTS5 rejects).

### 2. Recent sessions showed a frozen creation date instead of recent activity

`defaultSessionID` returned a static id (`manual-save-{project}`) that never rotated, so every observation saved without an explicit session id piled into one row whose `started_at` was frozen at first creation. `FormatContext` then displayed that stale date as the "recent" session — months out of date even with fresh activity.

**Fix:**
- rotate the default session id per UTC day (`manual-save-{project}-{YYYY-MM-DD}`);
- add `SessionSummary.LastActivityAt = MAX(COALESCE(o.created_at, s.started_at))` and display it in `FormatContext` instead of `started_at` (applied consistently to `RecentSessions` and `AllSessions`).

`started_at` is `NOT NULL`, so the `COALESCE` is never NULL and the scan target stays a plain `string`.

## Tests

Adds to `internal/store/store_test.go`: OR partial-match recall, BM25 full-vs-partial ranking, prompt multi-term search, and edge cases (empty query, single term, hyphenated term, embedded quote). Updates `internal/mcp/mcp_test.go` for the rotated session id. Pre-existing `TestSessionsOrderedByMostRecentActivity` already covers the recent-activity ordering.

`go build ./...` and `go vet ./internal/...` are clean. (On Windows, `TestMigrate_Idempotent` and `TestNewErrorBranches` fail only in `TempDir` cleanup — `modernc.org/sqlite` holds the db file handle past `Close()`; unrelated to this change.)

## Summary by CodeRabbit

* **New Features**
  * Sessions now include last activity timestamp, showing when the session was most recently used.

* **Improvements**
  * Search now matches any search term in your query, returning more relevant results.
  * Enhanced search query parsing to better handle edge cases and special characters.
  * Session organization improved with daily ID rotation.

### DIFF
--- a/internal/mcp/mcp.go
+++ b/internal/mcp/mcp.go
@@ -1716,14 +1716,17 @@ func jsonMarshal(v any) ([]byte, error) {
 
 // ─── Helpers ─────────────────────────────────────────────────────────────────
 
-// defaultSessionID returns a project-scoped default session ID.
-// If project is non-empty: "manual-save-{project}"
-// If project is empty: "manual-save"
+// defaultSessionID returns a project-scoped, date-rotating default session ID.
+// Rotating by day prevents all observations from piling up in a single June
+// session when the binary is long-lived.
+// If project is non-empty: "manual-save-{project}-{YYYY-MM-DD}"
+// If project is empty: "manual-save-{YYYY-MM-DD}"
 func defaultSessionID(project string) string {
+	date := time.Now().UTC().Format("2006-01-02")
 	if project == "" {
-		return "manual-save"
+		return "manual-save-" + date
 	}
-	return "manual-save-" + project
+	return "manual-save-" + project + "-" + date
 }
 
 func intArg(req mcp.CallToolRequest, key string, defaultVal int) int {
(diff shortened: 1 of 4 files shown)


totakid420:
Closing this for now — we'll keep these changes local for the moment. Thanks for the great project!

PR #558, opened 2026-06-30
fix: stamp version into cloud Docker image build

### LINKED ISSUE #557: Cloud Docker image reports `engram dev` instead of the release version
## What happens

The published cloud image (`ghcr.io/gentleman-programming/engram:latest` and the `vX.Y.Z` tags) reports `engram dev` instead of the release version:

```console
$ docker run --rm ghcr.io/gentleman-programming/engram:v1.17.0 version
engram dev
```

Homebrew binaries report the version correctly (`engram 1.17.0`), so the gap is specific to the Docker image.

## Root cause

`docker/cloud/Dockerfile` builds the binary without injecting the version:

```dockerfile
RUN CGO_ENABLED=0 ... go build -o /out/engram ./cmd/engram
```

`cmd/engram/main.go` defaults `var version = "dev"` and only recovers a real value from `-X main.version` (set by goreleaser) or from `debug.ReadBuildInfo()`. The Docker build uses neither: it skips the ldflag, and because it compiles from a source `COPY` (not `go install ...@vX.Y.Z`), `info.Main.Version` resolves to `(devel)`. So `version` stays `dev`.

`.goreleaser.yaml` injects the ldflag for the brew/archive artifacts, but the cloud image is built separately by `.github/workflows/cloud-image.yml`, which never passes the version through.

## Suggested fix

Add a `VERSION` build arg to the Dockerfile, inject it via `-ldflags "-X main.version=$VERSION"`, and pass `steps.meta.outputs.version` from the publish workflow. Small, two-file change.

Happy to send a PR (already have it ready and verified locally).

mtriggiano:
## Summary

The published cloud image reports `engram dev` instead of the release version. This injects the version into the Docker build so `engram version` matches the tag, the same way the brew/archive artifacts already do.

```console
# before
$ docker run --rm ghcr.io/gentleman-programming/engram:v1.17.0 version
engram dev
```

## Root cause

`docker/cloud/Dockerfile` compiled the binary without `-X main.version`, and from a source `COPY` rather than a versioned module, so `debug.ReadBuildInfo()` in `cmd/engram/main.go` fell back to `(devel)` and `version` stayed `dev`. `.goreleaser.yaml` injects the ldflag for brew/archives, but the cloud image is built separately by `cloud-image.yml`, which never passed the version through.

## Change

- `docker/cloud/Dockerfile`: add a `VERSION` build arg (default `dev`) and inject it via `-ldflags "-s -w -X main.version=$VERSION"`, plus `-trimpath` to match goreleaser's flags.
- `.github/workflows/cloud-image.yml`: pass `VERSION=${{ steps.meta.outputs.version }}` to the build, so release tags flow into the binary. Manual `workflow_dispatch` builds fall through to the `sha-...` metadata value.

Two files, no behavior change beyond the reported version. Local builds with no arg still report `dev`.

## Verification

Built the image locally with the arg and confirmed the binary picks it up:

```console
$ docker build -f docker/cloud/Dockerfile --build-arg VERSION=1.17.0-verify -t engram-verify .
$ docker run --rm engram-verify version
engram 1.17.0-verify
```

Closes #557

## Summary by CodeRabbit

* **New Features**
  * Cloud image builds now include release version information in the built app, making deployed builds easier to identify.
  * Build outputs are now produced in a more reproducible way.

* **Chores**
  * Updated the cloud image build process to pass version details through to the final image.

### DIFF
--- a/.github/workflows/cloud-image.yml
+++ b/.github/workflows/cloud-image.yml
@@ -72,5 +72,7 @@ jobs:
           file: docker/cloud/Dockerfile
           platforms: linux/amd64,linux/arm64
           push: true
+          build-args: |
+            VERSION=${{ steps.meta.outputs.version }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/docker/cloud/Dockerfile
+++ b/docker/cloud/Dockerfile
@@ -8,7 +8,12 @@ COPY go.mod go.sum ./
 RUN go mod download
 
 COPY . .
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH:-amd64} go build -o /out/engram ./cmd/engram
+
+# VERSION is injected via ldflags so `engram version` reports the release tag.
+# Defaults to "dev" for local builds; the release workflow passes the git tag.
+ARG VERSION=dev
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH:-amd64} \
+    go build -trimpath -ldflags="-s -w -X main.version=${VERSION}" -o /out/engram ./cmd/engram
 
 FROM alpine:3.21
 RUN apk add --no-cache ca-certificates tzdata


Alan-TheGentleman:
Closing this because it skips the issue-first process. No PR merges without a linked `status:approved` issue. #557 is now approved, so reopen this PR with `Closes #557` in the body and I'll review it. Not a rejection of the fix itself.

danielgap:
Hi @mtriggiano, this one is unblocked now. #557 is `status:approved` and the close here was process-only: reopening this PR with `Closes #557` in the body puts it straight back in review, per Alan's note above.

Your branch `fix/cloud-image-version-stamping` still exists on your fork, so it's literally a reopen plus one line in the description. Happy to test the built image locally once it's up if that helps.

PR #586, opened 2026-07-09
fix(store): escape interior double-quotes in sanitizeFTS to prevent FTS5 crash

### LINKED ISSUE #574: [Bug] FTS5 search returns HTTP 500 on queries containing an interior double-quote (e.g. foo"bar)
## 📋 Pre-flight Checks

## 📝 Bug Description
`GET /search` and `GET /prompts/search` return **HTTP 500** for ordinary search input that contains a double-quote *inside* a word (e.g. `foo"bar`):

```
$ curl -s 'http://127.0.0.1:<port>/search?q=foo%22bar'
{"error":"search: SQL logic error: unterminated string (1)"}
```

**Root cause:** `sanitizeFTS` (`internal/store/store.go`) wraps each whitespace-separated token in double-quotes so FTS5 treats it as a phrase. It strips *surrounding* quotes with `strings.Trim(w, "\"")` but leaves **interior** quotes untouched, so `foo"bar` becomes the phrase `"foo"bar"` — an unbalanced FTS5 string literal. SQLite aborts the whole `MATCH` query with `unterminated string`, `Store.Search` propagates the error, and the HTTP handler maps it to 500. Whether a given input crashes is data-dependent (an even count of interior quotes happens to balance; an odd count does not), which makes it easy to hit and hard to explain.

## 🔄 Steps to Reproduce
1. Start the engram HTTP server and create at least one observation.
2. `curl -s 'http://127.0.0.1:<port>/search?q=foo%22bar'`
3. Server responds `500` with `search: SQL logic error: unterminated string`.

## ✅ Expected Behavior
The query is treated as a literal search for the token `foo"bar` and returns `200` with normal results (any matching observations), instead of a 500.

## 🌐 Environment
- Component: `internal/store` FTS5 query sanitization; affects the `/search` and `/prompts/search` handlers.
- Reproduced against current `main`.

---
I have a minimal fix ready — escape interior double-quotes by doubling them (the FTS5 literal escape), so `foo"bar` sanitizes to the valid phrase `"foo""bar"`; plain queries are unchanged — plus fail-before/pass-after regression tests in `internal/store`. Happy to open a clean PR that references this issue with `Closes #N` once it's approved.

blak0p:
## 🔗 Linked Issue

Closes #574

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix

---

## 📝 Summary

- Fixes FTS5 search crash on queries containing an interior double-quote (e.g. `hello"world` → `SQL logic error: unterminated string (1)`).
- Doubles interior " inside `sanitizeFTS` per the FTS5 string-literal escape rule ("""), so `hello"world` → `"hello""world"`.
- Adds `TestSanitizeFTS` with 8 cases (written first, failing per TDD; all pass after the fix).

---

## 🔧 Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | Add `strings.ReplaceAll(w, \"\", \"\"\"\")` after `Trim` in `sanitizeFTS` to escape interior double-quotes for FTS5. |
| `internal/store/store_test.go` | Add `TestSanitizeFTS` covering plain words, multiple words, single/multiple interior quotes, already-quoted, just-quotes, mixed, and empty input. |

---

## 🧪 Test Plan

- [x] New test `TestSanitizeFTS` fails on unpatched code (RED) and passes after the fix (GREEN): `go test ./internal/store/ -run TestSanitizeFTS -v`
- [x] Full store package passes (no regressions): `go test ./internal/store/`
- [x] Built binary no longer crashes: `go build -o /tmp/engram-fix ./cmd/engram && /tmp/engram-fix search 'hello"world'` → clean "No memories found" (exit 0).
- [x] Normal search still works: `/tmp/engram-fix search 'sanitizeFTS'` → returns results.

---

## ✅ Contributor Checklist

- [x] Linked an approved issue (#574)
- [x] Added exactly one `type:*` label (`type:bug`)
- [x] Tests included with the fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search query handling so text containing double quotes no longer produces malformed search expressions, resulting in more reliable search behavior.
* **Tests**
  * Added unit test coverage for search-term formatting across common scenarios, including plain terms, multi-word input, inputs with interior quotes, mixed quoted/unquoted sequences, and empty input.

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -6587,6 +6587,11 @@ func sanitizeFTS(query string) string {
 	for i, w := range words {
 		// Strip existing quotes to avoid double-quoting
 		w = strings.Trim(w, `"`)
+		// Double interior double-quotes: FTS5 escapes a literal " inside a
+		// quoted phrase by doubling it (""). Without this, `hello"world`
+		// becomes `"hello"world"` — an unterminated string literal that crashes
+		// the query with "SQL logic error: unterminated string". See #574.
+		w = strings.ReplaceAll(w, `"`, `""`)
 		words[i] = `"` + w + `"`
 	}
 	return strings.Join(words, " ")
--- a/internal/store/store_test.go
+++ b/internal/store/store_test.go
@@ -8802,3 +8802,31 @@ func TestSearch_WeightedBM25Ranking(t *testing.T) {
 			results[1].ID, results[1].Title, results[1].Rank)
 	}
 }
+
+// TestSanitizeFTS verifies sanitizeFTS escapes interior double-quotes per FTS5
+// string-literal rules, preventing "unterminated string" crashes on inputs like
+// `hello"world`. See issue #574.
+func TestSanitizeFTS(t *testing.T) {
+	cases := []struct {
+		name  string
+		input string
+		want  string
+	}{
+		{"plain word", "foo", `"foo"`},
+		{"multiple words", "fix auth bug", `"fix" "auth" "bug"`},
+		{"interior double-quote (the bug)", `foo"bar`, `"foo""bar"`},
+		{"already quoted", `"hello"`, `"hello"`},
+		{"multiple interior quotes", `a"b"c`, `"a""b""c"`},
+		{"just quotes", `""`, `""`},
+		{"mixed quote and plain", `hello"world test`, `"hello""world" "test"`},
+		{"empty input", "", ""},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := sanitizeFTS(tc.input)
+			if got != tc.want {
+				t.Errorf("sanitizeFTS(%q) = %q, want %q", tc.input, got, tc.want)
+			}
+		})
+	}
+}


Alan-TheGentleman (changes requested):
Correct fix, exemplary size. `strings.ReplaceAll(w, "\"", "\"\"")` after the `Trim` is the right FTS5 literal escape, and the 8-case table-driven test is exactly what this needed.

- [ ] Only blocker: rebase on `main` to resolve the conflict. It's positional, not semantic — you and #537 both edit the region right after `sanitizeFTS`.

Worth noting for whoever reviews #537: **this fix is still required after that one lands.** Its LIKE fallback uses `searchTerms()` and bypasses `sanitizeFTS` entirely, but `foo"bar` is 7 runes so it still takes the FTS path and still crashes without your escape.

One cosmetic thing, not blocking: `Trim` still strips an unbalanced trailing quote (`a"` → `"a"`), losing a character. Not a crash, just slightly lossy.

`type:bug` applied. Ping me after the rebase and I'll merge.

Alan-TheGentleman:
Heads up: I rebased this on `main` for you and force-pushed to `fix/fts5-interior-quote`, since the conflict was purely mechanical and this is a crash fix worth landing quickly. **Your commit and authorship are unchanged** — only the parent moved (`9190069` → `08fd805`).

What the conflict actually was: both `main` and your branch appended new tests to the end of `internal/store/store_test.go`, so git interleaved them. I took `main`'s version of the file and re-appended your `TestSanitizeFTS` intact. Your one-line production change in `sanitizeFTS` applied cleanly with no conflict at all.

Verified locally before pushing: `go build ./...` clean, `gofmt` clean, and the full `internal/store` suite passes including all 8 of your table cases.

If you have local work on this branch, `git fetch && git reset --hard origin/fix/fts5-interior-quote` before continuing.

Merging once CI goes green. Thanks for the fix — the escape is exactly right, and it stays necessary even after #537 lands, since that PR's LIKE fallback bypasses `sanitizeFTS` but `foo"bar` is 7 runes and still takes the FTS path.

Alan-TheGentleman (approved):
Rebase done and CI green across all gates. The escape is correct and the coverage is solid — merging.

blak0p:
Thanks so much, Alan, really didn’t need to, but I appreciate it a lot. Thanks for the rebase and for the review too!

Issue #536, opened 2026-06-22
Windows: console flash when Pi auto-starts engram serve (startup and subagents)


viniciosrab:
## 📋 Pre-flight Checks

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/engram/issues) and this is not a duplicate
- [x] I understand this issue needs `status:approved` before a PR can be opened

## 📝 Bug Description

On Windows, when the Pi extension auto-starts `engram serve`, users can see brief terminal/console flashes during PI startup. The same flash can happen again when a subagent starts, because a child PI process re-runs the extension bootstrap path.

The issue is not the Engram server behavior itself; it is the Windows process launch behavior when the console binary is started on demand from the PI extension.

## 🔄 Steps to Reproduce

1. On Windows, install and use the Engram PI extension with automatic local server startup enabled.
2. Make sure no `engram serve` process is already running.
3. Start PI.
4. Observe a brief console/terminal flash while the extension starts Engram.
5. Start a PI subagent that triggers the same bootstrap path.
6. Observe the flash again.

## ✅ Expected Behavior

`engram serve` should still auto-start on demand, but without any visible console flash on Windows, both for the main PI process and for subagent processes.

## ❌ Actual Behavior

The server starts successfully, but Windows can briefly show a console flash during startup. This degrades startup UX and becomes more noticeable when subagents trigger the same path.

## 🖥️ Environment

- Operating System: Windows
- Engram Version: 1.17.0
- Agent / Client: Other (PI)

## 📋 Relevant Logs

```shell
Observed process path during reproduction:
- PI extension checks health
- PI extension starts `engram serve` when server is not already running
- On Windows, a visible console flash can occur during this startup
- Subagent startup can retrigger the same bootstrap path in a separate PI process
```

## 💡 Additional Context

A local fix that resolved the issue in practice was:

- use a Windows-specific hidden launcher instead of spawning `engram.exe` directly from the PI process
- add a cross-process startup lock/mutex so PI and subagents do not race to start the server simultaneously

That preserves current behavior:

- still auto-starts on demand
- no manual server startup required
- avoids visible startup flashes on Windows
- prevents concurrent bootstrap races across PI/subagents

developerFer:
## Still reproduces — root cause identified + working workaround

This is still happening on current versions (`engram` v1.20.0, `gentle-engram` v0.1.10, Windows). I did a process-level investigation and can add concrete evidence and a workaround.

### Root cause

[...]

PR #604, opened 2026-07-12
fix(pi): hide detached process windows on Windows

### LINKED ISSUE #599: fix(windows): Engram Git subprocesses open transient terminal windows
## Pre-flight Checks

- [x] I searched the existing issues. #536 reports the same visible-console symptom; the relationship and additional evidence are documented below.
- [x] I understand this issue requires `status:approved` before a PR can be opened.

## Bug Description

Starting Pi causes several transient console or Windows Terminal windows to appear. The same behavior occurs when Pi launches subagents.

Process monitoring identified Git commands spawned by `engram serve` as the primary source. These subprocesses create `conhost.exe`, which can subsequently activate `OpenConsole.exe` or `WindowsTerminal.exe`.

## Steps to Reproduce

1. Install and enable `gentle-engram` in Pi on native Windows.
2. Ensure no Engram daemon is currently running.
3. Start a new Pi session.
4. Observe several transient terminal windows.
5. Launch a Pi subagent.
6. Observe the windows appearing again.

## Expected Behavior

Engram and its Git subprocesses should run silently in the background without creating or activating visible terminal windows.

## Actual Behavior

Engram executes repository-discovery commands such as:

```text
git rev-parse --show-toplevel
git remote get-url origin
```

The observed process chain was:

```text
engram serve
└─ git.exe
   └─ conhost.exe
      └─ OpenConsole.exe / WindowsTerminal.exe
```

This caused multiple terminal windows to appear briefly during Pi startup and subagent creation.

## Operating System

Windows

## Engram Version

- Engram binary: 1.19.0
- `gentle-engram` Pi plugin: reproduced on 0.1.8 and still present on 0.1.10

## Agent / Client

Other: Pi Coding Agent 0.80.6

The session runs on native Windows using Nushell inside WezTerm.

## Relevant Logs

```text
engram.exe
Command line: engram serve

git.exe
Parent: engram.exe
Command line:
git -C "<project>" rev-parse --show-toplevel

conhost.exe
Parent: git.exe
Command line:
\??\C:\WINDOWS\system32\conhost.exe 0x4

git.exe
Parent: engram.exe
Command line:
git -C "<project>" remote get-url origin

OpenConsole.exe
Command line:
WindowsTerminal\OpenConsole.exe -Embedding
```

### Relationship to #536

#536 reports the same startup and subagent console-flash symptom and proposes a Windows-specific hidden launcher plus a cross-process startup lock. This issue adds process-monitor evidence that traces the visible windows to `engram serve` Git subprocesses and confirms that adding `windowsHide: true` to the existing detached spawn path eliminates the observed windows.

The minimal PR for this issue addresses process visibility only. If maintainers consider the startup race a separate concern, it can remain tracked in #536; otherwise, these issues can be consolidated.

### Verified workaround

A local workaround was applied in:

```text
gentle-engram/index.ts
```

The detached process configuration was changed to:

```ts
proc = spawn(command, [...args], {
  cwd,
  detached: true,
  stdio: "ignore",
  windowsHide: true,
});
```

The existing Engram daemon was then stopped and restarted so it inherited the corrected process configuration.

After applying the workaround:

- no Engram-owned Git processes produced visible consoles;
- no `OpenConsole.exe` or `WindowsTerminal.exe` processes appeared during startup;
- the user confirmed that no windows appeared visually.

Pi also runs multiple package update checks through:

```text
pnpm view <package> version --json
```

Those checks were observed using `cmd.exe`, but they did not produce visible windows during the final verification.

The workaround currently modifies `node_modules` directly and is therefore overwritten by package upgrades or reinstalls. The fix should be incorporated into `gentle-engram` itself.

MarsSall:
## 🔗 Linked Issue

Closes #599

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix

---

## 📝 Summary

- Hide detached Engram subprocess windows on Windows.
- Add a source regression test that protects the `windowsHide: true` spawn option.

## 📂 Changes

| File | Change |
|------|--------|
| `plugin/pi/index.ts` | Set `windowsHide: true` when spawning detached Engram processes. |
| `plugin/pi/test/index-source.test.mjs` | Verify detached process spawning keeps Windows console windows hidden. |

## 🧪 Test Plan

- [ ] Unit tests pass locally: `go test ./...`
- [ ] E2E tests pass locally: `go test -tags e2e ./internal/server/...`
- [x] Regression assertion passes: `cd plugin/pi && node --test test/index-source.test.mjs` (`spawnDetached hides detached process windows`)

The targeted regression assertion passes. The test file still reports two unrelated, pre-existing `Cannot use import statement outside a module` failures that reproduce on `upstream/main`.

---

## 🤖 Automated Checks

| **Check Issue Reference** | PR body contains `Closes/Fixes/Resolves #N` | ✅ Linked |
| **Check Issue Has status:approved** | Linked issue has `status:approved` label | ⏳ Maintainer action required |
| **Check PR Has type:* Label** | PR has exactly one `type:*` label | ⏳ Maintainer action required |
| **Unit Tests** | `go test ./...` passes | ⏳ Workflow approval required |
| **E2E Tests** | `go test -tags e2e ./internal/server/...` passes | ⏳ Workflow approval required |

---

## ✅ Contributor Checklist

- [ ] I linked an approved issue above (`Closes #N`) (linked, awaiting maintainer approval)
- [ ] I added exactly **one** `type:*` label to this PR (awaiting maintainer permissions)
- [x] Documentation is not required for this internal process-launch fix

---

## 💬 Notes for Reviewers

Issue #599 is not yet labeled `status:approved`, and this fork contributor cannot add repository labels. A maintainer needs to:

1. decide whether #599 should remain separate from the related #536;
2. add `status:approved` to #599 and `type:bug` to this PR;
3. approve the pending fork workflows so CI and PR Validation can run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented console windows from appearing when background processes start on Windows.
  * Improved the Windows experience when detecting projects through Git commands.

* **Tests**
  * Added platform-specific coverage to verify hidden process windows and preserve existing command behavior.

### DIFF
--- a/internal/project/command_other.go
+++ b/internal/project/command_other.go
@@ -1,0 +1,12 @@
+//go:build !windows
+
+package project
+
+import (
+	"context"
+	"os/exec"
+)
+
+func newProjectCommandContext(ctx context.Context, name string, args ...string) *exec.Cmd {
+	return exec.CommandContext(ctx, name, args...)
+}
--- a/internal/project/command_other_test.go
+++ b/internal/project/command_other_test.go
@@ -1,0 +1,20 @@
+//go:build !windows
+
+package project
+
+import (
+	"context"
+	"reflect"
+	"testing"
+)
+
+func TestNewProjectCommandContextPreservesDefaultProcessAttributes(t *testing.T) {
+	cmd := newProjectCommandContext(context.Background(), "git", "status")
+
+	if want := []string{"git", "status"}; !reflect.DeepEqual(cmd.Args, want) {
+		t.Fatalf("command arguments = %q; want %q", cmd.Args, want)
+	}
+	if cmd.SysProcAttr != nil {
+		t.Fatalf("SysProcAttr = %#v; want nil", cmd.SysProcAttr)
+	}
+}
--- a/internal/project/command_usage_test.go
+++ b/internal/project/command_usage_test.go
@@ -1,0 +1,61 @@
+package project
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"path/filepath"
+	"runtime"
+	"testing"
+)
+
+func TestProjectDetectionGitCommandsUsePlatformHelper(t *testing.T) {
+	_, testFile, _, ok := runtime.Caller(0)
+	if !ok {
+		t.Fatal("locate test source")
+	}
+
+	file, err := parser.ParseFile(token.NewFileSet(), filepath.Join(filepath.Dir(testFile), "detect.go"), nil, 0)
+	if err != nil {
+		t.Fatalf("parse detect.go: %v", err)
+	}
+
+	tests := []struct {
+		name string
+	}{
+		{name: "detectGitRootDir"},
+		{name: "detectFromGitRemote"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var function *ast.FuncDecl
+			for _, declaration := range file.Decls {
+				candidate, ok := declaration.(*ast.FuncDecl)
+				if ok && candidate.Name.Name == tt.name {
+					function = candidate
+					break
+				}
+			}
+			if function == nil {
+				t.Fatalf("function %s not found", tt.name)
+			}
+
+			helperCalls := 0
+			ast.Inspect(function.Body, func(node ast.Node) bool {
+				call, ok := node.(*ast.CallExpr)
+				if !ok {
+					return true
+				}
+				identifier, ok := call.Fun.(*ast.Ident)
+				if ok && identifier.Name == "newProjectCommandContext" {
+					helperCalls++
+				}
+				return true
+			})
+			if helperCalls != 1 {
+				t.Fatalf("%s uses newProjectCommandContext %d times; want 1", tt.name, helperCalls)
+			}
+		})
+	}
+}
--- a/internal/project/command_windows.go
+++ b/internal/project/command_windows.go
@@ -1,0 +1,15 @@
+//go:build windows
+
+package project
+
+import (
+	"context"
+	"os/exec"
+	"syscall"
+)
+
+func newProjectCommandContext(ctx context.Context, name string, args ...string) *exec.Cmd {
+	cmd := exec.CommandContext(ctx, name, args...)
+	cmd.SysProcAttr = &syscall.SysProcAttr{HideWindow: true}
+	return cmd
+}
--- a/internal/project/command_windows_test.go
+++ b/internal/project/command_windows_test.go
@@ -1,0 +1,23 @@
+//go:build windows
+
+package project
+
+import (
+	"context"
+	"reflect"
+	"testing"
+)
+
+func TestNewProjectCommandContextHidesWindow(t *testing.T) {
+	cmd := newProjectCommandContext(context.Background(), "git", "status")
+
+	if want := []string{"git", "status"}; !reflect.DeepEqual(cmd.Args, want) {
+		t.Fatalf("command arguments = %q; want %q", cmd.Args, want)
+	}
+	if cmd.SysProcAttr == nil {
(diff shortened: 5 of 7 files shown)


Alan-TheGentleman (changes requested):
Thanks for chasing this down — the bug in #599 is real and worth fixing.

The problem is that this fixes the wrong layer. The evidence chain in #599 is `engram serve → git.exe → conhost.exe → WindowsTerminal.exe`, i.e. the windows come from **Go-side subprocesses**, not from the Node spawn. Those live at `internal/project/detect.go:282` (`git -C … rev-parse --show-toplevel`) and `:381` (`git … remote get-url origin`) — the exact commands in your report. `rg "HideWindow|CreationFlags|SysProcAttr"` returns nothing across the repo, so no Go call site currently suppresses a console.

`windowsHide` on the Node spawn only suppresses the `engram serve` window. A console-subsystem child like `git.exe` launched from a console-less parent still allocates its own console, so the transient windows you're seeing would persist.

- [ ] Move the fix into Go: set `SysProcAttr{HideWindow: true}` on `detect.go:282` and `:381` behind a `//go:build windows` helper. This also matches the thin-adapter rule — plugins are shims, logic belongs in core.
- [ ] Keep `windowsHide: true` as a complementary plugin change if you like, but replace the source-regex test: asserting `windowsHide: true,` appears *before* `detached: true,` breaks on any key reorder without a behavior change.
- [ ] Rebase — #640 just landed and rewrote `plugin/pi/index.ts` substantially, including moving `spawnDetached`.

I'll get #599 labelled `status:approved` so this isn't blocked on process.

Alan-TheGentleman (approved):
This is exactly the change I was asking for, and then some. Approving.

You moved the fix to the layer where the bug actually lives: `newProjectCommandContext` behind `//go:build windows` setting `SysProcAttr{HideWindow: true}`, with a plain passthrough for every other platform, wired into `detectGitRootDir` and `detectFromGitRemote` — the two `git.exe` call sites named in #599's `engram serve → git.exe → conhost.exe` chain. Keeping `windowsHide: true` on the Node spawn as a complementary change is right too, and you dropped the brittle source-regex test I flagged.

What I verified locally on your branch, rebased onto current `main`:

- `GOOS=windows GOARCH=amd64 go build ./...` **passes**. Worth calling out because `command_windows.go` and its `syscall.SysProcAttr` never compile on the maintainers' macOS machines or in the default CI job, so this file could have been broken without anyone noticing. It isn't.
- `go build ./...` and the full `go test ./...` pass, 22 packages.
- `plugin/pi` suite passes 42/42, so the `index.ts` change composes cleanly with #640 which rewrote that file.
- `TestNewProjectCommandContextPreservesDefaultProcessAttributes` and `TestProjectDetectionGitCommandsUsePlatformHelper` both pass.

The AST-based test is a nice touch. Asserting that each detection function calls the helper exactly once means a future contributor who reaches for a raw `exec.CommandContext` there gets a failing test instead of silently reintroducing the console flash. That's the right way to guard a platform behavior you can't assert on directly from CI.

### Follow-up, not blocking this PR

Two other subprocess call sites are in the same bug class but outside #599's reported path, so I don't want to hold this fix for them:

- `internal/llm/claude.go:140` (`defaultRunCLI`) spawns the agent CLI and will flash a console on Windows the same way.
- `internal/setup/setup.go:36` (`runCommand`) does too, though it runs during interactive `engram setup` where a console window is far less surprising.

Happy to open a separate issue for those, or you can if you want to carry the work forward — the helper you built here is the obvious thing to reuse, probably lifted somewhere both packages can import.

Alan-TheGentleman:
Merged. Opened #641 to track the remaining call sites (`internal/llm/claude.go` and `internal/setup/setup.go`) and pre-approved it, so it's open if you want to carry the helper forward — lifting it out of `internal/project` into something all three packages can import is probably the cleaner shape. Thanks for reworking this into the right layer.

Issue #571, opened 2026-07-05
bug(store): single-conn pool from #271 has no recovery path when WAL-index goes stale


danielgap:
### 📋 Pre-flight Checks

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/engram/issues) and this is not a duplicate
- [x] I understand this issue needs `status:approved` before a PR can be opened

### 📝 Bug Description

This is a follow-up on the architectural decision in #271 (fixed in v1.14.6). That fix correctly enforced `db.SetMaxOpenConns(1)` so the `busy_timeout=5000` pragma always applies to every connection in the pool. The single-connection decision is sound for SQLite single-writer correctness.

However, the decision has an undocumented trade-off that amplifies the severity of #477 (stale WAL-index mmap in long-running processes): with a single connection, there is **zero concurrency headroom**. When the WAL-index mmap goes stale (per #477), the single connection is the ONLY connection, so every read AND write operation fails or hangs. There is no secondary path, and no code path triggers a reopen.

Concretely: `/health` continues to respond (trivial handler, no DB access), while every DB-backed endpoint (`/search`, `/observations`, writes) silently fails. The process cannot self-recover; it requires an external `kill` + restart.

**This is not a duplicate of #477 or #206.** It is the architectural gap between them:
- #477 documents the staleness mechanism (the "what").
- #206 requests a graceful reload feature (the "how").
- This issue documents WHY the single-conn design from #271 makes that reload mandatory rather than optional (the "why it matters"). Without #206, #271's design is correct but brittle.

### 🔄 Steps to Reproduce

1. Run `engram serve` as a long-lived daemon (no cloud sync, no MCP needed to trigger this).
2. After 1-7 days of uptime (reproduced here at ~1.7 days), any condition that replaces the `-shm`/`-wal` files on disk (corruption recovery, external checkpoint, or the natural staleness in #477) leaves the daemon operating on a divergent WAL view.
3. Observe: `/health` returns `ok`, `mem_doctor` reports 4/4 checks passing (including `checkpoint_log: 890`), but `/search` hangs and returns `SQLITE_BUSY` after `busy_timeout=5000`.
4. Confirm via `lsof -p $(pgrep -f 'engram serve') | grep db-shm` that the daemon holds a stale inode versus a fresh process.
5. The daemon cannot recover without a manual `kill` + restart.

[...]

Issue #508, opened 2026-06-16
bug(store): Database maintenance toolset. Multiple data inconsistencies found.


carolitascl:
### 📋 Pre-flight Checks

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/engram/issues) and this is not a duplicate
- [x] I understand this issue needs `status:approved` before a PR can be opened

### 🔍 Problem Description

Reviewed Engram database contents and found issues that engram options currently cannot address. Multiple data inconsistencies were detected across tables.

#### Add database maintenance toolset for Engram to address data inconsistencies.

**_1. Project deletion:_** Delete a specified project and all associated records across every related table. A project deletion must be fully cascading so that no residual records remain in sessions, observations, prompts, memory_relations, sync_enrolled_projects, sync_mutations, or any other project-bearing table. Currently, deleted projects can reappear in "engram tui" until "--hard" is used, indicating that standard deletion leaves orphaned records behind.

**_2. Project rename:_** Rename a specified project and propagate the new name consistently across all tables that store the denormalized project field. Since there is no master projects table, every affected record must be updated atomically to prevent project fragmentation.

**_3. Directory-to-project mapping repair:_** Rebuild the project field from the directory field for sessions where the project value is missing, empty, or inconsistent with the directory basename. This repairs records where a valid directory exists but the associated project identifier is absent or incorrect.

**_4. Project name normalization:_** Enforce canonical project naming by applying case-insensitive matching and stripping version suffixes at write time. This prevents project fragmentation caused by variants such as ProjectName, projectname, ProjectName-0.0.1, and ProjectName-0.0.1-2 being treated as separate projects.

**_5. Phantom record cleanup:_** Detect and optionally purge stale, unreachable, or invalid records, including:

* Soft-deleted observations whose deleted_at retention period has expired.
* Zero-size observations, sessions, or prompts.
* memory_relations entries referencing missing sessions, observations, prompts, or projects.
* sync_enrolled_projects entries with no corresponding sessions or content.
* Other dangling or unreachable records discovered through referential integrity checks.

**_6. Project cleanup:_** Detect and optionally batch delete orphaned projects that have no sessions, observations, prompts, or other meaningful content. This includes projects that were never active, projects whose primary records were already removed, and projects that exist only because of stale metadata.

**_7. Sync mutation audit and repair:_** Audit sync_mutations for unresolved or malformed entries and optionally repair or purge them. This includes:

[...]

PR #613, opened 2026-07-15
fix(store): persistent WAL + version-gated migrations to stop multi-process SQLite corruption

### LINKED ISSUE #477: bug(store): long-running processes keep stale WAL-index (-shm) mmap after file replacement — silent write loss + constraint failed (1555) on mem_session_summary
### 📋 Pre-flight Checks

### 📝 Bug Description

Long-running engram processes (`engram serve` and `engram mcp`) keep a stale WAL-index (`engram.db-shm`) mmap after that file is replaced on disk (e.g. by corruption recovery renaming `*.corrupt.*` files, or an external `PRAGMA wal_checkpoint(TRUNCATE)` from another connection). From that point the process operates on a divergent WAL view of the shared `~/.engram/engram.db`:

1. **Silent write loss** — the process commits successfully from its own perspective, but the data never becomes visible to any other connection and is gone after the next checkpoint. The HTTP API returns `201 created` for a session that never lands.
2. **`constraint failed (1555)`** (SQLITE_CONSTRAINT_PRIMARYKEY) — a poisoned `engram mcp` process consistently fails `mem_session_summary` because its phantom view collides on insert.

The user-visible failure chain (all observed in one real session):

- SessionStart hook → `POST /sessions` to the daemon → `201`, but the row is never visible (daemon had the stale view), so no active session exists.
- `mem_save` (MCP, healthy view at that point) falls back to creating `manual-save-{project}` — works.
- `mem_session_summary` → `Failed to save session summary: constraint failed (1555)` — repeatedly.
- `mem_session_start` via the same MCP process reports success, but the session row never becomes durable (that process had become stale by then too).

### 🔄 Steps to Reproduce

The trigger is environmental (the `-shm` file being replaced underneath a live process), so the repro is forensic rather than scripted:

1. Have a long-running `engram serve` plus one or more long-running `engram mcp` processes sharing `~/.engram/engram.db`.
2. Cause the WAL index to be replaced on disk while those processes are alive. In our case the DB dir contained `engram.db-shm.corrupt.<ts>` / `engram.db-wal.corrupt.<ts>` from a previous recovery; an external `sqlite3 ~/.engram/engram.db "PRAGMA wal_checkpoint(TRUNCATE)"` reproduces the same staleness on whichever process kept the old mmap.
3. Diagnose the split brain: `lsof -p <pid_serve> -p <pid_mcp> | grep db-shm` → same `engram.db` inode, **different `engram.db-shm` inodes** (observed: `72176760` vs `72426024` on the same `engram.db` inode `71107007`).
4. `curl -X POST http://127.0.0.1:7437/sessions -d '{"id":"diag-1","project":"x","directory":"/tmp"}'` → `201`, then `sqlite3 ~/.engram/engram.db "SELECT * FROM sessions WHERE id='diag-1'"` → empty.
5. Call `mem_session_summary` through the stale `engram mcp` → `constraint failed (1555)`.

**Control experiment proving data is fine:** copying the same `engram.db` to an isolated dir and replaying the exact MCP call with a *fresh* process succeeds:

```bash
printf '%s\n%s\n%s\n' \
  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"diag","version":"1.0"}}}' \
  '{"jsonrpc":"2.0","method":"notifications/initialized"}' \
  '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"mem_session_summary","arguments":{"content":"repro"}}}' \
  | ENGRAM_DATA_DIR=/tmp/engram-repro engram mcp --tools=agent
# → "Session summary saved" — same binary, same data, fresh process
```

`PRAGMA integrity_check` and FTS5 `integrity-check` on the real DB both pass — the fault is process state, not data.

### ✅ Expected Behavior

- A write acknowledged by the daemon/MCP (HTTP 201, MCP tool success) is durable and visible to other connections.
- Long-running processes detect that the WAL index they mmap'd no longer matches the file on disk (fsnotify on `-shm`/`-wal`, or verify inode before each transaction) and reopen the connection instead of silently diverging — same direction as #206 but as a correctness fix, not only a convenience reload.
- Defensively, the `mem_session_summary` fallback path should tolerate the collision (UPSERT semantics) instead of surfacing a raw `constraint failed (1555)` to the agent.

### ❌ Actual Behavior

- `POST /sessions` → `201 created`, row never visible to any other connection, lost after checkpoint.
- `mem_session_summary` → `Failed to save session summary: constraint failed (1555)` on every retry from the poisoned process, including after `mem_session_start` "succeeded" through it.
- No error, log line, or health-check signal indicates the divergence; `/health` returns `ok` and `engram doctor` / `mem_doctor` report all checks passing while writes are being lost.

### Recovery that worked (for anyone hitting this)

```bash
pkill -f "engram (mcp|serve)"
sqlite3 ~/.engram/engram.db "PRAGMA wal_checkpoint(TRUNCATE); PRAGMA integrity_check;"
# restart serve / agent sessions → fresh processes, everything consistent again
```

### Environment

- Operating System: macOS (Darwin 25.6.0, Apple Silicon)
- Engram Version: 1.16.1 (Homebrew, gentleman-programming/tap)
- Agent / Client: Claude Code (engram plugin 0.1.1)
- Shell: zsh
- Cloud sync enrolled (git-based sync configured in `~/.engram`)

### Additional Context

- Related: #206 (graceful DB reload — would likely fix the staleness half), #270 / #125 (earlier write-contention reports), Gentleman-Programming/gentle-ai#309 (session summary save failing with a DB lock — likely the same family: long-lived process + shared SQLite WAL).
- The DB dir contained `engram.db-shm.corrupt.20260422-2342` / `engram.db-wal.corrupt.20260422-2342`, so a past corruption-recovery path replaced the WAL files at least once while processes may have been live — a plausible original trigger for the divergence.

### LINKED ISSUE #559: fix(store): journal_mode=WAL switch dies instantly on concurrent cold start
### Pre-flight Checks

### Bug Description
When two engram processes cold-start against the same fresh `engram.db` at (almost) the same instant — the canonical case is the Claude Code plugin's `SessionStart` hook launching `engram serve` in the background while Claude Code itself spawns the `engram mcp` stdio server for the same session — one of them wins the race to switch `journal_mode` to WAL and the other dies immediately with "database is locked" instead of waiting it out. The MCP client sees the process exit and drops the connection for the rest of the session, with no retry.

`PRAGMA busy_timeout` does NOT protect this specific pragma: switching `journal_mode` to WAL requires exclusive access to the file, and SQLite returns `SQLITE_BUSY` immediately if another connection has the file open, regardless of `busy_timeout`. Confirmed by reproducing the exact race in a unit test (open one connection, hold a write transaction, concurrently call `store.New()` on a second connection against the same file) — reordering the pragmas alone does not fix it; retrying `PRAGMA journal_mode = WAL` at the application level does.

### Steps to Reproduce
1. Delete/rename `~/.engram/engram.db*` so the next start is a cold start
2. Trigger two engram processes to open the fresh DB at (almost) the same instant, e.g. the Claude Code plugin flow: `SessionStart` hook does `engram serve &` while Claude Code independently spawns `engram mcp --tools=agent` for the same session
3. Observe the MCP server log for that session

### Expected Behavior
The process that loses the race waits out the contention (bounded retry) and completes startup normally; both processes end up healthy.

### Actual Behavior
```
Server stderr: engram: engram: pragma "PRAGMA journal_mode = WAL": database is locked (261)
Connection failed after 553ms: MCP error -32000: Connection closed
```
The MCP tools (`mem_save`, `mem_search`, etc.) are silently absent for the rest of that Claude Code session. Reproduced identically across 6 separate real occurrences in `~/.cache/claude-cli-nodejs/*/mcp-logs-plugin-engram-engram/*.jsonl` (2026-06-20 x2, 2026-06-24, 2026-06-29, 2026-07-01), always right after a period where `engram serve` was not already running.

### Operating System
Linux (Fedora)

### Engram Version
1.15.13

### Agent / Client
Claude Code

### Relevant Logs
\`\`\`
{"debug":"Starting connection with timeout of 30000ms", ...}
{"error":"Server stderr: engram: engram: pragma \"PRAGMA journal_mode = WAL\": database is locked (261)\n", ...}
{"debug":"Connection failed after 553ms: MCP error -32000: Connection closed", ...}
{"error":"Connection failed: MCP error -32000: Connection closed", ...}
\`\`\`

### Additional Context
Fix implemented and verified locally on branch `fix/journal-mode-wal-startup-race`: added an explicit bounded retry (`execStartupPragmaWithRetry`, reusing the existing `isRetryableSQLiteLockError` classifier) around the startup pragma loop in `internal/store/store.go` (`New` and `newWithoutRepair`), plus a regression test (`TestNewSurvivesConcurrentJournalModeSwitchContention`) that reproduces the exact contention with two real `*sql.DB` connections over the same file. Full suite green, PR to follow once this issue is approved.

jarl9801:
Closes #477 · Fixes #559 · Addresses #571 (and the DB-lifecycle item in #508)

## Problem

With one `engram mcp` per agent session plus `engram serve` and periodic CLI/sync runs, ~10 processes hold `~/.engram/engram.db` open concurrently. Two corruption mechanisms have hit my store six times since April (details in #477):

1. **Startup migration race** — every subcommand runs the full `migrate()` suite plus `repairEnrolledProjectSyncMutations()` with no inter-process exclusion. The destructive check-then-act rebuilds (`migrateFTSTopicKey`, `migrateSyncChunksTable`, `migrateLegacyObservationsTable`) can double-run under concurrent cold starts → cross-linked pages / `database disk image is malformed (11)`.
2. **WAL/SHM unlink race** — SQLite's last-closer deletes `-wal`/`-shm`, deciding "last" from *its own* SHM mapping. A process whose SHM is a deleted orphan cannot see live lockers, so a clean shutdown deletes the LIVE WAL/SHM and orphans the next generation of processes (observed live on 2026-07-15; forensic signature: `lsof +L1` shows engram PIDs holding NLINK=0 `-wal`/`-shm` inodes).

Additionally `journal_mode=WAL` ran before `busy_timeout` was set, so concurrent cold starts die instantly on SQLITE_BUSY (#559).

## Fix

- **Per-connection config via DSN `_pragma` params** (`busy_timeout(5000)` → `journal_mode(WAL)` → `synchronous(NORMAL)` → `foreign_keys(1)`): applied by the driver on **every** physical connection — including pool replacements after a context-interrupted query, which previously would have silently dropped the session config (modernc discards interrupted conns via `ResetSession`/`IsValid`).
- **Persistent WAL** via a `sqlite.RegisterConnectionHook` calling `FileControlPersistWAL("main", 1)` on every new conn (strictly verified for the store's own DB in `primeConnection`; best-effort elsewhere). With persist-WAL, no closer ever unlinks `-wal`/`-shm`, eliminating mechanism 2 at the root.
- **`PRAGMA user_version`-gated migrations** (`schemaVersion = 1`): fast read-only startup when current; when stale, an **exclusive advisory file lock** (`<DataDir>/.migrate.lock`, `flock` on unix / `LockFileEx` on Windows, non-blocking retries bounded at 60s with a clear diagnostic) serializes the whole migrate+repair block across processes, with a double-checked re-read under the lock. `user_version` is stamped only after success, and a *newer* stamp is warn-and-skip (mixed-version safety). `repairEnrolledProjectSyncMutations` still runs on every open (original self-heal semantics) but now inside the lock.
- **Cold-start WAL-switch retry**: `primeConnection` retries the first pooled conn with bounded backoff, since DSN pragmas run inside `driver.Open` (#559).

## Tests

New `internal/store/startup_gate_test.go` (+ lock timeout test): persistent WAL survives `Close`; **connection-replacement test** (context-interrupted query → pool replaces conn → busy_timeout/journal_mode/foreign_keys/persist-WAL all still active, `-wal` survives close); concurrent cold start with 8 goroutines and with 3 real child processes (exactly one migration run, `integrity_check` ok); user_version gating (skip on second open, newer-stamp untouched); lock exclusion + timeout diagnostics. `go test ./internal/store/ -race -count=3` green; cross-builds verified for `windows/amd64` and `linux/amd64`.

Live smoke on a real workload (macOS ARM64): 4 concurrent cold-start `save` processes — all succeed, `-wal` persists after every exit, `user_version=1`, `integrity_check` ok. Running this build as my daily binary since 2026-07-15.

## Notes for review

- The advisory file lock was chosen over one wrapping `BEGIN IMMEDIATE` because `migrate()` helpers open their own transactions internally; each destructive rebuild keeps its per-step transaction for crash atomicity. The lock file is intentionally never unlinked.
- `schemaVersion` starts at 1 (= the fully migrated v1.19.0 schema); bump it whenever `migrate()` grows. Existing databases stamp on their first patched open (idempotent — the rebuilds skip when the schema is current).
- Happy to split into two PRs (WAL persistence vs migration gate) if you prefer — they're independent commits.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved database startup reliability when multiple application instances open the database simultaneously.
  - Prevented duplicate migrations and reduced risks of database corruption during interrupted or concurrent startups.
  - Improved recovery and consistency for SQLite write-ahead logging and interrupted queries.
  - Added safeguards for databases using newer schema versions, leaving them unchanged rather than applying incompatible migrations.
  - Improved background command execution, including hidden command windows on Windows.

- **Tests**
  - Expanded coverage for concurrent startup, migration locking, WAL persistence, schema compatibility, and connection recovery.


Alan-TheGentleman (dismissed):
This is strong work — the `FileControlPersistWAL` approach fixes #477 at the root rather than papering over it, the migration gate stamps `user_version` only after success so a crash mid-suite safely re-runs, and a downgraded binary still works because pre-gate builds run `migrate()` unconditionally. The 497:378 test-to-prod ratio is appreciated on a change this sensitive.

One thing before merge:

- [ ] `schemaVersion = 1` is currently protected only by a code comment. If anyone adds a step to `migrate()` without bumping it, that step is **silently skipped on every existing database** — which is the same class of silent schema drift this PR exists to eliminate. Please add a guard test that hashes the migration step list (or the schema string) and fails when it changes without a `schemaVersion` bump.

The +876 line count is fine here — splitting persist-WAL from the migration gate would leave either half incomplete. `type:bug` applied.

Alan-TheGentleman:
Follow-up: I approved the pending workflow runs on this PR (fork PRs here sit in `action_required` until a maintainer releases them — yours had been waiting since before the branch was current, which is why you saw no CI).

Results: `CI` is fully green — Unit Tests and E2E both pass. One gate fails, and it's a body problem rather than a code problem:

**`Check Issue Has status:approved` fails** because the linked-issue line closes more than one issue:

```
Closes #477 · Fixes #559 · Addresses #571 (and the DB-lifecycle item in #508)
```

`#477` is approved, but `#559` and `#571` carry no labels at all and `#508` is `status:needs-review`, so the gate sees unapproved references and blocks.

- [ ] Keep exactly one closing keyword — `Closes #477` — and demote the rest to plain prose without `Fixes`/`Closes`/`Resolves` (e.g. "Also addresses the WAL-switch failure described in #559 and the pool issue in #571"). The gate only inspects closing keywords, so that clears it without losing the cross-references.

That plus the `schemaVersion` guard test from my earlier review are the only two things between this and merge. The fix itself is in good shape.

danielgap:
The branch had gone stale against main (rebase marked dirty), so I rebased the work locally onto current main (59cb8f2) and fixed what the rebase surfaced. Nothing pushed — happy to share the branch or a patch file if useful.

- **Rebase**: conflicts resolved. Current main defers repair until the first synchronization, so the older startup-repair expectations no longer apply; this slice keeps repair deferred as well.
- **Deterministic cold-start tests**: the multi-process race suite now rendezvous through `.coldstart-ready-N` / `.coldstart-release` marker files instead of sleeps, so process ordering is proven rather than timing-dependent.
- **Migration gate**: `user_version` is re-read after acquiring the exclusive inter-process migration lock, and a test pins an AST-normalized fingerprint of the `migrate` body per `schemaVersion`, so editing `migrate()` without bumping the version fails loudly instead of silently no-op'ing on already-stamped databases.
- **Future schemas**: a database stamped by a newer engram fails closed instead of getting old migrations applied against it.
- **Regression found and fixed**: the original gate skipped recurring snapshot maintenance (the legacy-snapshot redaction that main added after this branched). Now split into version-gated schema migrations vs `runStartupMaintenance()` — maintenance runs on the current schema and is skipped only for future ones.

Validated on the rebased branch: focused race tests repeated, `go test -race ./internal/store`, server E2E, `go vet`, cross-compiles for `windows/amd64` + `darwin/arm64`, and `go test ./...` green with one exception: `TestManagerPanicSetsBackoff` in `internal/cloud/autosync`. That one is a pre-existing timing flake on main itself — reproduced on a clean checkout of 59cb8f2, where the test's 10 ms backoff can expire and return to `healthy` before the assertion observes `PhaseBackoff`. I'll report it separately.

Total vs main: ~1.1k lines across `internal/store`, mostly tests.

danielgap:
Rebased onto current main (`975490b`) and resolved the four blockers from the Aug-25 direction. Branch pushed to [danielgap/engram `fix/pr-613-review-blockers`](https://github.com/danielgap/engram/tree/fix/pr-613-review-blockers) so anyone can pull or cherry-pick:

- **Rebase**: 149 commits of drift resolved; startup-gate conflicts in `store.go` keep main's close-on-failure and DSN-pragma behavior.
- **Synchronization**: cold-start children now rendezvous through `.coldstart-ready-N` / `.coldstart-release` marker files instead of sleeps, so process ordering is proven rather than timing-dependent.
- **user_version**: re-read unconditionally after acquiring the inter-process lock; databases stamped by a newer engram fail closed.
- **Schema guard**: `TestSchemaVersionTracksMigrationBody` pins an AST-normalized fingerprint of `migrate()` per `schemaVersion`, so the suite cannot change without a version bump.
- Also reworded `34c292e` and `4e384d6` during the rebase: their messages described enrolled-repair-under-lock behavior the code never had.

One design decision needs your confirmation. `e8c52b9` (landed one day after the blocker list) made the `payload_sync_id` column add + backfill converge on every open, and its comment explicitly rejects version-marker skipping for that convergence ("a schema-version marker advanced past this point would skip it too"). A literal version gate breaks the four new `relation_deferred_identity` tests. The split keeps both contracts: `migrate()` holds the destructive, check-then-act suite (still gated and fingerprinted), while `convergeDeferredPayloadIdentity()` runs inside `runStartupMaintenance()` on every open unless the database is stamped by a newer engram. `TestMaintenanceConvergesWithoutRerunningMigrations` pins that a stamped-current reopen converges via maintenance without re-running the gated suite. If you'd rather drop the gate and converge everything under the lock, the split is cleanly revertible (top commit `65596a5`).

Validation on the pushed branch: `go build ./...`, `go vet ./internal/store/`, full `go test -race -count=1 ./internal/store/` green (excluding the item below), and a Windows cross-build of `internal/store`.

Heads-up, separate from this PR: `TestScanProject_PageProof` (from `9ab5e35`) hangs deterministically under `-race` on pure main `975490b` (90s+ timeouts, goroutines parked in `database/sql` awaits; passes in ~19s without `-race`). Happy to file a separate issue with the dump.

danielgap:
Follow-up from the race note above: filed #857 with the diagnosis. It is not a deadlock; the seeding loop needs ~510s under -race (26x the no-race cost) and the default 10m package timeout is what kills full-suite runs. Timings, stack excerpt, and two fix directions are in the issue.

dnlrsls:
@coderabbitia review

dnlrsls:
@coderabbit review

dnlrsls:
@coderabbit resume

Issue #534, opened 2026-06-22
UserPromptSubmit hook injects wrong tool prefix when engram is installed as a plugin


JantonioFC:
## Summary
The `UserPromptSubmit` hook (`scripts/user-prompt-submit.sh`, function `print_toolsearch_message`) injects a `ToolSearch select:` instruction using the `mcp__engram__*` tool prefix. This is correct when engram is registered as a **direct MCP server** named `engram`, but **wrong when engram is installed as a Claude Code plugin** (e.g. via the `engram@engram` marketplace).

When installed as a plugin, the real tool names are prefixed `mcp__plugin_engram_engram__*`. As a result, the injected `ToolSearch` matches **no tools**, and the first-message memory bootstrap silently fails — the agent never loads `mem_context`, `mem_save`, etc. on the first turn.

## Environment
- Claude Code, engram installed as plugin (`enabledPlugins: { "engram@engram": true }`)
- engram plugin cache version `0.1.0`
- Linux

## Steps to reproduce
1. Install engram as a Claude Code plugin (not as a direct MCP server).
2. Start a new session and send the first message.
3. Observe the injected `systemMessage`: `select:mcp__engram__mem_save,...`
4. The `ToolSearch` resolves to nothing because the actual tools are `mcp__plugin_engram_engram__*`.

## Suggested fix
The hook should detect the installation mode (plugin vs direct MCP server) and emit the matching prefix — e.g. resolve the prefix dynamically, or fall back to a name-agnostic `ToolSearch` keyword query (`"engram memory tools"`) instead of hardcoding `select:mcp__engram__*`.

## Note
Hardcoding either prefix breaks the other install mode. A keyword-based `ToolSearch` or runtime detection would cover both.

pablontiv:
Confirmed on macOS with a plugin install (engram 1.19.0, plugin 0.1.1) against Claude Code `2.1.216`: the deferred tools resolve as `mcp__plugin_engram_engram__*`, and a `select:mcp__engram__mem_search` matches nothing — so the first-message bootstrap loads no memory tools.

[...]

PR #642, opened 2026-07-20
fix(claude-plugin): use plugin-scoped MCP tool prefix in UserPromptSubmit hint

cristdar17:
## Problem

The UserPromptSubmit hook (`plugin/claude-code/scripts/user-prompt-submit.ps1` / `.sh`) emits a first-turn hint telling the agent to run:

```
ToolSearch select:mcp__engram__mem_save,mcp__engram__mem_search,...
```

But when engram is installed as a Claude Code **plugin** (via the marketplace), the MCP server is registered with the plugin-scoped prefix, so the actual tool names are `mcp__plugin_engram_engram__mem_*`. The `select:` query matches nothing and the agent cannot load the memory tools from the hint.

`internal/setup/setup.go` already handles both prefixes for the permissions allowlist — only the plugin hook scripts were left on the old prefix.

## Fix

Replace `mcp__engram__` with `mcp__plugin_engram_engram__` in both claude-code hook scripts (2 lines).

## Notes

- `plugin/codex/scripts/user-prompt-submit.sh` also references `mcp__engram__` — left untouched since Codex tool naming may differ; happy to update it here too if the same prefix applies.
- Verified locally on Windows 11 + Claude Code with the plugin installed from the marketplace: the corrected `select:` list loads all 13 tools.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated tool-loading guidance so memory tools can be discovered and invoked correctly.
  * Applied the fix consistently across supported command-line environments.

### DIFF
--- a/plugin/claude-code/scripts/user-prompt-submit.ps1
+++ b/plugin/claude-code/scripts/user-prompt-submit.ps1
@@ -19,7 +19,7 @@ function Write-EmptyHookResponse {
 }
 
 function Write-ToolSearchMessage {
-  $message = "CRITICAL FIRST ACTION — Execute this ToolSearch NOW before responding to the user:`nselect:mcp__engram__mem_save,mcp__engram__mem_search,mcp__engram__mem_context,mcp__engram__mem_session_summary,mcp__engram__mem_session_start,mcp__engram__mem_session_end,mcp__engram__mem_get_observation,mcp__engram__mem_suggest_topic_key,mcp__engram__mem_capture_passive,mcp__engram__mem_save_prompt,mcp__engram__mem_update,mcp__engram__mem_current_project,mcp__engram__mem_judge`n`nAfter loading tools, call mem_context to check for prior session history before responding."
+  $message = "CRITICAL FIRST ACTION — Execute this ToolSearch NOW before responding to the user:`nselect:mcp__plugin_engram_engram__mem_save,mcp__plugin_engram_engram__mem_search,mcp__plugin_engram_engram__mem_context,mcp__plugin_engram_engram__mem_session_summary,mcp__plugin_engram_engram__mem_session_start,mcp__plugin_engram_engram__mem_session_end,mcp__plugin_engram_engram__mem_get_observation,mcp__plugin_engram_engram__mem_suggest_topic_key,mcp__plugin_engram_engram__mem_capture_passive,mcp__plugin_engram_engram__mem_save_prompt,mcp__plugin_engram_engram__mem_update,mcp__plugin_engram_engram__mem_current_project,mcp__plugin_engram_engram__mem_judge`n`nAfter loading tools, call mem_context to check for prior session history before responding."
   [PSCustomObject]@{ systemMessage = $message } | ConvertTo-Json -Compress
 }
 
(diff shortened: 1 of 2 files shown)


pablontiv:
Thanks for catching this @cristdar17. Flagging some overlap and a couple of workflow blockers before it can land.

**Substance:** swapping `mcp__engram__*` → `mcp__plugin_engram_engram__*` fixes plugin installs but breaks direct-MCP installs (server id `engram`) — the mirror of the #192 (`3b99f0a`) change this is trying to undo. A single `select:` list carrying **both** prefixes fixes plugin installs without breaking direct-MCP ones; `ToolSearch select:` returns the names that exist and ignores the rest (verified live).

**Workflow (per CONTRIBUTING):**
- Needs `Closes #N` pointing at an issue with `status:approved`; #534 isn't approved and there's no `Closes` link here, so the PR-validation checks will block it.
- Needs exactly one `type:*` label.
- It trips the existing `TestClaudeCodeUserPromptHookUsesCurrentMCPServerID` (`internal/setup/setup_test.go`), which forbids `select:mcp__plugin_engram_engram__` and requires the `mcp__engram__*` names — so swapping the prefix removes required names and adds a forbidden one.

I've opened #653 / #654 for the broader Claude Code hook chain (systemMessage delivery, dead SubagentStop capture, matcher) that folds in the dual-prefix fix and updates those tests to require both prefixes. Happy to defer to whichever the maintainer prefers — closing this in favor of the consolidated fix, or rebasing around it. Just flagging so we don't double-merge the same function.

franncov:
This fixes plugin installs, but it moves the bug rather than removing it. Hardcoding `mcp__plugin_engram_engram__` breaks the standalone/user-level install, where the server id is `engram` and the tools are exposed as `mcp__engram__*`. That selector now becomes the one that can never match.

Both shapes are supported today, so a fixed string can only ever be right for one of them. `internal/setup/setup.go:107` allowlists both prefixes precisely for this reason, and `docs/AGENT-SETUP.md:278` documents the durable user-level MCP config.

Detecting the install shape covers both. `CLAUDE_PLUGIN_ROOT` is set when the hook runs from the plugin, with the script path as a fallback:

```bash
if [ -n "${CLAUDE_PLUGIN_ROOT:-}" ]; then
  ENGRAM_TOOL_PREFIX="mcp__plugin_engram_engram__"
else
  case "$0" in
    */plugins/cache/*|*/plugins/marketplaces/*)
      ENGRAM_TOOL_PREFIX="mcp__plugin_engram_engram__" ;;
    *)
      ENGRAM_TOOL_PREFIX="mcp__engram__" ;;
  esac
fi
```

One constraint worth keeping: `print_toolsearch_message` is also called from the Windows bash-builtin-only fast path, so the detection has to avoid forks and command substitution. The snippet above stays inside that limit, and the equivalent in the `.ps1` can key off `$env:CLAUDE_PLUGIN_ROOT` and `$PSCommandPath`.

I have this running locally against both install shapes. Happy to send it as a patch on top of your branch if that is useful.

dnlrsls:
Thanks for identifying the plugin-scoped tool-name mismatch. The diagnosis is valid, but this patch replaces one hardcoded namespace with the other, which breaks direct MCP installs and fails TestClaudeCodeUserPromptHookUsesCurrentMCPServerID.

The complete fix is being consolidated in #654, which supports both namespaces, fixes UserPromptSubmit model-context delivery, and adds regression coverage. Closing this as superseded so we do not carry two overlapping implementations. Thanks again for surfacing the issue.

PR #673, opened 2026-07-27
fix(pi): archive current Pi compaction shape and avoid duplicate save

### LINKED ISSUE #951: fix(pi): preserve compaction recovery across current payloads and ambiguous archive outcomes
### 📋 Pre-flight Checks

### 📝 Bug Description

The Pi integration does not fully preserve compaction recovery across the current runtime payload and failure outcomes. Current Pi emits the compacted summary as:

```ts
{ compactionEntry: { summary: "..." }, fromExtension, reason, willRetry }
```

`extractCompactedSummary` does not inspect `compactionEntry.summary`, so the summary is skipped and no automatic archive is attempted. The next turn is then always instructed to save manually.

The lifecycle also has two related reliability gaps: strict session registration or provider initialization can return before fallback guidance is queued, and a timed-out archive request can be treated like a definite failure even though the write may already have reached Engram. That can either lose recovery guidance or direct a duplicate manual save.

### 🔄 Steps to Reproduce

1. Run Engram `2.0.0 rc4` with the Pi integration on Windows 11.
2. Trigger a Pi `session_compact` event whose summary is present at `compactionEntry.summary`.
3. Observe that the current extractor does not return the summary and `/observations` is not called.
4. Trigger a registration or initialization failure and observe that the hook can return before queuing manual recovery guidance.
5. Let the archive POST time out and observe that the outcome is not distinguished from a definite failure.

### ✅ Expected Behavior

The Pi adapter should extract `compactionEntry.summary` before legacy fallback fields, archive only after the exact runtime session is acknowledged, and always preserve next-turn recovery guidance. A confirmed archive should suppress manual saving; a definite failure should retain the manual fallback; and a timeout with an unknown write outcome should require verification before any retry. Compaction context should be limited to the exact runtime session.

### ❌ Actual Behavior

The current Pi payload is missed, strict registration or initialization failures can prevent fallback injection, and timeout ambiguity can direct an unnecessary duplicate save. Compaction recovery also reads project-wide context instead of the existing session-scoped compaction context endpoint.

### Operating System

Windows

### Engram Version

2.0.0 rc4

### Agent / Client

Other

### 📋 Relevant Logs

_No response_

### 💡 Additional Context

Agent/client: Pi. Windows edition: Windows 11.

PR #673 contains the initial event-shape correction but links no issue and needs adaptation to the current session lifecycle. Issue #381 was previously closed as covered by PR #368 with an explicit request to open a fresh issue carrying the exact missed Pi payload.

noxos-os:
Closes #951

> Review size: 435 authored changed lines. Maintainer-approved size:exception rationale: 216 lines are executable Pi lifecycle coverage, and separating the tests from the behavior would break the cohesive verification and rollback boundary.

## Problem

`gentle-engram`'s `session_compact` handler never auto-archived the compaction summary to Engram on **current Pi**, because `extractCompactedSummary` only inspected legacy top-level / nested paths and missed Pi's documented event shape:

```ts
// Pi session_compact event (docs: docs/compaction.md)
{ compactionEntry: CompactionEntry, fromExtension, reason, willRetry }
// CompactionEntry.summary holds the summary text
```

As a result, `summary` was `undefined`, the `/observations` POST was skipped, and the agent was **always** asked to save the summary manually via the `FIRST ACTION REQUIRED: mem_session_summary` notice on the next turn — even though that was meant to be a fallback.

Separately, `buildRecoveryNotice` always appended the manual-save instruction, so once extraction is fixed the agent would duplicate the save it just completed.

## Fix

**`plugin/pi/compaction-recovery.js`**
- Add `["compactionEntry", "summary"]` as the first `SUMMARY_FIELD_PATH` so the documented event shape is extracted (legacy fallbacks are preserved).
- `buildRecoveryNotice(project, context, persisted = false)`: when `persisted`, emit a continuity-only acknowledgement (still surfaces retrieved Engram context) and **omit** the manual `FIRST ACTION REQUIRED: mem_session_summary` instruction. Default `false` keeps existing behavior for any caller.

**`plugin/pi/index.ts`**
- In `session_compact`, treat the archive POST's confirmed response as `persisted` and pass it to `buildRecoveryNotice`. `bestEffortEngramFetch` returns `null` on a failed/unreachable write, so a failed archive still falls back to the manual save instruction — only a confirmed success suppresses it.

## Behavior matrix

| summary extracted | archive succeeded | notice on next turn |
|---|---|---|
| yes | yes | context + "already saved" (no manual save) |
| yes | no (server down) | context + manual `FIRST ACTION REQUIRED` (retry) |
| no | n/a | context + manual `FIRST ACTION REQUIRED` (unchanged) |

## Tests

- `extractCompactedSummary` now covers `{ compactionEntry: { summary } }` (native + extension-provided, plus whitespace-only ignored).
- `buildRecoveryNotice(..., true)` asserts no `FIRST ACTION REQUIRED` while still surfacing context.

Full local run: `node --test plugin/pi/test/*.test.mjs` → **44 pass / 0 fail**.

Backward compatible: `buildRecoveryNotice` third arg defaults to `false`, and the new extraction path is strictly additive.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery after session compaction by correctly recognizing current summary formats.
  * Recovery notices now accurately indicate whether the summary was saved successfully.
  * Empty summaries are handled without generating misleading recovery instructions.

* **Tests**
  * Added coverage for current compaction events, empty summaries, and saved-summary notifications.

### DIFF
--- a/plugin/pi/compaction-recovery.js
+++ b/plugin/pi/compaction-recovery.js
@@ -1,4 +1,9 @@
 const SUMMARY_FIELD_PATHS = [
+  // Current Pi session_compact event shape:
+  //   { compactionEntry: CompactionEntry, fromExtension, reason, willRetry }
+  // where CompactionEntry.summary holds the summary text. Listed first so the
+  // documented shape wins over generic legacy fallbacks.
+  ["compactionEntry", "summary"],
   ["summary"],
   ["compactedSummary"],
   ["compacted_summary"],
@@ -47,6 +52,11 @@ export function extractCompactedSummary(event) {
   return undefined;
 }
 
+/**
+ * Manual fallback used when the compaction summary could NOT be archived to
+ * Engram during the session_compact handler (no summary extracted, no session,
+ * or the Engram write failed). Asks the agent to save it on the next turn.
+ */
 export function recoveryInstruction(project) {
   return (
     `CRITICAL INSTRUCTION FOR COMPACTED SUMMARY:\n` +
@@ -57,8 +67,30 @@ export function recoveryInstruction(project) {
   );
 }
 
-export function buildRecoveryNotice(project, context) {
-  const instruction = recoveryInstruction(project);
+/**
+ * Used when the compaction summary was already archived to Engram by
+ * gentle-engram. Still surfaces retrieved project context for continuity, but
+ * omits the manual `FIRST ACTION REQUIRED: mem_session_summary` instruction so
+ * the agent does not duplicate the save it just completed.
+ */
+function persistedAcknowledgement(project) {
+  return (
+    `Compaction recovery summary was already saved to Engram (${project}) by gentle-engram. ` +
+    `No manual mem_session_summary call is needed for this compaction. ` +
+    `Call mem_context if you need additional recent project memory.`
+  );
+}
+
+/**
+ * Build the recovery notice injected before the next agent turn.
+ *
+ * @param {string} project Engram project name.
+ * @param {string|undefined} context Retrieved Engram project context, if any.
+ * @param {boolean} [persisted=false] True when the compaction summary was
+ *   already archived to Engram; suppresses the manual save instruction.
+ */
+export function buildRecoveryNotice(project, context, persisted = false) {
+  const instruction = persisted ? persistedAcknowledgement(project) : recoveryInstruction(project);
   const trimmedContext = typeof context === "string" ? context.trim() : "";
   return trimmedContext ? `${trimmedContext}\n\n${instruction}` : instruction;
 }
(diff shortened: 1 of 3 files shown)


dnlrsls:
Thanks for identifying the current `compactionEntry.summary` shape and preparing the initial fix.

The complete recovery work has now landed in #989, including exact runtime-session attribution, archive outcome handling, and session-scoped recovery notices. Since #989 supersedes this change and #951 is closed, I am closing this PR to keep the queue clear.

Your report and implementation direction helped us reach the final fix. Thank you.

PR #685, opened 2026-07-30
feat(store,server,claude-code): bound the SessionStart context injection

### LINKED ISSUE #163: perf: SessionStart hook injects ~10 KB (~2.5k tokens) into every Claude Code session

adrian-ruda:
## 🔗 Linked Issue

Closes #163

---

## 🏷️ PR Type

> I can't apply labels from outside the repo (needs triage permission), so **"Check PR Has type:* Label" will fail until a maintainer adds `type:feature`**. Flagging it so it doesn't read as a broken PR.

---

## 📝 Summary

- Makes every section of the context blob boundable: `ContextOptions{Observations, Prompts, Sessions, Pinned, Compact}`, one convention for all four ints — **0 keeps the legacy default · >0 caps · <0 omits the section and its header**.
- Caps the pinned section, which has had no `LIMIT` at all since #484 and was the one part of the blob that could grow without any ceiling.
- **Defaults are byte-identical.** `FormatContext` keeps its signature and output; a zero-value `ContextOptions` reproduces today's bytes exactly, and a test asserts precisely that.

## 📂 Changes

| File | What changed |
|---|---|
| `internal/store/store.go` | `ContextOptions` + `FormatContextWithOptions`; `FormatContext` becomes a thin wrapper; unexported pinned query with `LIMIT` (exported `PinnedObservations` signature untouched) |
| `internal/store/store_test.go` | `TestFormatContextWithOptions` — 11 subtests: byte-identity for zero value, cap/omit per section with cross-checks that siblings are untouched, compact, pinned cap |
| `internal/server/server.go` | `GET /context` accepts `observations`, `prompts`, `sessions`, `pinned`, `compact`, reusing the existing `queryInt`/`queryBool` helpers |
| `internal/server/server_e2e_test.go` | `TestContextQueryParamsE2E` — no params ≡ legacy, caps shrink, negative omits, garbage input still 200 |
| `plugin/claude-code/scripts/session-start.sh`, `post-compaction.sh` | *(separable last commit)* hooks request `compact=1&pinned=20` |

## 📊 Measured on a 2,976-observation project

Real install, `project` scope, ~4 months of daily use — measured against a snapshot of that database, not synthetic data:

| Request | Bytes | vs today |
|---|---|---|
| today (no params) | 9,252 | — |
| `prompts=-1` | 7,946 | −14% |
| `observations=10` | 5,395 | −42% |
| **`compact=1`** | **3,152** | **−66%** |
| `compact=1&prompts=-1` | 1,846 | −80% |
| `compact=1&prompts=-1&observations=10` | 1,039 | −89% |

End to end, running the hook scripts themselves against that database: **9,404 B → 3,179 B per session start**.

That last combo lands at −89%, which reproduces @todie's −88% from #161/#162 on an install roughly 15× larger — the win holds as history grows, because every section is a fixed row count regardless of database size.

## 🤔 The default I chose, and why it's conservative

The hook commit asks only for `compact=1&pinned=20`:

- **`compact=1`** drops the 300-char body preview from observation bullets. **No row disappears** — every type and title still renders, and the body is one `mem_get_observation` away when the agent actually wants it. That alone is −66%.
- **`pinned=20`** gives the blob a real ceiling. Without it there is still no hard cap, since a user who pins 40 observations to make context *better* silently makes the injection several times larger.

I deliberately did **not** default `prompts=-1`, even though it's another −14%: your v1.19.0 notes describe the context block as "load-bearing", so which sections earn their bytes is your call, not mine. The table above is there so you can pick a stronger default in one line if you want it.

## 🙏 Credit

The `ContextOptions` + thin-wrapper shape is **@todie's design from #162**, rebased onto current `main` and credited with `Co-authored-by:` on every commit. #161/#162 were closed on process, not merit; @todie filed first, did the original measurement, and hasn't been active since April. This picks it up and extends it to what #162 predates: the hardcoded `RecentPrompts(project, 10)` and the unbounded pinned section from #484. If @todie wants to take it back, I'll close this.

## ✅ Verification

Run locally on this branch, plus a clean `upstream/main` worktree as a baseline for comparison:

- `go test ./...` — 22 packages ok, 0 failures, identical to the baseline run
- `go test -tags e2e ./internal/server/...` — ok
- `gofmt -l` clean · `go vet -tags e2e ./internal/server/...` clean
- Hook scripts checked with `bash -n` and executed against an isolated daemon (`ENGRAM_DATA_DIR` pointed at a snapshot, never the live database)

## 📋 Note for review

The three commits are meant to be separable: store (mechanism), server (transport), hooks (the only behavior change). Happy to reshape any of it — including switching to a hard server-side ceiling that no caller can exceed, if you'd rather the bound not be up to the caller at all.

## Summary by CodeRabbit

* **New Features**
  * Enhanced `GET /context` with per-section limits and compact output that reduces inline previews.
  * Added controls to include or omit observations, prompts, sessions, and pinned items.
  * Invalid or excessive limits fall back safely without errors.
* **Improvements**
  * Session startup and post-compaction retrieval now request compact context with up to 20 pinned items.
* **Documentation**
  * Updated `GET /context` documentation with parameter behavior and fallback rules.
* **Tests**
  * Added coverage for limits, omission, compact output, and invalid values.


adrian-ruda:
Addressed all three review comments in `7b3a09d`:

- **Duplicated bullet rendering** — extracted to `writeObservationBullet`, shared by both loops. Fair catch: the two copies were byte-identical, so a future change to `truncate(obs.Content, 300)` or any redaction would have landed in one section only.
- **Error paths untested** — added `TestFormatContextWithOptionsErrorBranches`, four subtests that close the store first and use the option convention to skip the earlier sections (`{Sessions: -1, Pinned: -1}` to reach the observations branch, and so on), so each of the four `return "", err` is actually exercised.
- **Undocumented params** — `DOCS.md` now lists all five on `GET /context`, with the `0 / >0 / <0` convention and the fact that invalid values fall back to the default instead of returning 400.

Output is unchanged — the byte-identity test still passes without being touched. `go test ./...` (22 packages) and `go test -tags e2e ./internal/server/...` both green.

On the inconclusive **Linked Issues** check: it's right that this PR doesn't implement the env-tunable limits described in #163's original writeup. That's deliberate — @Alan-TheGentleman's approval narrowed the scope to "a bounded-size injection. The scope is the size cap", and I offered in [that thread](https://github.com/Gentleman-Programming/engram/issues/163#issuecomment-5131286786) to build a hard server-side ceiling instead of (or on top of) per-caller params if that's the shape he prefers. I'd rather wait for that call than widen the diff on a guess.

adrian-ruda:
Rebased onto current `main` — the branch is now 4 ahead / 0 behind, no conflicts.
`go build ./...`, `go vet ./...` and `go test ./...` all pass locally on go1.26.

Two things still need a maintainer here: the `type:feature` label (the label check
can't pass without it, and I can't apply labels from outside the repo), and approval
to run the workflows on a fork PR.

No rush — flagging it in case it fell off the radar.

adrian-ruda:
Rebased onto main (9428a21) and addressed both CodeRabbit findings.

**Rebase.** The four commits are unchanged in intent — the diff is byte-identical to the pre-rebase one (7 files, +574/−21). Three upstream changes landed in the same lines and were merged, not overridden: `resolveRequestProject` (#875), the reworked `FormatCompactionContext` (#797), and the `CONTEXT=""` guard in the session hooks.

**CodeRabbit #1 — ceiling on positive limits.** Fixed. The repo already had the convention (`clampConflictsLimit`, `conflictsMaxLimit = 500`), so `clampContextLimit` follows it with the same ceiling. One deliberate difference: `clampConflictsLimit` collapses `<= 0` to the default, which would have destroyed this PR's three-state contract. `clampContextLimit` only caps from above — `0` (legacy default) and negatives (omit the section) pass through untouched.

**CodeRabbit #2 — HTTP-boundary coverage.** Fixed. `sessions` and `pinned` were exercised in `store_test.go` but never through the URL, so a typo in a param name would have passed every test. Added `sessions=-1`, `pinned=-1`, `pinned=1`, `sessions=1` and `observations=999999`, with pinned fixtures created up front so the absence assertions can't pass on an empty section.

Local gates against a clean `upstream/main` baseline: no regressions. `internal/store` and `internal/server` pass, including with `-tags=e2e` (worth noting: `go test ./...` alone skips `server_e2e_test.go`). The vet warning and the failures in `cmd/engram` and `internal/project` reproduce identically on an untouched main.

On the measurements in #163: two people independently reproduced the cost on their own installs — one showed the eager blob leaking cross-project context into 3 of 6 unrelated answers (0 of 6 without it), the other showed the payload is flat regardless of corpus size, at ~312k tokens across 119 sessions. Their numbers are better evidence than the original report.

#163 carries `status:approved`, and this PR is the focused re-file with the link that was asked for when #161/#162 were closed. Still unlabeled — the "Has type:* Label" check needs `type:feature` from someone with triage rights, and the fork workflows need a maintainer to approve the run.

dnlrsls:
Thanks for the work and measurements here. A maintainer-owned replacement is now open at #1012, preserving this branch's commits and attribution while adding the reviewed total-byte safety cap. Closing this PR as superseded so review can continue in one place.

Issue #668, opened 2026-07-26
Save-nudge never fires (or fires too eagerly) outside UTC+0, and never fires for a project's first observation


Alexter2003:
### 📋 Pre-flight Checks
- [x] I have searched existing issues and this is not a duplicate
- [x] I understand this issue needs `status:approved` before a PR can be opened

### 📝 Bug Description

The save-nudge mechanism (the `MEMORY REMINDER` system message injected by the `UserPromptSubmit` hook to prompt the agent to call `mem_save`) never fires in practice for non-UTC hosts, and never fires at all for a project's very first observation — regardless of timezone. Across 4 real projects and 15 sessions on my machine, this produced **0 observations ever saved**, despite substantial work happening in every session.

Two independent root causes, both in `parse_epoch()` / the nudge-gating logic, shared verbatim between `plugin/claude-code/scripts/user-prompt-submit.sh` and `plugin/codex/scripts/user-prompt-submit.sh` (the `plugin/opencode/engram.ts` implementation only has the second bug — its `toEpochSecs` already normalizes to UTC correctly):

**1. `parse_epoch()` parses naive sqlite timestamps as local time instead of UTC.**

`sessions.started_at` and `observations.created_at` are written by sqlite as `datetime('now')`, which returns `YYYY-MM-DD HH:MM:SS` in UTC with no timezone marker. The final fallback branch of `parse_epoch()` parses this string without `-u`:

```bash
date -j -f "%Y-%m-%dT%H:%M:%S" "$TS" "+%s" 2>/dev/null \
  || date -j -f "%Y-%m-%d %H:%M:%S" "$TS" "+%s" 2>/dev/null \
  || date -d "$TS" "+%s" 2>/dev/null
```

`date` interprets the naive string in the host's local timezone, so the computed epoch is skewed by the host's UTC offset:
- **Negative offsets** (e.g. UTC-6, Latin America): the parsed epoch is pushed *into the future* relative to the real UTC epoch, so `SESSION_AGE_SECS` / `ELAPSED` come out negative or near-zero. The session/last-save always looks "just started," and the nudge condition (`elapsed > 900`) can never become true. **The nudge never fires, for any project, ever.**
- **Positive offsets** (e.g. UTC+1/+2, Spain/Europe): the parsed epoch is pushed *into the past*, so elapsed time is inflated. The nudge fires far too eagerly — potentially right after the 5-minute session-age floor, and repeating every cooldown window even right after a save.
[...]

PR #702, opened 2026-08-06
fix(mcp): track save freshness by project

### LINKED ISSUE #703: mem_search reports stale no-save warning after explicit-session save
### 📋 Pre-flight Checks

### 📝 Bug Description

`mem_search` and `mem_context` can return a freshly saved observation while simultaneously appending:

```text
⚠️ No mem_save calls for this project in N minutes.
```

This happens when `mem_save` uses an explicit `session_id`. The save is recorded in `SessionActivity` under that concrete session ID, while project read tools evaluate the nudge under `defaultSessionID(project)`. The observation store and the activity tracker therefore disagree.

This is related to #668 because both affect save nudges, but it is a different server-side root cause. #668 covers client-hook timestamp parsing and first-observation bootstrap behavior.

### 🔄 Steps to Reproduce

1. Create activity older than the configured nudge threshold under `defaultSessionID(project)`.
2. Call `mem_save` for the same project with a different explicit `session_id`.
3. Immediately call `mem_search` for the newly saved title with an explicit project.
4. Observe that the new memory is returned together with the stale no-save warning.

### ✅ Expected Behavior

A successful save in any session for a project should refresh the project's read-tool save freshness. Immediate `mem_search` and `mem_context` responses should not include the inactivity warning. Per-session activity scores must remain isolated.

### ❌ Actual Behavior

The newly saved observation is visible, but the response also claims that no `mem_save` call occurred for the project. Updating the synthetic default session directly would hide the warning but incorrectly increments its per-session save count.

### Operating System

Linux (Ubuntu/Debian)

### Engram Version

1.20.0

### Agent / Client

OpenCode

### 📋 Relevant Logs

```shell
Found 1 memories:

[1] fresh memory

⚠️ No mem_save calls for this project in 15 minutes.
```

### 💡 Additional Context

The issue reproduces on current `main` and v1.20.0. A regression-tested fix keeps project save freshness separate from per-session activity scores, covers both `mem_search` and `mem_context`, records `mem_session_summary` freshness, and lazily expires stale project entries.

Draft implementation: #702. The PR will remain draft until this issue receives `status:approved`.

sparkiesweden:
## Summary
- track recent successful saves separately at project scope for read-tool nudges
- keep per-session activity scores isolated when saves use explicit session IDs
- suppress contradictory `mem_search`/`mem_context` warnings after `mem_save` or `mem_session_summary`
- lazily expire project freshness entries after the configured nudge interval

## Root cause
`mem_save` records activity under the concrete observation session ID, while `mem_search` and `mem_context` evaluate nudges under `defaultSessionID(project)`. A fresh observation could therefore be returned together with `No mem_save calls for this project ...`.

Updating the synthetic default session would make the warning disappear, but it breaks the existing explicit-session ActivityScore contract. This change keeps project freshness as separate state instead.

Related history: #449 improved active-session routing for saves, but does not address project-level read nudges.
Related nudge issue: #668 covers client-hook timezone and first-observation behavior, not this server-side identity mismatch.

Tracked by #703. This PR remains draft until that issue receives `status:approved`; `Closes #703` will be added after approval.

## Test plan
- TDD regression: explicit-session save → immediate project search returns the observation without stale warning
- context regression: explicit-session save suppresses immediate project context warning
- session-summary regression: summary save refreshes project read activity without incrementing default-session save score
- unit coverage for expiration and session-score isolation
- targeted activity/save/nudge suite under Go 1.25 Docker
- patch applies and tests cleanly against v1.20.0; live-verified through Hermes on two hosts

## Notes
The full `internal/mcp` suite in an anonymous checkout has unrelated existing project-detection fixture failures because the checkout's detected project is not backed in test stores. The targeted affected suite is green, and the fix received two review passes with no remaining Critical or Important findings.

### DIFF
--- a/internal/mcp/activity.go
+++ b/internal/mcp/activity.go
@@ -14,10 +14,11 @@ const ambiguousProjectRecoveryTTL = 5 * time.Minute
 
 // SessionActivity tracks tool call activity for save reminders and activity scores.
 type SessionActivity struct {
-	mu         sync.Mutex
-	sessions   map[string]*sessionState
-	nudgeAfter time.Duration
-	now        func() time.Time // injectable for testing
+	mu           sync.Mutex
+	sessions     map[string]*sessionState
+	projectSaves map[string]time.Time
+	nudgeAfter   time.Duration
+	now          func() time.Time // injectable for testing
 }
 
 type sessionState struct {
@@ -44,9 +45,10 @@ type ambiguousProjectRecovery struct {
 // NewSessionActivity creates a new activity tracker with the given nudge threshold.
 func NewSessionActivity(nudgeAfter time.Duration) *SessionActivity {
 	return &SessionActivity{
-		sessions:   make(map[string]*sessionState),
-		nudgeAfter: nudgeAfter,
-		now:        time.Now,
+		sessions:     make(map[string]*sessionState),
+		projectSaves: make(map[string]time.Time),
+		nudgeAfter:   nudgeAfter,
+		now:          time.Now,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


sparkiesweden:
Process update: dedicated issue #703 is open and this PR is now draft pending its `status:approved` label. I cannot apply the required `type:bug` PR label as an external contributor (HTTP 403); please add it during triage.

PR #705, opened 2026-08-08
fix(codex): wrap SessionStart hook output in hookSpecificOutput envelope

uhhexe:
plugin/codex/scripts/session-start.sh and post-compaction.sh emit the Memory Protocol text as raw markdown via `cat <<'PROTOCOL'`, not Codex's expected `{"hookSpecificOutput":{"hookEventName":...,"additionalContext":...}}` JSON envelope. Codex's SessionStart hook parser rejects raw text as a hook failure — non-fatal (the turn still completes), but every run reports `hook: SessionStart Failed` for these two hooks. Both register under the SessionStart event (hooks.json: `startup|resume|clear` for session-start.sh, `compact` for post-compaction.sh), so `hookEventName` is `SessionStart` for both. The claude-code/ variant is untouched — Claude Code's SessionStart contract is tolerant of raw text, so it isn't affected.

Content is now written to temp file(s) and read into `jq` via `--rawfile`, never captured with `$(cat <<'EOF' ...)`. macOS ships bash 3.2 as /bin/bash, and bash 3.2 mis-parses a heredoc nested inside a command substitution once the heredoc body contains an apostrophe (e.g. "user's") — it throws `unexpected EOF while looking for matching `)'` at script-load time. Both protocol blocks contain that apostrophe, so a naive `$(heredoc)` fix broke immediately under stock macOS bash; routing through a file sidesteps the bug.

Verified locally:
```
echo '{"session_id":"x","cwd":"<repo>","hook_event_name":"SessionStart","source":"startup"}' | bash plugin/codex/scripts/session-start.sh
```
(and post-compaction.sh with `source:"compact"`) now produce valid JSON with `hookSpecificOutput.hookEventName == "SessionStart"`, exit 0. `bash -n` and `shellcheck -S warning` clean on both files.

Already landed and running on our own fork (uhhexe/engram#1); opening here as a contribution back upstream.

## Summary by CodeRabbit

- **Improvements**
  - Improved session startup and recovery after compaction.
  - Protocol guidance and relevant project context are now delivered together more reliably.
  - Reduced the risk of missing or malformed session instructions during hook execution.
  - Temporary processing files are cleaned up automatically.

### DIFF
--- a/plugin/codex/scripts/post-compaction.sh
+++ b/plugin/codex/scripts/post-compaction.sh
@@ -31,8 +31,24 @@ fi
 ENCODED_PROJECT=$(printf '%s' "$PROJECT" | jq -sRr @uri)
 CONTEXT=$(curl -sf "${ENGRAM_URL}/context?project=${ENCODED_PROJECT}" --max-time 3 2>/dev/null | jq -r '.context // empty')
 
-# Inject Memory Protocol + compaction instruction + context
-cat <<'PROTOCOL'
+# Build Memory Protocol + compaction instruction + context, then emit it as
+# Codex's hookSpecificOutput JSON envelope. Codex's SessionStart parser (this
+# hook registers under the SessionStart "compact" matcher — see hooks.json)
+# rejects raw stdout text as a hook failure (non-fatal, but reports
+# "hook: SessionStart Failed" every run — see codex-review SKILL.md "Failure
+# modes", 2026-08-08 silent-death-after-hooks). Claude Code's SessionStart
+# contract stays tolerant of raw text, so this is a codex/-only change;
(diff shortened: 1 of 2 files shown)


dnlrsls:
Thanks for the contribution. We revalidated the reported failure against the Codex SessionStart contract and confirmed that successful plain stdout is already accepted as model context, so raw Markdown is not the demonstrated cause of the failure. Without a reproducible failing case and a contract test, this change would add new jq and temporary-file failure paths without fixing proven behavior.

Closing for now. We would be happy to reconsider a focused fix with the exact Codex version, hook input, exit status and stderr, parser evidence, and a regression test that fails on the current implementation.

Issue #701, opened 2026-08-06
bug(sync): project chunk export writes relation mutations whose endpoints never travel (personal-scope / pre-watermark observations)


elorenzo368:
### 📋 Pre-flight Checks

- [x] I have searched existing issues and this is not a duplicate
- [x] I understand this issue needs `status:approved` before a PR can be opened

### 📝 Bug Description

`engram sync` (local git chunk export) writes relation mutations into project-scoped chunks whose endpoint observations are **not in the chunk and will never be exported**:

1. **Personal-scope endpoints**: relations created between project observations and `scope: personal` observations get exported into the project chunk, but personal observations never travel in project chunks by design.
2. **Pre-watermark endpoints**: relations whose endpoint observation was created before the incremental export watermark are exported, but the endpoint itself is not re-exported.

The result is a chunk that is **orphaned at birth**: every receiving machine stalls on it with `relation FK precondition not met: referenced observation missing` (the import-side stall is tracked in #649). And because incremental export is watermark-based, running `engram sync` again on the source machine does NOT re-export the missing endpoints, so the system can never self-heal. The only fixes are manual chunk surgery or an out-of-band export/import.

This is arguably fallout from #353: relations used to be silently dropped from chunk export; now they are included, but without guaranteeing their endpoints can be resolved on the receiving side.

### 🔄 Steps to Reproduce

1. On machine A, save a personal-scope observation (`scope: personal`) and a project observation, then create a relation between them (e.g. via `mem_judge` / conflict scan).
2. Run `engram sync` inside the project repo. The generated project chunk contains the relation mutation but not the personal endpoint.
3. Commit and push `.engram/`. On machine B, pull and run `engram sync --import`.
4. Machine B stalls: `dependency-safe local import stalled after N pass(es); chunk <id>: apply chunk mutation <n>: relation FK precondition not met: referenced observation missing`.
5. Back on machine A, run `engram sync` again: it reports nothing new to export (watermark), so the missing endpoints are never shipped. The stall on B is permanent.

### ✅ Expected Behavior

Chunks are self-contained and appliable: export either (a) includes the endpoint observations required by any exported relation (even if personal-scope or pre-watermark), or (b) skips relations whose endpoints cannot travel, with a visible warning. Either way, a receiving machine can always drain the chunk.

### ❌ Actual Behavior

Real cross-machine case (both machines engram 1.20.0, same repo, git-based sync):

[...]

Issue #615, opened 2026-07-17
fix(sync): export never picks up observations older than its watermark — "all memories already exported" with the project at zero chunks


pqem:
### 📋 Pre-flight Checks
- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/engram/issues) and this is not a duplicate
- [x] I understand this issue needs `status:approved` before a PR can be opened

### 📝 Bug Description
`engram sync` never exports observations that predate its export watermark. For those it reports **"all memories already exported"** even when the project has **zero chunks** on disk. Everything created *after* the watermark exports fine, so day-to-day sync looks healthy while an arbitrarily large backlog stays invisible to the fleet forever.

This is related in spirit to #447 (silent "Nothing new to sync" for edited observations) but affects observations that were **never exported at all**.

Measured on two machines (chunks opened and counted one by one, 33 chunks total):

| Machine | Observations in DB | Present in chunks | Never exported |
| --- | --- | --- | --- |
| Windows desktop | 546 | 304 | 242 |
| Linux NUC | 411 | 302 | 109 |

`engram export <file>` is a different code path and is NOT affected: it dumps all 546 / 411 correctly.

### 🔄 Steps to Reproduce
1. Use engram on a machine where git sync is not yet initialized (`~/.engram` has `engram.db` but no repo). Accumulate observations over time.
2. Initialize git sync later (clone the memory repo, materialize `manifest.json` + `chunks/`).
3. Run `engram sync` (also tried `engram sync --project <name>` and `engram sync --all`).
4. Compare `mem_stats` / DB observation count against the content of the exported chunks.

### ✅ Expected Behavior
`engram sync` exports every observation that is not present in any chunk, regardless of when it was created. If it decides to skip observations, it should say how many were skipped and why — not report "already exported" for data that exists in no chunk.

### ❌ Actual Behavior
- `engram sync` responds `all memories already exported` and exports nothing (`Observations: 0`).
- `engram sync --project X`, where project X has 94 observations in the DB and **0 chunks**, also responds "already exported".
- Observations created after the last successful sync DO export correctly — only the pre-watermark backlog is skipped.
- Behavior persists in **1.19.0**: tested against a copy of a real `~/.engram` with an isolated release binary (sha256 verified). `sync` exported `Observations: 0` with 242 unexported observations in the DB.

### Operating System
Linux (Linux Mint 22 / kernel 6.17) and Windows 10 — same behavior on both.

### Engram Version
[...]

PR #715, opened 2026-08-12
fix(sync): isolate invalid relation imports

### LINKED ISSUE #649: bug(sync): import stalls forever on self-referential and orphan relation mutations
### 📋 Pre-flight Checks

### 📝 Bug Description

`engram sync --import --all` stalls permanently when a chunk contains a **self-referential** relation (`source_id == target_id`) or an **orphan** relation (an endpoint observation that exists in no chunk and not locally). The dependency-safe import loop retries the offending chunk every pass, never makes progress, and aborts. `engram doctor` reports the store healthy and `sync_apply_deferred` stays empty (0 rows) — the deferral safety net is never reached on this path.

**Two independent root causes in `internal/store/store.go`:**

**1. Self-referential relations can never apply.** `applyRelationUpsertTx` checks the FK precondition with `SELECT count(*) FROM observations WHERE sync_id IN (?, ?)` over `(source, target)` and requires `count >= 2`. When `source_id == target_id`, `IN (?, ?)` collapses to one distinct id, so `count` is `1` even though the observation exists → `1 < 2` → `ErrRelationFKMissing` forever. These are valid, judged relations (e.g. system "artifact completeness" self-references).

**2. Chunk import has no FK-miss/dead resilience.** The per-mutation cloud pull path (`ApplyPulledMutation`) already implements the design §9 skip+log policy: a relation `ErrRelationFKMissing` is written to `sync_apply_deferred` and the seq is ACK'd so the cursor advances. But `ApplyPulledChunk` — used by the local/file import path — aborts the entire chunk on **any** `applyPulledMutationTx` error. So a single orphan/self-referential relation kills the whole chunk, the dependency-safe loop can never drain it, and the import stalls. Orphan relations (endpoint dropped by per-project chunk export) are permanently unresolvable and should become `dead`, not block every import.

### 🔄 Steps to Reproduce

1. Create two observations and a relation with `source_id == target_id` (or a relation referencing a nonexistent observation).
2. Export to a chunk; run `engram sync --import --all` into a store that lacks the endpoint observation.
3. See the stall.

### ✅ Expected Behavior

The self-referential relation applies (its observation exists). Unresolvable orphan relations are deferred (then marked `dead` after the retry cap) while the rest of the chunk imports. Import completes; a single bad relation never stalls the whole run.

### ❌ Actual Behavior

```
engram: dependency-safe local import stalled after 1 pass(es); chunk e8661c3a:
apply chunk mutation 407: relation FK precondition not met: referenced observation missing
```

Import aborts; `sync_apply_deferred` has 0 rows. Current field workaround is manually pruning the relation mutations out of the `.jsonl.gz` chunk — lossy and error-prone.

### 🖥️ Operating System

Linux (Arch/Manjaro)

### Engram Version

1.20.0

### Agent / Client

Claude Code

### 💡 Additional Context

Proposed fix: (1) require `count >= 1` when `source_id == target_id`, else `>= 2`; (2) make `ApplyPulledChunk` reuse the same `ErrRelationFKMissing` → `deferred` / `ErrApplyDead` → `dead` policy as `ApplyPulledMutation` (shared helper so the two paths cannot drift), continuing with the rest of the chunk while non-relation errors still propagate and preserve chunk atomicity. Fix + regression tests ready locally, pending `status:approved`.

pzentenoe:
## Summary
- Fix local chunk imports that stall on self-referential or permanently invalid relation mutations.
- Defer missing-endpoint relations and mark malformed relation payloads as dead while allowing the rest of the chunk to apply.
- Preserve atomic rollback for non-relation errors and keep the existing cloud import policy shared.

## Changes
- Allow a self-referential relation when its single observation exists.
- Extract relation failure recording into a transaction-scoped helper reused by `ApplyPulledMutation` and `ApplyPulledChunk`.
- Continue local chunk imports after `ErrRelationFKMissing` or `ErrApplyDead`, advance the cursor, and record the chunk as synced.
- Add regression coverage for self-relations, deferred orphan relations, dead payloads, cursor advancement, chunk registration, and existing atomic rollback.

## Validation
- [x] `go test ./... -count=1` (2368 passed)
- [x] `go test -race ./internal/store ./internal/sync ./internal/cloud/... ./internal/server/... ./internal/mcp/... -count=1` (1480 passed)
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `git diff --check`
- [x] Validated an isolated SQLite backup of the active database with `quick_check` and `integrity_check`; the active database was never modified.

## Scope
This addresses the local-import side of #649. Export-side orphan generation remains covered separately by #701 and #615.

Closes #649

## Summary by CodeRabbit

* **Bug Fixes**
  * Relation updates with missing targets are now deferred while valid updates continue processing.
  * Permanently invalid relation payloads are recorded and acknowledged without blocking other updates.
  * Self-referential relations can now be applied successfully.
  * Existing dead records are protected from being downgraded to deferred status.

* **Tests**
  * Added coverage for mixed valid and invalid relation batches, malformed payloads, self-referential relations, cursor advancement, and acknowledgment behavior.

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -4208,41 +4208,9 @@ func (s *Store) ApplyPulledMutation(targetKey string, mutation SyncMutation) err
 
 		applyErr := s.applyPulledMutationTx(tx, mutation)
 		if applyErr != nil {
-			// Phase E: per-entity skip+log policy (design §9).
-			// For relation FK misses, write to sync_apply_deferred and ACK the seq
-			// so the cursor can advance. All other errors propagate and halt the pull.
-			if mutation.Entity == SyncEntityRelation && errors.Is(applyErr, ErrRelationFKMissing) {
-				log.Printf("[store] ApplyPulledMutation: relation FK miss seq=%d entity_key=%s — deferring",
-					mutation.Seq, mutation.EntityKey)
-				if _, deferErr := s.execHook(tx, `
-					INSERT INTO sync_apply_deferred
-						(sync_id, entity, payload, apply_status, retry_count, first_seen_at)
-					VALUES (?, ?, ?, 'deferred', 0, datetime('now'))
-					ON CONFLICT(sync_id) DO UPDATE SET
-						payload            = excluded.payload,
-						last_attempted_at  = datetime('now')
-				`, mutation.EntityKey, mutation.Entity, mutation.Payload); deferErr != nil {
-					return fmt.Errorf("ApplyPulledMutation: write deferred row: %w", deferErr)
-				}
-				// Fall through to advance the cursor (ACK the seq).
-			} else if mutation.Entity == SyncEntityRelation && errors.Is(applyErr, ErrApplyDead) {
-				// Payload is permanently undecodable — write directly as dead and ACK.
-				// There is no point retrying; a malformed payload will never become valid.
-				log.Printf("[store] ApplyPulledMutation: relation payload dead seq=%d entity_key=%s err=%v — marking dead",
-					mutation.Seq, mutation.EntityKey, applyErr)
-				if _, deferErr := s.execHook(tx, `
-					INSERT INTO sync_apply_deferred
-						(sync_id, entity, payload, apply_status, retry_count, first_seen_at)
-					VALUES (?, ?, ?, 'dead', 0, datetime('now'))
-					ON CONFLICT(sync_id) DO UPDATE SET
-						payload           = excluded.payload,
-						apply_status      = 'dead',
-						last_attempted_at = datetime('now')
-				`, mutation.EntityKey, mutation.Entity, mutation.Payload); deferErr != nil {
-					return fmt.Errorf("ApplyPulledMutation: write dead row: %w", deferErr)
-				}
-				// Fall through to advance the cursor (ACK the seq).
-			} else {
+			if handled, err := s.recordRelationApplyFailureTx(tx, mutation, applyErr); err != nil {
+				return err
+			} else if !handled {
 				return applyErr
 			}
 		}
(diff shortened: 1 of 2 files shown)


pzentenoe:
Addressed the follow-up review comment in commit 5a423fe: the orphan-relation regression now also asserts that the missing relation remains absent from memory_relations. Focused test passes.\n\nCI status note: the latest GitHub Actions runs for this fork PR are completing as action_required with no jobs created, so they are awaiting maintainer approval before execution rather than failing.\n\nThe PR closes #649. Issues #615 and #701 remain separate export-side follow-ups.

pzentenoe:
Addressed CodeRabbit's latest review comment in commit 7d3c8f1: the deferred-relation regression now places the valid relation after the orphan relation, proving processing continues after the deferred failure.\n\nValidation remains green: full tests 2368, race tests 1480, vet, build, and diff checks. GitHub Actions are still  with no jobs because maintainer approval is required for fork workflows.

pzentenoe:
Correction: the follow-up commit is 4930127 (not 7d3c8f1). The deferred-relation test now places the valid relation after the orphan relation. Full tests: 2368 passed; race tests: 1480 passed; vet and build clean. GitHub Actions remain action_required with no jobs because maintainer approval is required for fork workflows.

FabianDaSilva (approved):

dnlrsls:
Thank you for the implementation and follow-up test improvements. I rebased the four commits onto current main, preserved their original authorship, and added the current-main cloud import expectation in #782. I could not update this fork branch with the available permissions, so #782 is the delivery branch for the approved issue; this PR remains the credited source of the implementation.

dnlrsls:
Thank you for the detailed relation-import implementation and extensive regression coverage. The approved solution for issue #649 was carried forward and rebased into maintainer PR #782, merged as `047a91363850decc7b442ec3d72545cd4f61fcd4`; that PR explicitly documents this lineage. Since #649 is now resolved and this PR is fully superseded, we’ll close it. The separate export-side concern was handled independently in #701/#784.

PR #766, opened 2026-08-25
fix(autosync): isolate project push failures

### LINKED ISSUE #696: bug(cloud/autosync): one failing project aborts the whole push loop, stalling cloud sync for every other project
### 📋 Pre-flight Checks

### 📝 Bug Description

`Manager.push` aborts the entire push on the first project that fails. Projects later in the batch are never attempted, so **one undeliverable payload in one project stops cloud sync for every project on the machine, indefinitely**.

`internal/cloud/autosync/manager.go:508-538`:

```go
for _, project := range order {
	...
	result, err := m.transport.PushMutations(entries)
	if err != nil {
		return fmt.Errorf("transport push project %q: %w", project, err)   // ← aborts the loop
	}
	...
}
```

The grouping immediately above deliberately partitions pending mutations by project, and the transport is called once per project — so the projects are already independent in every respect except this early `return`.

Retrying does not help: the same poisoned project is grouped first again on the next cycle, fails again, and the healthy projects behind it are never reached. The backoff then classifies the whole target as degraded.

### Why this is worth fixing separately

The project already holds the principle this needs, in its own code. `internal/store/store.go:56-62`, for the inbound relation apply path:

> `ErrRelationFKMissing` is returned by `applyRelationUpsertTx` when one or both observations referenced by the relation payload do not exist locally yet. **The caller must write the mutation to `sync_apply_deferred` and ACK the sequence so the cursor does not stall.**

That machinery is all present and shipped: the `sync_apply_deferred` table (`store.go:1071`), `apply_status='dead'` for permanently invalid payloads (`ErrApplyDead`, `store.go:70-75`), and `deferred_count` / `dead_count` surfaced through the manager (`autosync/manager.go:161-163`), the HTTP status endpoint (`server.go:913-914`) and the CLI (`conflicts.go:296-297`).

The outbound push uses none of it. A payload the cloud will never accept blocks the cursor forever — exactly the stall the inbound contract was written to prevent.

### 🔄 Steps to Reproduce

1. Have pending mutations in two or more projects.
2. Ensure one project holds a mutation the cloud rejects — a blank observation title or a session payload with no `directory` both do it.
3. Run autosync, or `engram sync --cloud`.
4. Observe that no project drains, including projects whose payloads are entirely valid.

### ✅ Expected Behavior

A project that cannot push should not prevent other projects from draining. Its failure should still be reported.

### ❌ Actual Behavior

The first failure aborts the loop. `last_sync_at` stays `null` for every project and the queue grows without bound.

### Observed impact

On one installation, three session mutations missing `directory` in a single project blocked **2,682 mutations across 12 projects for 8 days**, including 1,253 in the user's main working project, which had no invalid payloads of its own. The reported error named only the poisoned project, giving no hint that everything else was collateral.

### Engram Version

`1.20.0`

### Operating System

Linux (Ubuntu/Debian)

### Agent / Client

Claude Code

### 💡 Proposed fix

Collect per-project failures and continue, reporting them together at the end:

```go
var failures []error

for _, project := range order {
	...
	result, err := m.transport.PushMutations(entries)
	if err != nil {
		failures = append(failures, fmt.Errorf("transport push project %q: %w", project, err))
		continue
	}
	...
	if err := m.store.AckSyncMutationSeqs(m.cfg.TargetKey, seqs); err != nil {
		// An ack failure is local and not project-specific; stop here
		// rather than risk repeatedly re-pushing acked work.
		return fmt.Errorf("ack project %q: %w", project, err)
	}
}

if len(failures) > 0 {
	return errors.Join(failures...)
}
```

An ack failure still returns immediately: it is local rather than project-specific, and continuing risks re-pushing acked work.

**One dependency worth flagging, because it is a latent bug on its own.** Returning a joined error broke `TestManagerSurfacesAuthRequiredOn401` and `TestManagerSurfacesPolicyForbiddenOn403`, both degrading to `transport_failed`. The cause is `unwrapTransportStatusError` (`manager.go:465-473`), which walks the chain with `errors.Unwrap`. That cannot follow `Unwrap() []error`, so any joined error hides the status. Fixed by walking multi-error branches too:

```go
if multi, ok := err.(interface{ Unwrap() []error }); ok {
	for _, sub := range multi.Unwrap() {
		if te, ok := unwrapTransportStatusError(sub); ok {
			return te, true
		}
	}
	return nil, false
}
```

This would misclassify any joined error reaching that helper, today, independently of this change.

Verified on a patched `v1.20.0` tree:

- `go build ./...` and `go vet ./...` clean; `go test ./...` exits 0, 22 packages, no failures — including the two classification tests above.
- New test `TestPushIsolatesFailingProjectFromHealthyOnes`: a transport that rejects one project's batch with the real canonicalization 500, asserting the healthy projects are still pushed and acked while the failing project's seq is not. On unpatched code it fails with `healthy projects should still have been pushed, got []` — zero, which is the outage. Plus `TestPushReportsAllFailingProjects` asserting a healthy project never appears in the error.

### Scope note

This makes the blast radius one project instead of all of them. It does not quarantine the poisoned mutation itself — that project still cannot drain until repaired. Wiring the outbound path into the existing `sync_apply_deferred` / `apply_status='dead'` mechanism would close that too, and I would be glad to attempt it, but it is a larger change and I did not want to bundle it with a fix this small. Happy to open either or both once approved.

dnlrsls:
## Linked Issue

Closes #696

---

## PR Type

- [x] `type:bug` - Bug fix
- [ ] `type:feature` - New feature
- [ ] `type:docs` - Documentation only
- [ ] `type:refactor` - Code refactoring (no behavior change)
- [ ] `type:chore` - Maintenance, dependencies, tooling
- [ ] `type:breaking-change` - Breaking change

---

## Summary

- Continues autosync pushes after project-local transport and response failures.
- ACKs successful projects while leaving failed project mutations pending for retry.
- Preserves earlier remote failures when a later local ACK fails.
- Stops before starting another project push after context cancellation.

## Changes

| File | Change |
|------|--------|
| `internal/cloud/autosync/manager.go` | Isolate remote failures per project, preserve joined failure context, and honor cancellation between requests. |
| `internal/cloud/autosync/manager_test.go` | Add per-project failure, ACK, cancellation, backoff, status, and restart-durability coverage. |

## Test Plan

- [ ] Full unit suite completed locally: `go test ./...`
- [ ] Full E2E suite completed locally: `go test -tags e2e ./internal/server/...`
- [x] `go test ./internal/cloud/autosync -count=1`
- [x] `go vet ./internal/cloud/autosync`
- [x] Focused Store ACK-state tests pass locally.

The complete local unit and E2E commands were attempted on the initial candidate but did not finish within the 10-minute execution window. GitHub CI completed both successfully. CI will rerun for the correction commit.

---

## Automated Checks

These checks rerun automatically for the corrected head:

| Check | Status |
|-------|--------|
| Check Issue Reference | Pending |
| Check Issue Has status:approved | Pending |
| Check PR Has type:* Label | Pending |
| Unit Tests | Pending |
| E2E Tests | Pending |

---

## Contributor Checklist

- [x] I linked an approved issue above (`Closes #696`)
- [x] I ran unit tests locally
- [x] I ran E2E tests locally
- [x] Docs updated (not required; no public contract changed)

---

## Notes for Reviewers

CodeRabbit feedback led to a bounded follow-up that preserves accumulated remote failures on local ACK errors and honors cancellation between project requests.

Native RDD reviewed and approved both the initial candidate and the exact corrected commit tree. Delivery remains governed by ordinary repository policy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Autosync now continues processing other projects when one project’s synchronization fails.
  * Reports all project-level synchronization failures together for clearer troubleshooting.
  * Preserves immediate handling of local acknowledgement failures.
  * Prevents incomplete synchronization cycles from incorrectly proceeding to subsequent steps.
  * Improved error classification and consistency across restarts.

### DIFF
--- a/internal/cloud/autosync/manager.go
+++ b/internal/cloud/autosync/manager.go
@@ -463,13 +463,21 @@ func autosyncFailureMessage(targetKey, message string, err error) string {
 
 // unwrapTransportStatusError walks the error chain looking for transportStatusError.
 func unwrapTransportStatusError(err error) (transportStatusError, bool) {
-	for err != nil {
-		if te, ok := err.(transportStatusError); ok {
-			return te, true
-		}
-		err = errors.Unwrap(err)
+	if err == nil {
+		return nil, false
 	}
-	return nil, false
+	if te, ok := err.(transportStatusError); ok {
+		return te, true
+	}
+	if multi, ok := err.(interface{ Unwrap() []error }); ok {
+		for _, wrapped := range multi.Unwrap() {
+			if te, ok := unwrapTransportStatusError(wrapped); ok {
+				return te, true
+			}
+		}
+		return nil, false
+	}
+	return unwrapTransportStatusError(errors.Unwrap(err))
 }
 
 // ─── Push ────────────────────────────────────────────────────────────────────
@@ -507,6 +515,7 @@ func (m *Manager) push(ctx context.Context) error {
 		groups[project] = append(groups[project], mut)
 	}
 
+	var failures []error
 	for _, project := range order {
 		batch := groups[project]
 		entries := make([]MutationEntry, len(batch))
@@ -524,20 +533,23 @@ func (m *Manager) push(ctx context.Context) error {
 
 		result, err := m.transport.PushMutations(entries)
 		if err != nil {
-			return fmt.Errorf("transport push project %q: %w", project, err)
+			failures = append(failures, fmt.Errorf("transport push project %q: %w", project, err))
+			continue
 		}
 		if result == nil {
-			return fmt.Errorf("transport push project %q: missing accepted seqs for %d mutations", project, len(entries))
+			failures = append(failures, fmt.Errorf("transport push project %q: missing accepted seqs for %d mutations", project, len(entries)))
+			continue
 		}
 		if len(result.AcceptedSeqs) != len(entries) {
-			return fmt.Errorf("transport push project %q: cloud accepted %d of %d mutations; refusing to ack local seqs", project, len(result.AcceptedSeqs), len(entries))
+			failures = append(failures, fmt.Errorf("transport push project %q: cloud accepted %d of %d mutations; refusing to ack local seqs", project, len(result.AcceptedSeqs), len(entries)))
+			continue
 		}
 		if err := m.store.AckSyncMutationSeqs(m.cfg.TargetKey, seqs); err != nil {
 			return fmt.Errorf("ack project %q: %w", project, err)
 		}
 	}
 
-	return nil
+	return errors.Join(failures...)
 }
 
 // ─── Pull ────────────────────────────────────────────────────────────────────
--- a/internal/cloud/autosync/manager.go
+++ b/internal/cloud/autosync/manager.go
@@ -517,6 +517,10 @@ func (m *Manager) push(ctx context.Context) error {
 
 	var failures []error
 	for _, project := range order {
+		if err := ctx.Err(); err != nil {
+			failures = append(failures, err)
+			return errors.Join(failures...)
+		}
 		batch := groups[project]
 		entries := make([]MutationEntry, len(batch))
 		seqs := make([]int64, len(batch))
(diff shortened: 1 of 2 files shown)


Alan-TheGentleman (approved):
Verified in depth against current main at head 0777336e: per-project failure isolation matches the approved #696 proposal, the joined-error unwrap fix restores 401/403 classification, ack failures preserve prior remote failures, and the test suite proves isolation for all three failure shapes plus real-store restart durability. Every new test fails on unpatched main. Merging. Follow-up worth filing: a permanently poisoned project still drives the manager into terminal PhaseBackoff after MaxConsecutiveFailures, so healthy projects stall again until repair; that is the outbound quarantine #696 deferred.

PR #787, opened 2026-08-26
fix(store): detect replaced SQLite generations

### LINKED ISSUE #477: bug(store): long-running processes keep stale WAL-index (-shm) mmap after file replacement — silent write loss + constraint failed (1555) on mem_session_summary

dnlrsls:
## Linked Issue

Closes #477
Refs #571

---

## PR Type

- [x] `type:bug` - Bug fix
- [ ] `type:feature` - New feature
- [ ] `type:docs` - Documentation only
- [ ] `type:refactor` - Code refactoring (no behavior change)
- [ ] `type:chore` - Maintenance, dependencies, tooling
- [ ] `type:breaking-change` - Breaking change

---

## Summary

- Detect replacement or disappearance of the live SQLite database, WAL, or shared-memory generation and fail loudly with restart guidance.
- Coordinate Store lifetime and orphan migration with canonical cross-platform leases and rollback-safe generation moves.
- Throttle identity checks and guard connections, transactions, statements, and row iteration.
- Complements the lifecycle gaps not covered by #613; this PR does not close or modify #613.

## Changes

| File | Change |
|------|--------|
| `internal/store/generation_guard.go` | Guards connections, transactions, statements, and row iteration with throttled database/WAL/SHM identity checks. |
| `internal/store/generation_identity_{unix,windows}.go` | Provides platform-specific file identity implementations. |
| `internal/store/store_lease*.go` | Adds in-directory Store and migration leases plus rollback-safe database generation moves. |
| `internal/store/store.go` | Acquires/releases lifecycle resources and captures the stable generation after initialization. |
| `cmd/engram/main.go` | Routes orphan migration through the coordinated Store move operation. |
| `internal/store/*_test.go` | Covers replacement, disappearance, throttling, transaction/row boundaries, aliases, leases, and rollback. |
| `go.mod` | Promotes `golang.org/x/sys` for Windows identity and lease APIs. |
| `DOCS.md` | Documents detector/TOCTOU limits and sticky restart behavior. |

This is one cohesive cross-platform lifecycle work unit; the guard, leases, migration API, platform backends, and regression tests must land together to preserve compilation and behavior.

## Test Plan

- [x] Focused Store tests pass locally: `go test ./internal/store -count=1`
- [x] Linux/amd64 and Darwin/arm64 Store tests cross-compile successfully.
- [ ] Race detector: unavailable on this Windows host because CGO is disabled.
- [ ] Manual runtime testing was not performed.

Linux and Darwin checks prove compilation only; those test binaries were not executed on their target operating systems.

---

## Automated Checks

These run automatically and all must pass before merge:

| Check Issue Reference | PR body contains `Closes #N`, `Fixes #N`, or `Resolves #N` | Pending |
| Check Issue Has status:approved | Linked issue has `status:approved` | Pending |
| Check PR Has type label | PR has exactly one `type:*` label | Pending |
| Unit Tests | `go test ./...` passes | Pending |
| E2E Tests | `go test -tags e2e ./internal/server/...` passes | Pending |

---

## Contributor Checklist

- [x] Linked approved issue #477 and referenced the remaining #571 health/reopen work.
- [x] Added exactly one `type:*` label.
- [x] Ran unit tests locally: `go test ./...`.
- [x] Ran E2E tests locally: `go test -tags e2e ./internal/server/...`.
- [x] Updated docs for detector and restart behavior.
- [x] Commits follow Conventional Commits.
- [x] No `Co-Authored-By` trailers.

---

## Notes for Reviewers

The Store intentionally fails loudly after detecting a generation change rather than reopening a potentially stale SQLite connection. Restarting Engram opens the new generation safely.

#571 remains referenced rather than closed because transparent reopen and a database-aware health signal are outside this PR.

## Summary by CodeRabbit

* **New Features**
  * Added safeguards that detect unexpected database, WAL, or shared-memory file changes and stop affected operations safely.
  * Added coordinated access protection to prevent conflicting store usage across processes.
  * Added cross-platform support for database locking and file-change detection.
* **Bug Fixes**
  * Database and statistics errors are now surfaced instead of being silently ignored.
  * Updated tools and endpoints to report clear errors when the database requires a restart.
* **Documentation**
  * Documented database safety safeguards and restart requirements.
* **Tests**
  * Expanded coverage for file replacement, locking conflicts, symlinked paths, and error handling.


dnlrsls:
@coderabbitai review

dnlrsls:
@coderabbitai review

dnlrsls:
@coderabbitai review

Alan-TheGentleman (changes requested):
Requesting changes at head `b1c08b24`. The guard mechanism itself is careful and genuinely catches the #477 failure mode (sticky error, late-sidecar adoption, correct per-platform identity semantics, advisory locks that die with the process), but the lease placement introduces two real regressions.

The lease file lands in the parent of the data dir (`store_lease.go:56`), and `newStore` hard-fails if it cannot be created. That breaks any deployment where the parent of `ENGRAM_DATA_DIR` is read-only (Docker `--read-only`, K8s `readOnlyRootFilesystem`), which could open the store fine on main. Worse, orphan candidates live at `/.engram` and `C:\.engram`, so the move lease lands at the drive root, where non-root Unix users and default Windows ACLs cannot create files. The orphan migration now always fails for exactly the users it was written for, where main's plain rename succeeded. Placing the lock inside the data dir fixes both, since `MoveDatabaseGeneration` moves the three db files, not the directory.

Also: `TestStoreLeasePathAllowsMissingDataDirectory` fails on macOS (symlink canonicalization mismatch, CI is ubuntu-only so it stays green), and the measured per-query cost is 3.5x on point reads (~9 stat syscalls per query), so the open CodeRabbit throttle suggestion is a real ask before merge, not a nit. Two design notes worth a doc sentence: a driver-discarded last connection can unlink WAL/SHM and permanently poison the store until restart, and the check-then-use window means one silently-lost write is still possible. Finally, #571's `/health` probe is not addressed, so consider whether `Closes #571` is accurate.

danielgap:
Went through the requested changes and prepared a patch that covers the four blockers. Nothing pushed here — sharing the diff in case it's useful directly or as a reference.

- **Lease location**: `storeLeasePath` now keeps `.engram.store.lock` inside the canonical data directory, so acquisition no longer requires a writable parent (read-only container roots, orphan databases under `/`). New `TestAcquireStoreLeaseWithReadOnlyParent` covers that case, and the move test now expects the lease file to remain in the source directory.
- **macOS/symlinks**: the lease path is derived from the symlink-resolved canonical data directory, so `/tmp` vs `/private/tmp` and symlinked data dirs resolve to a single identity; the path tests now assert against the canonicalized directory.
- **Guard cost**: full DB/WAL/SHM identity scans are throttled to one per 10 ms, with an injectable clock so tests drive the throttle without sleeps. The sticky `ErrDatabaseGenerationChanged` state performs zero additional scans, and a dedicated regression test disables the throttle to prove the replacement path itself still detects.
- **TOCTOU limits**: the guard doc now states explicitly that it is a detector, not a fence — replacement can land between checks or between check and driver use — and documents the driver-discarded-connection case where WAL/SHM unlinking makes the failure sticky and requires a restart.

Validated with `go test -race ./internal/store`, full `go test ./...`, server E2E (`go test -tags e2e ./internal/server/...`), `go vet ./internal/store`, and cross-compiles for `windows/amd64` and `darwin/arm64`.

On the #571 link: detection + fail-loud is the core of what that issue asks, but the DB-aware health signal is still open. I'd suggest `Refs #571` plus a follow-up issue for health, or keep `Closes #571` if you consider this slice sufficient to close it.

<details>
<summary>Patch (6 files, +167/−13)</summary>

```diff
diff --git a/internal/store/generation_guard.go b/internal/store/generation_guard.go
index 2223711..cc570c5 100644
--- a/internal/store/generation_guard.go
+++ b/internal/store/generation_guard.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"sync"
+	"time"
 
 	sqlite "modernc.org/sqlite"
 )
@@ -24,21 +25,48 @@ type generationFile struct {
 	identity databaseFileIdentity
 }
 
+const defaultDatabaseGenerationCheckInterval = 10 * time.Millisecond
+
+// databaseGeneration detects unsupported replacement of a live SQLite file
+// generation. It is a detector, not an atomic transaction fence: replacement
+// can still occur in the bounded interval between checks or between check and
+// driver use, allowing operations in that window to reach the stale generation
+// before a later check poisons the Store. Checks are therefore frequent but
+// throttled to avoid repeating the
+// same DB/WAL/SHM identity syscalls at every database/sql wrapper layer.
+//
+// A driver-discarded last connection can also unlink WAL/SHM. When that occurs,
+// the next full check makes the failure sticky and requires an Engram restart;
+// the guard deliberately never attempts an unsafe transparent reopen.
+//
+// The clock and interval are fields so tests can drive the throttle without
+// sleeps. Production instances always use time.Now and the default interval.
+// All mutable state remains protected by mu.
+//
+// TODO: PR #613's persistent-WAL connection setup reduces the last-connection
+// unlink risk when that independent prevention slice lands.
 type databaseGeneration struct {
-	mu      sync.Mutex
-	files   []generationFile
-	enabled bool
-	changed bool
+	mu               sync.Mutex
+	files            []generationFile
+	enabled          bool
+	changed          bool
+	now              func() time.Time
+	minCheckInterval time.Duration
+	lastFullCheck    time.Time
 }
 
 var readDatabaseFileID = databaseFileID
 
 func newDatabaseGeneration(dbPath string) *databaseGeneration {
-	return &databaseGeneration{files: []generationFile{
-		{name: "database", path: dbPath, required: true},
-		{name: "WAL", path: dbPath + "-wal"},
-		{name: "shared-memory", path: dbPath + "-shm"},
-	}}
+	return &databaseGeneration{
+		files: []generationFile{
+			{name: "database", path: dbPath, required: true},
+			{name: "WAL", path: dbPath + "-wal"},
+			{name: "shared-memory", path: dbPath + "-shm"},
+		},
+		now:              time.Now,
+		minCheckInterval: defaultDatabaseGenerationCheckInterval,
+	}
 }
 
 func (g *databaseGeneration) capture() error {
@@ -80,6 +108,14 @@ func (g *databaseGeneration) check() error {
 		return ErrDatabaseGenerationChanged
 	}
 
+	now := g.now()
+	if g.minCheckInterval > 0 &&
+		!g.lastFullCheck.IsZero() &&
+		!now.Before(g.lastFullCheck) &&
+		now.Sub(g.lastFullCheck) < g.minCheckInterval {
+		return nil
+	}
+
 	for i := range g.files {
 		file := &g.files[i]
 		identity, err := readDatabaseFileID(file.path)
@@ -101,6 +137,7 @@ func (g *databaseGeneration) check() error {
 			return g.changedError(file.name, "identity changed")
 		}
 	}
+	g.lastFullCheck = now
 	return nil
 }
 
diff --git a/internal/store/generation_guard_test.go b/internal/store/generation_guard_test.go
index 6ed2c5a..6a07741 100644
--- a/internal/store/generation_guard_test.go
+++ b/internal/store/generation_guard_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 )
 
 func TestDatabaseGenerationDetectsReplacedSidecar(t *testing.T) {
@@ -44,6 +45,7 @@ func TestDatabaseGenerationAllowsLateSidecars(t *testing.T) {
 		t.Fatalf("write database: %v", err)
 	}
 	generation := newDatabaseGeneration(dbPath)
+	generation.minCheckInterval = 0
 	if err := generation.capture(); err != nil {
 		t.Fatalf("capture without sidecars: %v", err)
 	}
@@ -138,3 +140,61 @@ func TestDatabaseGenerationPreservesIdentityReadErrors(t *testing.T) {
 		t.Fatalf("check error = %v, must not be a generation replacement", err)
 	}
 }
+
+func TestDatabaseGenerationThrottlesIdentityScans(t *testing.T) {
+	dir := t.TempDir()
+	dbPath := filepath.Join(dir, "engram.db")
+	if err := os.WriteFile(dbPath, []byte("database"), 0o600); err != nil {
+		t.Fatalf("write database: %v", err)
+	}
+
+	generation := newDatabaseGeneration(dbPath)
+	now := time.Unix(1, 0)
+	generation.now = func() time.Time { return now }
+	generation.minCheckInterval = 10 * time.Millisecond
+	if err := generation.capture(); err != nil {
+		t.Fatalf("capture generation: %v", err)
+	}
+
+	original := readDatabaseFileID
+	reads := 0
+	t.Cleanup(func() { readDatabaseFileID = original })
+	readDatabaseFileID = func(path string) (databaseFileIdentity, error) {
+		reads++
+		return original(path)
+	}
+
+	if err := generation.check(); err != nil {
+		t.Fatalf("initial check: %v", err)
+	}
+	if reads != len(generation.files) {
+		t.Fatalf("initial identity reads = %d, want %d", reads, len(generation.files))
+	}
+
+	replacement := dbPath + ".replacement"
+	if err := os.WriteFile(replacement, []byte("replacement"), 0o600); err != nil {
+		t.Fatalf("write replacement database: %v", err)
+	}
+	if err := os.Rename(replacement, dbPath); err != nil {
+		t.Fatalf("replace database: %v", err)
+	}
+
+	if err := generation.check(); err != nil {
+		t.Fatalf("check inside throttle interval: %v", err)
+	}
+	if reads != len(generation.files) {
+		t.Fatalf("throttled identity reads = %d, want %d", reads, len(generation.files))
+	}
+
+	now = now.Add(generation.minCheckInterval)
+	if err := generation.check(); !errors.Is(err, ErrDatabaseGenerationChanged) {
+		t.Fatalf("check after throttle interval = %v, want ErrDatabaseGenerationChanged", err)
+	}
+	readsAfterChange := reads
+	if err := generation.check(); !errors.Is(err, ErrDatabaseGenerationChanged) {
+		t.Fatalf("sticky generation check = %v, want ErrDatabaseGenerationChanged", err)
+	}
+	if reads != readsAfterChange {
+		t.Fatalf("sticky generation check performed %d extra identity reads", reads-readsAfterChange)
+	}
+}
diff --git a/internal/store/generation_guard_unix_test.go b/internal/store/generation_guard_unix_test.go
index 98bafba..b3238ae 100644
--- a/internal/store/generation_guard_unix_test.go
+++ b/internal/store/generation_guard_unix_test.go
@@ -22,6 +22,9 @@ func TestStoreGenerationGuardRejectsLiveSidecarReplacement(t *testing.T) {
 	if err := s.CreateSession("before-replacement", "engram", "/work/engram"); err != nil {
 		t.Fatalf("create session before replacement: %v", err)
 	}
+	// This regression proves the replacement path itself, independent of the
+	// production hot-path throttle covered by TestDatabaseGenerationThrottlesIdentityScans.
+	s.generation.minCheckInterval = 0
 
 	shmPath := filepath.Join(dir, "engram.db-shm")
 	if _, err := os.Stat(shmPath); err != nil {
diff --git a/internal/store/store_lease.go b/internal/store/store_lease.go
index 6281634..8dc86d0 100644
--- a/internal/store/store_lease.go
+++ b/internal/store/store_lease.go
@@ -11,6 +11,8 @@ import (
 
 var ErrDatabaseStoreInUse = errors.New("database store is in use")
 
+const storeLeaseFilename = ".engram.store.lock"
+
 var (
 	storeLeaseTimeout       = 60 * time.Second
 	storeLeaseRetryInterval = 25 * time.Millisecond
@@ -48,12 +50,18 @@ func acquireStoreLease(dataDir string, exclusive bool) (*storeLease, error) {
 	}
 }
 
+// storeLeasePath keeps the lease inside the Store data directory. Store data
+// directories must already be writable for SQLite, while their parents may be
+// read-only (for example read-only container roots or orphan databases under a
+// filesystem root). MoveDatabaseGeneration moves only the database generation,
+// not the directory, so the stable lease inode remains available for aliases
+// and concurrent open/move exclusion.
 func storeLeasePath(dataDir string) (string, error) {
 	canonicalDir, err := canonicalStoreLeaseDir(dataDir)
 	if err != nil {
 		return "", err
 	}
-	return filepath.Join(filepath.Dir(canonicalDir), "."+filepath.Base(canonicalDir)+".engram.store.lock"), nil
+	return filepath.Join(canonicalDir, storeLeaseFilename), nil
 }
 
 func canonicalStoreLeaseDir(dataDir string) (string, error) {
diff --git a/internal/store/store_lease_test.go b/internal/store/store_lease_test.go
index 712cf36..0e075e9 100644
--- a/internal/store/store_lease_test.go
+++ b/internal/store/store_lease_test.go
@@ -49,8 +49,8 @@ func TestMoveDatabaseGenerationRefusesLiveStore(t *testing.T) {
 	if err != nil {
 		t.Fatalf("read source directory after move: %v", err)
 	}
-	if len(entries) != 0 {
-		t.Fatalf("source directory retained entries after move: %v", entries)
+	if len(entries) != 1 || entries[0].Name() != storeLeaseFilename {
+		t.Fatalf("source directory entries after move = %v, want only %q", entries, storeLeaseFilename)
 	}
 }
 
@@ -167,15 +167,35 @@ func TestMoveDatabaseGenerationPreservesExistingDestination(t *testing.T) {
 	}
 }
 
+func TestStoreLeasePathLivesInsideDataDirectory(t *testing.T) {
+	dataDir := t.TempDir()
+	canonicalDir, err := filepath.EvalSymlinks(dataDir)
+	if err != nil {
+		t.Fatalf("canonicalize data directory: %v", err)
+	}
+
+	lockPath, err := storeLeasePath(dataDir)
+	if err != nil {
+		t.Fatalf("derive lease path: %v", err)
+	}
+	if want := filepath.Join(canonicalDir, storeLeaseFilename); lockPath != want {
+		t.Fatalf("lease path = %q, want %q", lockPath, want)
+	}
+}
+
 func TestStoreLeasePathAllowsMissingDataDirectory(t *testing.T) {
 	parent := t.TempDir()
+	canonicalParent, err := filepath.EvalSymlinks(parent)
+	if err != nil {
+		t.Fatalf("canonicalize parent directory: %v", err)
+	}
 	dataDir := filepath.Join(parent, "missing")
 
 	lockPath, err := storeLeasePath(dataDir)
 	if err != nil {
 		t.Fatalf("derive lease path: %v", err)
 	}
-	if want := filepath.Join(parent, ".missing.engram.store.lock"); lockPath != want {
+	if want := filepath.Join(canonicalParent, "missing", storeLeaseFilename); lockPath != want {
 		t.Fatalf("lease path = %q, want %q", lockPath, want)
 	}
 	if _, err := os.Stat(dataDir); !errors.Is(err, os.ErrNotExist) {
diff --git a/internal/store/store_lease_unix_test.go b/internal/store/store_lease_unix_test.go
index 0608dc5..cf73da4 100644
--- a/internal/store/store_lease_unix_test.go
+++ b/internal/store/store_lease_unix_test.go
@@ -9,6 +9,32 @@ import (
 	"testing"
 )
 
+func TestAcquireStoreLeaseWithReadOnlyParent(t *testing.T) {
+	if os.Geteuid() == 0 {
+		t.Skip("root bypasses parent directory permissions")
+	}
+	parent := t.TempDir()
+	dataDir := filepath.Join(parent, "store")
+	if err := os.Mkdir(dataDir, 0o755); err != nil {
+		t.Fatalf("create data directory: %v", err)
+	}
+	if err := os.Chmod(parent, 0o555); err != nil {
+		t.Fatalf("make parent read-only: %v", err)
+	}
+	t.Cleanup(func() { _ = os.Chmod(parent, 0o755) })
+
+	lease, err := acquireStoreLease(dataDir, false)
+	if err != nil {
+		t.Fatalf("acquire lease with read-only parent: %v", err)
+	}
+	if err := lease.Close(); err != nil {
+		t.Fatalf("close lease: %v", err)
+	}
+	if _, err := os.Stat(filepath.Join(dataDir, storeLeaseFilename)); err != nil {
+		t.Fatalf("lease file inside data directory: %v", err)
+	}
+}
+
 func TestMoveDatabaseGenerationRefusesLiveStoreThroughSymlink(t *testing.T) {
 	useShortStoreLeaseTimeout(t)
 	base := t.TempDir()
```
</details>

danielgap:
One complement to the identity-guard axis, offered from the work in #887 (now closed as superseded by this PR): a post-commit visibility gate for critical writes.

The gap it closes is the one called out in review: the guard is a detector, not a fence, so a replacement landing between an identity check and the driver using the connection still allows exactly one silently-lost write. Inversely, after a critical write commits on the live single connection, a fresh independent read-only connection opening the current generation will always see that committed row — WAL readers observe committed data. So verifying the just-written row through a fresh connection after commit (AddObservation, CreateSession) converts the silent-loss window into a loud failure: an invisible row, a missing schema, or a fresh-reader I/O error after a successful live commit all mean the live connection is operating on a divergent generation, and can set the same sticky restart state as the identity guard.

A deterministic repro harness for that exact scenario, no multi-day uptime needed: seed a store, rename the `-wal` file aside mid-session, then write again. Measured behavior on current main: the live connection still acknowledges the write (it lands in the old inode's WAL), while a fresh open fails with `disk I/O error (522)` from the orphaned `-shm` — the acknowledgement is the silent write loss, and the fresh reader's failure is the detection signal. Both the gate and this harness are implemented and tested in #887's branch (`internal/store/generation.go`, `internal/store/generation_test.go`) if any of it is useful to fold in.

One incidental for the sqlite error classification if the sticky path surfaces driver errors: `modernc.org/sqlite`'s `Error.Code()` returns the extended code (522), not the primary (10), so primary-code matching needs a `& 0xff` mask.

dnlrsls:
Superseded by #902, which restarts from current main with a bounded driver-level generation fence and removes the migration, lease, health, and recovery scope from this proposal. Preserving this PR as implementation history.

PR #798, opened 2026-08-26
fix(store): prevent empty session identity replay

### LINKED ISSUE #630: fix(cloud): empty source sessions.id is re-emitted as a broken mutation; repair only fixes the journal
### 📋 Pre-flight Checks

### 📝 Bug Description
Split from #340 (blocker class D). Repairing the `sync_mutations` payload is not enough when the source `sessions` row has `id=''`: the daemon regenerates a fresh broken mutation from the source row minutes later. `doctor`/`repair` never scan or fix `sessions.id` in the source table. Reported by @iamdavidbravo on #340.

### 🔄 Steps to Reproduce
1. Have a live `sessions` row with `id=''` (observed as 35+ rows created in a ~1-day window under parallel git worktrees on the same project).
2. `doctor` reports `session payload id is required`; fix the payload manually.
3. Minutes later the daemon re-emits a new empty-id mutation for the same project.
4. `SELECT id, project, directory, started_at, ended_at FROM sessions WHERE id IS NULL OR TRIM(id)='';` returns a live row.

### ✅ Expected Behavior
`doctor` detects empty `sessions.id` in the source table; `repair` fixes the source row (cascading to `observations.session_id` / `user_prompts.session_id`) or reports it unfixable with full row context. Dry-run by default, backup before apply.

### ❌ Actual Behavior
Empty id is only detected inside a pending mutation payload, never in the source table, so the daemon keeps re-emitting broken mutations. Server enforcement (required even for delete ops): `internal/cloud/cloudserver/cloudserver.go:678`, `:726` and `internal/cloud/chunkcodec/chunkcodec.go:362`.

### Operating System
Linux (Other)

### Engram Version
1.19.0 (self-hosted, client upgraded 1.16.1 → 1.19.0 during investigation)

### Agent / Client
Other

### 📋 Relevant Logs
```
status 400: invalid push payload: session payload id is required
```

### 💡 Additional Context
Child of #340. Harder than the directory backfill: an empty id cannot be matched by any `WHERE sessions.id=?` join, and reassigning an id must cascade. Suspected root cause (unconfirmed): a race generating the session id when multiple agents start against the same project from different worktree directories.

dnlrsls:
## 🔗 Linked Issue

Closes #630

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix

---

## 📝 Summary

- Reject blank session identities at Store boundaries while preserving opaque and legacy IDs.
- Prevent corrupt source sessions from emitting broken mutations without starving valid-session backfill.
- Add doctor evidence for invalid source, references, and journal entries without inventing replacement IDs.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | Enforce session identity invariants, canonical mutation payloads, safe backfill, and legacy pull compatibility. |
| `internal/store/diagnostic.go` | Report invalid source identities, dependent references, and journal inconsistencies. |
| `internal/diagnostic/*` | Register the new blocking check and explicit non-repairable plan. |
| `cmd/engram/doctor*`, `internal/mcp/mcp_test.go`, `internal/store/store_test.go` | Cover CLI, MCP, Store, import, pull, backfill, and startup behavior. |

## 🧪 Test Plan

- [x] Focused Store, MCP, diagnostic, and doctor regression tests pass.
- [ ] Full unit suite PASS locally: `go test ./...` timed out after 10 minutes without a result; CI remains required.
- [x] Native Gentle AI RDD approved the corrected immutable candidate after one bounded correction.

---

## 🤖 Automated Checks

| **Check Issue Reference** | PR body contains `Closes #630` | ⏳ |
| **Check Issue Has status:approved** | Linked issue has `status:approved` | ⏳ |
| **E2E Tests** | Server E2E suite passes | ⏳ |

---

## ✅ Contributor Checklist

- [x] I linked an approved issue above (`Closes #630`)
- [x] I ran unit tests locally; the full suite timed out and is disclosed above
- [x] I ran e2e tests locally
- [x] Docs reviewed; no documentation update is required for this internal validation behavior

---

## 💬 Notes for Reviewers

- The approved size exception covered the 605-line frozen candidate; RDD then admitted one bounded correction within its 200-line budget.
- Historical empty IDs remain unchanged unless a future workflow provides an explicit canonical replacement ID. Doctor returns `cannot_repair_without_explicit_canonical_session_id` instead of mutating identity silently.
- RDD approval is informational and was burned after targeted validation; ordinary repository policy governs this PR.

## Summary by CodeRabbit

- **New Features**
  - Added diagnostics for sessions with missing or invalid identities, including related records and evidence.
  - Added validation to ensure session identifiers are present and match across synchronization and import operations.
- **Bug Fixes**
  - Prevented blank or whitespace-only session IDs from being created, imported, or persisted.
  - Improved handling of inconsistent session mutations, invalid backfills, and legacy database upgrades.
  - Preserved valid opaque identifiers exactly as entered.
- **Repair**
  - Doctor repair checks now recognize invalid session identities and safely report when automatic repair is unavailable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ Breaking changes are always marked with a `type:breaking-change` label and docum
 
 <!-- Changes that are merged but not yet released are tracked here until the next tag. -->
 
+### Memory core
+
+- **fix(store):** reject empty or whitespace-only observation titles consistently on create and update, before any side effect. `engram save` and `mem_save` now refuse a titleless write before opening the store or creating a session, `POST /observations` validates the title before the session lookup so a bad session or project can no longer mask the documented `400`, and `PATCH /observations/{id}` answers `400` rather than `404`. Persisting a titleless observation also enqueues a cloud upsert that the sync validators reject, which blocks every later mutation for the project.
+
 ### Cloud sync
 
 - **fix(cloud):** make chunk and mutation push payload limits configurable with `ENGRAM_CLOUD_MAX_PUSH_BYTES` while preserving the 8 MiB default.
--- a/DOCS.md
+++ b/DOCS.md
@@ -134,10 +134,12 @@ Engram is local-first: local SQLite is authoritative; cloud features are optiona
 ### Observations
 
 - `POST /observations` — Add observation. Body: `{session_id, type, title, content, tool_name?, project?, scope?, topic_key?}`
+  - `400` when `title` or `content` is missing, empty, or whitespace-only. The observation-create paths (`engram save`, `mem_save`, `POST /observations`) enforce the same title rule because cloud sync rejects observation upserts without a title, and one rejected mutation blocks every later mutation for the project
 - `GET /observations` — Recent observations compatibility endpoint. Query: `?project=X&scope=project|personal|global&limit=N&sort=created_at:desc`
 - `GET /observations/recent` — Recent observations. Query: `?project=X&scope=project|personal|global&limit=N`
 - `GET /observations/{id}` — Get single observation by ID
 - `PATCH /observations/{id}` — Update fields. Body: `{title?, content?, type?, project?, scope?, topic_key?}`
+  - `400` when `title` or `content` is provided but empty or whitespace-only. Omitting a field leaves its current value unchanged
 - `DELETE /observations/{id}` — Delete observation (`?hard=true` for hard delete, soft delete by default)
   - `200` when deleted
   - `404` when observation does not exist
--- a/DOCS.md
+++ b/DOCS.md
@@ -171,7 +171,8 @@ Engram is local-first: local SQLite is authoritative; cloud features are optiona
 
 ### Context
 
-- `GET /context` — Formatted context. Query: `?project=X&scope=project|personal|global`
+- `GET /context` — Manual formatted context scoped by project and optional scope. Query: `?project=X&scope=project|personal|global`
+- `GET /context/compaction` — Runtime compaction context scoped strictly to one persisted session. Query: `?session_id=X`. The server derives the session project; this endpoint does not accept project or scope selection.
 
 ### Passive Capture
 
@@ -549,6 +550,7 @@ Inspect or replay the `sync_apply_deferred` queue.
 - `engram cloud upgrade rollback --project <project>` — restore pre-upgrade local snapshot before `bootstrap_verified`; blocked afterwards
 - `engram cloud repair materialize-mutations --project <project> (--dry-run|--apply)` — explicit server-side Postgres repair that backfills existing `cloud_mutations` into compatible `cloud_chunks` without deleting remote data
 - `engram cloud bootstrap admin --username <name> [--email <email>] [--grant-project <project>]... [--issue-token [name]]` — create the first managed admin (see [Managed users, tokens, and CLI bootstrap](#managed-users-tokens-and-cli-bootstrap))
+- `engram cloud bootstrap recover-token [--name <name>]` — recover the one stranded managed admin token state described below
 
 Cloud auth token is provided at runtime via `ENGRAM_CLOUD_TOKEN` (not by a dedicated CLI subcommand).
 Cloud server startup fails closed when the token is missing unless `ENGRAM_CLOUD_INSECURE_NO_AUTH=1` is explicitly set for local insecure development.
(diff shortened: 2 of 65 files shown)


dnlrsls:
@Alan-TheGentleman, could you please rerun the failed Unit Tests job? It failed in the unrelated `internal/cloud/autosync.TestManagerPanicSetsBackoff` test. PR #798 does not modify that package, and the isolated test passes locally. Failed run: https://github.com/Gentleman-Programming/engram/actions/runs/32942686269

Alan-TheGentleman (changes requested):
Requesting changes at head `36df1171`. The core is good: blank means trimmed-empty only, opaque and legacy IDs are preserved byte-exact, doctor is not bricked (opens on corrupt enrolled stores, reports blocked, explicit noop plan), and 9 of the 10 new store tests genuinely fail on main. But the PR's own claim, rejecting blank session identities at Store boundaries, has two holes, one proven empirically on this head.

`EndSession` and `DeleteSession` are unguarded: on a store containing the exact corrupt `sessions.id = ''` row from #630, `EndSession("")` matches it and enqueues a session upsert with a blank ID (a probe test on this head fails with the re-emitted mutation), and `mem_session_end` passes its argument straight through, so it is a live trigger. Centralizing the guard in `enqueueSyncMutationTx` for session entities would close the class instead of the instances.

Second, historical pulled chunks with blank identities go from tolerated to poison pills: `applyPulledMutationTx` now errors, the pull halts, and the cursor never advances, a permanent retry loop with no skip, defer, or dead-letter, hitting exactly the pre-enforcement self-hosted servers where #630's reporter lives. If fail-closed is the intent here, say so explicitly; skip-plus-evidence looks safer. Smaller items: `projectNeedsBackfill` still counts blank rows so affected stores run a no-op write transaction on every open, and removing the entity-key trim quietly loosened the observation/prompt diagnostics for whitespace-only keys.

dnlrsls:
@Alan-TheGentleman, addressed the requested changes in `b3a3ff1`: `EndSession`, `DeleteSession`, and the central session enqueue boundary now reject blank IDs; historical pulled blank session identities are skipped with durable dead-letter evidence while the cursor advances atomically; chunk pulls continue past only typed identity failures; backfill ignores blank rows; and observation/prompt whitespace keys are diagnosed consistently. Added Store, migration, and MCP regressions. Focused tests and `go test ./... -run '^$'` pass. Could you please re-review this head?

Alan-TheGentleman (changes requested):
Requesting changes. Go `TrimSpace` rejects tabs and newlines, but SQLite `trim(id)` only strips spaces, so a corrupt legacy ID can bypass Doctor and later roll back a valid backfill. Use one consistent validity rule across Go and storage. Add source-row tests for tab- and newline-containing IDs before merging.

Alan-TheGentleman (approved):
Resolved. Both of my earlier reviews are addressed, and one correction to the record.

The centralized guard is in `enqueueSyncMutationTx` for session entities, so `EndSession`, `DeleteSession` and any future caller are covered as a class rather than instance by instance. The pull is no longer a poison pill: skip-plus-evidence is the deliberate semantics, documented, with an undecodable payload staying fail-closed since that is a transport fault rather than known-corrupt history. Good call extending doctor to report the skipped evidence — leaving it green on a store that had silently dropped remote rows would have contradicted the no-silent-drops rule this PR exists to enforce.

On the whitespace rule, the fix goes further than I asked and correctly so. `isBlankSessionID` is the single Go authority and the SQL sites use a trim set generated from `unicode.White_Space`, applied to all three predicates rather than only the two I named. The failing tests showed the bug was worse than I described: tab-, newline-, CR-, VT-, FF- and Unicode-space-only IDs bypassed doctor entirely, and the backfill guard rejection killed the whole transaction, taking a valid session with it and failing enrollment outright on an affected store.

The correction: an adversarial pass flagged the dead-letter `sync_id` as collision-prone because it hashes `targetKey + Seq` under `ON CONFLICT DO UPDATE`. I checked and that does not hold — `ApplyPulledChunk` overwrites `mutation.Seq` with a strictly increasing counter before applying, and `ApplyPulledMutation` returns early when the sequence has not advanced. The mirror defect was real though: keying on the local cursor breaks idempotency, since the same discarded session bundled into a different chunk lands at a new cursor position and writes a second evidence row. Identity now derives from the mutation's own content, which keeps distinct mutations apart by construction instead of resting on an invariant maintained in two other functions.

A pre-existing defect surfaced in the same area and is filed as #838 rather than widened into this PR: `recordRelationApplyFailureTx` uses a raw `entity_key` as `sync_id`, so blank-keyed relations overwrite each other and the success-path cleanup can orphan the survivor. It needs a design pass on the relation retry contract.

The merge with #819 and #811 unions the repair-check switch — either side alone would have rejected a shipped repair path. Doctor behavior verified against a built binary across local-only, quarantined, genuinely-blocked and pulled-session-evidence stores. 610 store, 436 cmd, 25 diagnostic tests passing, e2e green.

PR #445, opened 2026-05-28
fix(cli): prevent noisy directory consolidation

Maicololiveras:
## 🔗 Linked Issue

Closes #283

---

## 🏷️ PR Type

- [x] `type:bug`: Bug fix
- [ ] `type:feature`: New feature
- [ ] `type:docs`: Documentation only
- [ ] `type:refactor`: Code refactoring (no behavior change)
- [ ] `type:chore`: Maintenance, dependencies, tooling
- [ ] `type:breaking-change`: Breaking change

---

## 📝 Summary

- Prevent `projects consolidate --all` from grouping unrelated projects through noisy shared directories such as `$HOME`.
- Preserve legitimate grouping through small shared-directory sets and project-name similarity.

## 📂 Changes

| File | Change |
|------|--------|
| `cmd/engram/main.go` | Skip shared-directory grouping when more than three distinct projects use the directory. |
| `cmd/engram/consolidate_grouping_test.go` | Cover noisy directories, legitimate shared directories, name similarity, and command-level dry-run behavior. |

## 🧪 Test Plan

- [x] Focused unit tests pass locally: `go test ./cmd/engram -run "Test(GroupSimilarProjects|CmdProjectsConsolidateAllDoesNotMergeNoisySharedDir)" -count=1`
- [ ] Full unit suite: `go test ./...` (not rerun; this Windows host has a known unrelated package-suite failure and timeout)
- [ ] E2E tests: `go test -tags e2e ./internal/server/...` (not required for this CLI-only grouping change)
- [x] Scoped diff verified against `main`: only `cmd/engram/main.go` and `cmd/engram/consolidate_grouping_test.go`

---

## 🤖 Automated Checks

These run automatically and all must pass before merge:

| Check | What it verifies | Status |
|-------|------------------|--------|
| **Check Issue Reference** | PR body contains `Closes #N` / `Fixes #N` / `Resolves #N` | Pending |
| **Check Issue Has status:approved** | Linked issue has `status:approved` label | Pending |
| **Check PR Has type:* Label** | PR has exactly one `type:*` label | Pending |
| **Unit Tests** | `go test ./...` passes | Pending |
| **E2E Tests** | `go test -tags e2e ./internal/server/...` passes | Pending |

---

## ✅ Contributor Checklist

- [x] I linked an approved issue above (`Closes #283`)
- [x] I added exactly one `type:*` label to this PR
- [ ] I ran unit tests locally: `go test ./...`
[...]

PR #801, opened 2026-08-26
fix(cli): make project cleanup safe

### LINKED ISSUE #283: MCP server fails on multi-repo cwd + 3 related bugs (ambiguity, path-as-name, case-sensitivity, unsafe consolidate)

dnlrsls:
## 🔗 Linked Issue

Closes #283

---

## 🏷️ PR Type

- [x] type:bug — Bug fix
- [ ] type:feature — New feature
- [ ] type:docs — Documentation only
- [ ] type:refactor — Code refactoring
- [ ] type:chore — Maintenance, dependencies, tooling
- [ ] type:breaking-change — Breaking change

---

## 📝 Summary

- Replace directory-based project consolidation with deterministic direct name matching in both CLI modes.
- Add path-only cleanup with FK-safe pruning for sessions referenced by soft-deleted observations.
- Report only successful prune operations; this maintainer solution supersedes the incomplete approaches in #445 and #647.

## 📂 Changes

| File | Change |
|------|--------|
| cmd/engram/main.go | Safe consolidation, paths-only pruning, and accurate result reporting |
| cmd/engram/main_test.go | CLI regression coverage for consolidation and pruning |
| internal/store/store.go | Preserve sessions referenced by active or soft-deleted observations |
| internal/store/store_test.go | FK-safe prune and mixed-session coverage |

## 🧪 Test Plan

- [x] Focused CLI regressions: go test ./cmd/engram -run targeted-tests -count=1
- [x] Focused store regressions: go test ./internal/store -run targeted-tests -count=1
- [x] E2E server tests: go test -tags e2e ./internal/server/...
- [ ] Full unit suite: go test ./... exceeded the 10-minute local Windows budget; CI remains authoritative

---

## 🤖 Automated Checks

All repository checks must pass before merge.

---

## ✅ Contributor Checklist

- [x] Linked approved issue #283
- [x] Added exactly one type:* label
- [x] Ran unit tests locally; focused regressions passed and full-suite timeout is disclosed above
- [x] Ran E2E tests locally
- [x] CLI usage output was updated with the behavior change
- [x] Commit follows conventional commits
- [x] No Co-Authored-By trailers

---

## 💬 Notes for Reviewers

Native four-lens RDD approved the exact frozen candidate. The review produced only non-blocking advisory findings and consumed its authority. No contributor branches were modified.

## Summary by CodeRabbit

* **New Features**
  * Added a `--paths-only` option to target path-based project names during pruning.
  * Pruning now reports successful removals and individual errors.
  * Consolidation reports moved records and identifies projects with no merged records.
* **Bug Fixes**
  * Consolidation now groups projects only when their names are directly similar.
  * Pruning preserves sessions linked to soft-deleted observations and removes only unreferenced sessions.
* **Documentation**
  * Updated command usage text to describe the new pruning option.

### DIFF
--- a/cmd/engram/main.go
+++ b/cmd/engram/main.go
@@ -1956,6 +1956,47 @@ func findNormalizationEquivalentProjects(name string, existing []string) []proje
 	return matches
 }
 
+// mergedRecordCount reports how many records a merge actually moved. The store
+// validates every source against the canonical name and fail-closes on the ones
+// it cannot prove normalization-equivalent, so a merge can succeed while moving
+// nothing at all. Callers must report that outcome honestly instead of
+// announcing a completed merge.
+func mergedRecordCount(result *store.MergeResult) int64 {
+	if result == nil {
+		return 0
+	}
+	return result.ObservationsUpdated + result.SessionsUpdated + result.PromptsUpdated
+}
+
+// reportUnmergedSources names the selected sources the store left untouched, so
+// a partially applied merge never reads as a complete one.
+func reportUnmergedSources(sources []string, result *store.MergeResult) {
+	merged := make(map[string]bool, len(result.SourcesMerged))
+	for _, name := range result.SourcesMerged {
+		merged[name] = true
+	}
+
+	// SourcesMerged holds the trimmed spelling the store actually rewrote, while
+	// sources holds the raw spellings the operator selected. Only a source that
+	// is literally the canonical name was a no-op by request.
+	var skipped []string
+	seen := make(map[string]bool, len(sources))
+	for _, source := range sources {
+		if strings.TrimSpace(source) == "" || source == result.Canonical {
+			continue
+		}
+		if merged[strings.TrimSpace(source)] || seen[source] {
+			continue
+		}
+		seen[source] = true
+		skipped = append(skipped, source)
+	}
+	if len(skipped) == 0 {
+		return
+	}
+	fmt.Printf("  Not merged (no records moved): %s\n", strings.Join(skipped, ", "))
+}
+
 func cmdProjectsConsolidate(cfg store.Config) {
 	doAll := false
 	dryRun := false
@@ -2067,10 +2108,17 @@ func cmdProjectsConsolidate(cfg store.Config) {
 			fatal(err)
 		}
 
-		fmt.Printf("Done! Merged into %q:\n", result.Canonical)
+		if mergedRecordCount(result) == 0 {
+			fmt.Printf("Nothing merged into %q: the store moved no records for the %d selected project(s).\n",
+				result.Canonical, len(sources))
+			return
+		}
+
+		fmt.Printf("Done! Merged %d project(s) into %q:\n", len(result.SourcesMerged), result.Canonical)
 		fmt.Printf("  Observations: %d\n", result.ObservationsUpdated)
 		fmt.Printf("  Sessions:     %d\n", result.SessionsUpdated)
 		fmt.Printf("  Prompts:      %d\n", result.PromptsUpdated)
+		reportUnmergedSources(sources, result)
 		return
 	}
 
@@ -2188,8 +2236,14 @@ func cmdProjectsConsolidate(cfg store.Config) {
 			fmt.Println()
 			continue
 		}
-		fmt.Printf("  Merged: %d obs, %d sessions, %d prompts\n",
-			result.ObservationsUpdated, result.SessionsUpdated, result.PromptsUpdated)
+		if mergedRecordCount(result) == 0 {
+			fmt.Printf("  Nothing merged into %q: the store moved no records for the %d selected project(s).\n",
+				mergeCanonical, len(sources))
+		} else {
+			fmt.Printf("  Merged: %d obs, %d sessions, %d prompts\n",
+				result.ObservationsUpdated, result.SessionsUpdated, result.PromptsUpdated)
+			reportUnmergedSources(sources, result)
+		}
 
 		if renameTarget != "" && renameTarget != mergeCanonical {
 			migrateResult, err := s.MigrateProject(mergeCanonical, renameTarget)
--- a/cmd/engram/main.go
+++ b/cmd/engram/main.go
@@ -95,7 +95,8 @@ var (
 	storeDeleteProject     = func(s *store.Store, name string, hard bool) (*store.DeleteProjectResult, error) {
 		return s.DeleteProject(name, hard)
 	}
-	storeTimeline = func(s *store.Store, observationID int64, before, after int) (*store.TimelineResult, error) {
+	storePruneProject = func(s *store.Store, name string) (*store.PruneResult, error) { return s.PruneProject(name) }
+	storeTimeline     = func(s *store.Store, observationID int64, before, after int) (*store.TimelineResult, error) {
 		return s.Timeline(observationID, before, after)
 	}
 	storeFormatContext = func(s *store.Store, project, scope string) (string, error) { return s.FormatContext(project, scope) }
@@ -1850,7 +1851,7 @@ func cmdProjects(cfg store.Config) {
 		fmt.Fprintf(os.Stderr, "unknown projects subcommand: %s\n", subCmd)
 		fmt.Fprintln(os.Stderr, "usage: engram projects list")
 		fmt.Fprintln(os.Stderr, "       engram projects consolidate [--all] [--dry-run]")
-		fmt.Fprintln(os.Stderr, "       engram projects prune [--dry-run]")
+		fmt.Fprintln(os.Stderr, "       engram projects prune [--dry-run] [--paths-only]")
 		exitFunc(1)
 	}
 }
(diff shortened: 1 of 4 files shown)


Alan-TheGentleman (changes requested):
Requesting changes. Project cleanup still calls fuzzy `project.FindSimilar` substring and Levenshtein matching, then merges projects that are not proven equivalent. Case-only variants can report completion with zero data movement. Restrict merging to normalized equivalence validated at the store boundary, preferably on top of corrected #810, and add a regression for the case-only path.

Alan-TheGentleman (approved):
Resolved on top of the corrected #810, as the review asked.

Fuzzy matching no longer reaches any merge: grouping is by `store.NormalizeProject` equivalence and `project.FindSimilar` is unreachable from every cleanup path — its one remaining non-test caller is a save-time warning in MCP, which surfaces rather than merges. There is coverage pinning that, across both routes, for substring (`engram` / `engram-memory`) and Levenshtein (`engram` / `engramm`) neighbours.

The zero-movement report was still real and is fixed. The reachable path is trimmed variants, not case-only ones: `" engram "` normalizes to `engram`, is offered as a candidate, passes validation, then gets skipped by `projectMergeSourceVariants` — and the CLI still printed `Done! Merged into "engram"` with 0/0/0. The report is now derived from what the store actually moved: nothing moved says so, a partial merge names the sources left untouched, and the count reflects `SourcesMerged` rather than what was requested. The case-only regression is in too, cross-checked against real row counts.

Full suite and e2e green.

One behavior change worth your eye, inherited from #810 rather than introduced here: cleanup no longer surfaces similar-looking projects at all — a non-equivalent neighbour just yields `No similar project names found`. Left as main has it. Say the word if you want that surfacing back as a read-only suggestion.

PR #809, opened 2026-08-26
fix(store): preserve UTF-8 truncation boundaries

### LINKED ISSUE #697: bug(store): oversized-body truncation slices UTF-8 by bytes, and the prompt path truncates with no warning
### 📋 Pre-flight Checks

### 📝 Bug Description

Two defects in the oversized-body truncation path. Line references are against the `v1.20.0` tag, commit `ba9e46ce`.

**First, what is *not* being reported.** #41 ("Silent truncation of long Engram entries") was fixed and closed: the cap moved from 2,000 to 50,000 and `mem_save` / `mem_update` now append an explicit `⚠ WARNING` (`internal/mcp/mcp.go:1298`, `internal/mcp/mcp.go:1430`). That remedy works, and this issue does not claim observation truncation is silent.

What follows is what that fix did not cover.

**Defect 1 — the truncation slices UTF-8 by bytes and can corrupt the boundary character.**

Three call sites truncate identically:

```go
if len(content) > s.cfg.MaxObservationLength {
	content = content[:s.cfg.MaxObservationLength] + "... [truncated]"
}
```

- `internal/store/store.go:2259-2260` — `AddObservation`
- `internal/store/store.go:2644-2645` — `preparePromptContent`
- `internal/store/store.go:2887-2888` — `UpdateObservation`

`len()` counts bytes and `content[:n]` slices bytes, so when byte 50,000 falls inside a multi-byte rune the rune is split and an invalid UTF-8 sequence is persisted. Any non-ASCII body over the cap can be affected — the higher the proportion of accented, CJK or emoji characters, the higher the chance the boundary lands mid-rune.

Worth noting that the codebase already treats this as something to get right elsewhere: `truncate` (`internal/store/store.go:6350`) converts to `[]rune` before slicing, and `TestTruncateUTF8` (`internal/store/store_test.go:4814`) covers it with Spanish-accent, emoji and mixed cases. That helper is only used for display previews (`store.go:3331`, `3342`, `3351`, `3360`). The three persistence sites never got the same treatment, and no test covers them.

The same byte/character conflation has a second, more likely consequence: the cap is enforced in **bytes** while the configuration field, the warning text and the docs all say *chars*. A body of 50,100 characters that encodes to 50,101 bytes is truncated, and a CJK body reaches the cap at roughly 16,600 characters rather than 50,000. The warning then reports byte counts labelled as characters, so the numbers an agent receives do not match what it sent.

**Defect 2 — the prompt path truncates with no warning at all.**

`preparePromptContent` (`internal/store/store.go:2643-2647`) applies the same cap, but the `mem_save_prompt` handler (`internal/mcp/mcp.go:1595-1609`) never compares lengths and never emits the `⚠ WARNING` its two sibling handlers do. `AddPrompt` returns no truncation signal for it to surface. So for prompts, the exact behaviour #41 was closed for is still present.

**Minor, same code path.** The warning's length check runs on the *pre-redaction* content (`internal/mcp/mcp.go:1262`, `internal/mcp/mcp.go:1418-1421`) while the store applies `stripPrivateTags` before measuring. Since redaction only ever shortens a body, this can warn when nothing was actually truncated, but it cannot miss a real truncation. Cosmetic, noted only because a fix touches the same lines.

### 🔄 Steps to Reproduce

**Defect 1 (boundary corruption):**

1. Build a body where byte 50,000 falls inside a multi-byte rune — for example 49,999 ASCII bytes followed by `é` (`C3 A9`), then any tail.
2. Record its byte count and SHA-256.
3. Save it through `mem_save`.
4. Read it back with `mem_get_observation` (full content, not a search preview).
5. Inspect the bytes immediately before the `... [truncated]` marker.

Reproduced against a real `Store` built with `New(FallbackConfig(t.TempDir()))` on a clean checkout of `v1.20.0`, going through `AddObservation` and reading back with `GetObservation` — no reimplementation of the truncation logic:

```
=== RUN   TestOversizedBodyKeepsValidUTF8/2-byte_rune_(é)
    observation id=1  source byte[cap]=0xa9  stored byte[cap-1]=0xc3
    persisted body is NOT valid UTF-8; first invalid byte at index 49999 (0xc3); source=50101 bytes stored=50015 bytes
=== RUN   TestOversizedBodyKeepsValidUTF8/3-byte_rune_(€)
    observation id=2  source byte[cap]=0x82  stored byte[cap-1]=0xe2
    persisted body is NOT valid UTF-8; first invalid byte at index 49999 (0xe2); source=50102 bytes stored=50015 bytes
=== RUN   TestOversizedBodyKeepsValidUTF8/4-byte_rune_(emoji)
    observation id=3  source byte[cap]=0x9f  stored byte[cap-1]=0xf0
    persisted body is NOT valid UTF-8; first invalid byte at index 49999 (0xf0); source=50103 bytes stored=50015 bytes
```

In each case the surviving byte at index 49,999 is the lead byte of the split rune (`C3`, `E2`, `F0`) with its continuation bytes cut away.

One note for anyone reproducing this: give each case a distinct title. With a shared title the three saves dedupe onto one observation and the later cases silently read back the first one's content, which looks like a result but is not.

**Defect 2 (prompt path):**

1. Call `mem_save_prompt` with content longer than `MaxObservationLength`.
2. Read the tool response.
3. Retrieve the prompt and compare against the source.

### ✅ Expected Behavior

1. Truncation should cut on a rune boundary so the persisted body is always valid UTF-8.
2. The cap and every count reported to the caller should use one consistent unit, and the warning text should name the unit it actually measures.
3. `mem_save_prompt` should surface the same truncation warning as `mem_save` and `mem_update`, so the guarantee #41 established holds on every path that applies the cap.

### ❌ Actual Behavior

1. `content[:MaxObservationLength]` can split the boundary rune and persist invalid UTF-8.
2. The cap is applied in bytes while being reported as characters, so non-ASCII bodies are truncated earlier than the documented limit and the reported numbers do not match the input.
3. Oversized prompts are truncated with no warning on any layer.

### Operating System

Linux (Other) — Ubuntu 26.04, x86_64

### Engram Version

1.20.0

### Agent / Client

Other (Codex CLI 0.146.0)

### 📋 Relevant Logs

No error envelope is produced — both defects occur on paths that report success.

### Additional Context

**Proposed fix.** Cut on a rune boundary in a single shared helper used by all three call sites:

```go
func truncateForStorage(s string, max int) string {
	if len(s) <= max {
		return s
	}
	cut := max
	for cut > 0 && !utf8.RuneStart(s[cut]) {
		cut--
	}
	return s[:cut] + "... [truncated]"
}
```

This backtracks at most three bytes and leaves ASCII behaviour byte-for-byte identical, so existing tests that assert the `... [truncated]` marker keep passing. Verified on a patched tree: the helper restores valid UTF-8 for all three rune widths, never backtracks more than three bytes, and leaves ASCII truncation unchanged.

Environment for the run above: Go 1.26.0, `go build ./...` clean on the unmodified `v1.20.0` tree.

For defect 2, threading a truncation signal out of the prompt path lets the `mem_save_prompt` handler reuse the message already used at `internal/mcp/mcp.go:1298`.

**Scope note.** This is deliberately narrow. It does **not** explain the sub-50,000 incomplete mirrors described in `Gentleman-Programming/gentle-ai#2224` — those bodies never reach the cap, so a different mechanism is responsible and that one stays tracked there. Filing this separately was suggested in that thread.

Happy to open the PR, with a regression test asserting the persisted body is valid UTF-8 when the cap falls inside a multi-byte rune, once this carries `status:approved`.

dnlrsls:
## 🔗 Linked Issue

Closes #697

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix

---

## 📝 Summary

- Preserve valid UTF-8 when oversized observation or prompt content is truncated.
- Keep the existing byte-prefix cap and truncation marker unchanged.
- Cover 2-, 3-, and 4-byte rune boundaries across every affected persistence path.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | Add a shared UTF-8-safe byte-boundary truncation helper and use it for observation add, prompt preparation, and observation update. |
| `internal/store/store_test.go` | Add boundary and persistence-route regression coverage. |

## 🧪 Test Plan

- [x] Focused regression tests pass: `go test ./internal/store -run 'Test(TruncateContentPreservesUTF8BytePrefix|PersistenceRoutesTruncateContentAtUTF8Boundary)$'`
- [ ] Full local store package: `go test ./internal/store` (Windows SQLite TempDir cleanup locks remain in `TestMigrate_Idempotent` and `TestNewErrorBranches`; new tests pass)
- [ ] E2E tests: not run; this change is isolated to store truncation
- [x] `git diff --check` passes
- [x] Gentle AI RDD four-lens review approved the frozen candidate

---

## 🤖 Automated Checks

---

## ✅ Contributor Checklist

- [x] I linked an approved issue above (`Closes #697`)
- [x] Docs are not required because the documented byte-limit semantics are unchanged

---

## 💬 Notes for Reviewers

The correction intentionally preserves the existing byte cap instead of converting it to a rune-count cap. When the cap lands inside a rune, the prefix backs up to the rune start before appending the existing marker.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed content truncation to preserve valid UTF-8 characters.
  * Prevented multibyte characters from being split at the configured byte limit.
  * Ensured truncation markers remain within the limit for saved observations and prompts.
  * Applied limits after redaction and reported accurate byte counts.

* **New Features**
  * Added truncation metadata and warnings to save, update, and prompt responses.

* **Documentation**
  * Clarified byte-based limits, redaction behavior, truncation metadata, and warnings.

* **Tests**
  * Added coverage for UTF-8-safe truncation, metadata, warnings, and persistence scenarios.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -858,7 +858,7 @@ Save structured observations. The tool description teaches agents the format:
 
 Exact duplicate saves are deduplicated in a rolling time window using a normalized content hash + project + scope + type + title.
 When `topic_key` is provided, `mem_save` upserts the latest observation in the same `project + scope + topic_key`, incrementing `revision_count` and attributing it to the latest writer session.
-Save responses include lifecycle metadata for the saved observation: computed `state` (`active` or `needs_review`) and `review_after` when the observation type has a review cycle.
+Save responses include lifecycle metadata for the saved observation: computed `state` (`active` or `needs_review`) and `review_after` when the observation type has a review cycle. Content is redacted before the configured storage limit is applied; that limit and truncation metadata (`original_bytes`, `limit_bytes`) are UTF-8 bytes. MCP save/update responses include `truncated`, and warn when truncation occurs.
 
 ### mem_update
 
@@ -885,7 +885,7 @@ Delete an observation by ID. Uses soft-delete by default (`deleted_at`); optiona
 
 ### mem_save_prompt
 
-Save user prompts — records what the user asked so future sessions have context about user goals.
+Save user prompts — records what the user asked so future sessions have context about user goals. It applies the same post-redaction byte limit and truncation metadata as `mem_save`; `mem_save_prompt` warns when it truncates.
 When called in the same MCP process, this also feeds process-local current prompt context used by later `mem_save` calls with `capture_prompt=true`. The same MCP process lifecycle must receive the prompt context before the later save; prompt capture is best-effort and `mem_save` still succeeds when no context is available.
 
 ### mem_context
--- a/internal/mcp/mcp.go
+++ b/internal/mcp/mcp.go
@@ -68,6 +68,13 @@ var loadMCPStats = func(s *store.Store) (*store.Stats, error) {
 	return s.Stats()
 }
 
+func truncationWarning(metadata store.TruncationMetadata) string {
+	if !metadata.Truncated {
+		return ""
+	}
+	return fmt.Sprintf("\n⚠ WARNING: Content was truncated from %d to %d bytes. Consider splitting into smaller observations.", metadata.OriginalBytes, metadata.LimitBytes)
+}
+
 func currentWorkingDirectory() string {
 	cwd, err := os.Getwd()
 	if err != nil {
@@ -1261,7 +1268,7 @@ func handleSave(s *store.Store, cfg MCPConfig, activity *SessionActivity) server
 		// Ensure the implicit MCP session exists with the current working directory.
 		_ = ensureImplicitSessionWithCWD(s, sessionID, project)
 
-		truncated := len(content) > s.MaxObservationLength()
+		truncation := s.ContentTruncation(content)
 
 		savedID, err := s.AddObservation(store.AddObservationParams{
 			SessionID: sessionID,
@@ -1296,9 +1303,7 @@ func handleSave(s *store.Store, cfg MCPConfig, activity *SessionActivity) server
 		if topicKey == "" && suggestedTopicKey != "" {
 			msg += fmt.Sprintf("\nSuggested topic_key: %s", suggestedTopicKey)
 		}
-		if truncated {
-			msg += fmt.Sprintf("\n⚠ WARNING: Content was truncated from %d to %d chars. Consider splitting into smaller observations.", len(content), s.MaxObservationLength())
-		}
+		msg += truncationWarning(truncation)
 		if normWarning != "" {
 			msg += "\n" + normWarning
 		}
@@ -1308,7 +1313,7 @@ func handleSave(s *store.Store, cfg MCPConfig, activity *SessionActivity) server
 
 		// Post-transaction conflict candidate detection (REQ-001).
 		// Errors are logged and swallowed — detection failure never fails the save.
-		extra := map[string]any{}
+		extra := map[string]any{"truncation": truncation}
 		// Build CandidateOptions, forwarding any MCPConfig overrides.
 		// nil fields mean "use store defaults"; explicit pointer values override.
 		candOpts := store.CandidateOptions{
@@ -1417,9 +1422,10 @@ func handleUpdate(s *store.Store) server.ToolHandlerFunc {
 			return mcp.NewToolResultError("provide at least one field to update"), nil
 		}
 
-		var contentLen int
+		var truncation *store.TruncationMetadata
 		if update.Content != nil {
-			contentLen = len(*update.Content)
+			metadata := s.ContentTruncation(*update.Content)
+			truncation = &metadata
 		}
 
 		obs, err := s.UpdateObservation(id, update)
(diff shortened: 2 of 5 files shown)


Alan-TheGentleman (changes requested):
Requesting changes. `mem_save_prompt` still truncates silently, and its documented character limit is implemented in bytes. Expose post-redaction truncation metadata and provide warning parity for callers. Add unit coverage for the unit contract and truncation behavior before treating #697 as closed.

Alan-TheGentleman (approved):
The requested prompt warning, post-redaction byte metadata, and regression coverage are present. The exact correction passed RDD, full local tests, real MCP process validation, and all required CI checks. Approved for merge.

PR #820, opened 2026-08-27
fix(store): escape quotes in any-mode FTS queries

### LINKED ISSUE #664: bug(store): #574 double-quote escaping fix not applied to sanitizeFTSCandidates (match_mode="any" still errors)
### 📋 Pre-flight Checks

### 📝 Bug Description

The interior-double-quote fix from #574 was applied to `sanitizeFTS`, but **not** to its sibling `sanitizeFTSCandidates` — the builder used by `match_mode: "any"`. So the exact bug #574 closed is still reachable, just through the other code path.

`main` @ `763a6ba4`:

**`internal/store/store.go:6585` — fixed:**
```go
func sanitizeFTS(query string) string {
	words := strings.Fields(query)
	for i, w := range words {
		w = strings.Trim(w, `"`)
		// Double interior double-quotes: FTS5 escapes a literal " inside a
		// quoted phrase by doubling it (""). ... See #574.
		w = strings.ReplaceAll(w, `"`, `""`)
		words[i] = `"` + w + `"`
	}
	return strings.Join(words, " ")
}
```

**`internal/store/relations.go:1039` — not fixed:**
```go
func sanitizeFTSCandidates(title string) string {
	words := strings.Fields(title)
	// ...
	for _, w := range words {
		w = strings.Trim(w, `"`)          // <-- only trims the ENDS
		if w != "" {
			quoted = append(quoted, `"`+w+`"`)
		}
	}
	return strings.Join(quoted, " OR ")
}
```

`strings.Trim` only removes leading/trailing quotes, so an **interior** quote survives and produces an unterminated FTS5 string literal.

**`internal/store/store.go:3167` — where `match_mode: "any"` selects it:**
```go
if opts.MatchMode == "any" {
	ftsQuery = sanitizeFTSCandidates(query)
} else {
	ftsQuery = sanitizeFTS(query)
}
```

Given the query `hello"world`:

| builder | produced FTS5 query | result |
|---|---|---|
| `sanitizeFTS` (`match_mode: "all"`) | `"hello""world"` | ✅ valid |
| `sanitizeFTSCandidates` (`match_mode: "any"`) | `"hello"world"` | ❌ `unterminated string` |

`sanitizeFTSCandidates` is also called from the relation-scan path in `relations.go`, where observation **titles** become the query — so any saved title containing a `"` can trip the same error there.

### 🔄 Steps to Reproduce

Self-contained proof of the mechanism, no Go needed (any SQLite ≥ 3.9 with FTS5):

```sql
CREATE VIRTUAL TABLE t USING fts5(x);
INSERT INTO t VALUES ('hello world');

-- what sanitizeFTSCandidates produces for  hello"world
SELECT * FROM t WHERE t MATCH '"hello"world"';   -- Error: unterminated string

-- what sanitizeFTS produces for the same input
SELECT * FROM t WHERE t MATCH '"hello""world"';  -- OK, 1 row
```

Through the MCP tool on `main`:

```
mem_search(query: "hello\"world", match_mode: "any")
```

1. Save any observation.
2. Call `mem_search` with a query containing an interior double-quote and `match_mode: "any"`.
3. Search fails instead of returning results.

### ✅ Expected Behavior

`match_mode: "any"` handles interior double-quotes exactly like `match_mode: "all"` does after #574 — the quote is escaped by doubling, and the search returns results instead of erroring.

### ❌ Actual Behavior

```
Search error: search: SQL logic error: unterminated string (1). Try simpler keywords.
```

(Captured on 1.11.0, which predates #574 and therefore fails on **both** modes. On `main` the `"all"` path is fixed and only the `"any"` path still fails — that asymmetry is what this issue is about, and it is established from the source above plus the SQLite repro, not from running 1.11.0.)

### 🖥️ Environment

- **OS:** Windows
- **Engram Version:** 1.11.0 locally; bug analysis is against `main` @ `763a6ba4`
- **Agent / Client:** Claude Code

### 💡 Additional Context

The fix looks like one line — reusing the same escaping `sanitizeFTS` already does:

```go
w = strings.Trim(w, `"`)
w = strings.ReplaceAll(w, `"`, `""`)   // same as sanitizeFTS, see #574
```

Two things worth considering alongside it, entirely your call:

1. **Factor the escaping into one helper** used by both builders, so the two paths can't drift apart again. That drift is what produced this issue.
2. A regression test on the shared helper covering `foo"bar`, `"foo`, `foo"`, and `"` alone would pin it.

Happy to open the PR with the fix plus tests once this is approved — just let me know if you'd prefer the one-line version or the shared-helper refactor.

dnlrsls:
## 🔗 Linked Issue

Closes #664

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix

---

## 📝 Summary

- Escape embedded double quotes in shared FTS candidate sanitization.
- Preserve already doubled FTS5 quote escapes.
- Cover `match_mode="any"` and relation candidate lookup regressions.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/relations.go` | Double raw interior quotes while preserving existing escaped pairs. |
| `internal/store/store_test.go` | Add any-mode quoted-query regression and escaped-query negative control. |
| `internal/store/relations_test.go` | Add `FindCandidates` regression for interior quotes. |

## 🧪 Test Plan

- [x] Focused regressions pass: `go test ./internal/store -run '^(TestSearchMatchMode_AnyEscapesInteriorQuotes|TestFindCandidates_EscapesInteriorQuotes)$' -count=1`
- [x] Pre-fix regressions reproduced `unterminated string`; post-fix regressions pass.
- [ ] Full local package suite: blocked on Windows by pre-existing temp SQLite file locks in `TestMigrate_Idempotent` and `TestNewErrorBranches/fails_when_migration_encounters_conflicting_object`.
- [ ] E2E tests: not applicable to this local store sanitizer change.

---

## 🤖 Automated Checks

---

## ✅ Contributor Checklist

- [x] I linked an approved issue above (`Closes #664`)
- [ ] I ran the full unit suite locally; focused candidate verification passed and the unrelated Windows blocker is documented above
- [ ] I ran e2e tests locally; they are not applicable to this store-only correction
- [x] Docs updated (not required; no user-facing contract changed)

---

## 💬 Notes for Reviewers

The frozen candidate completed the native four-lens RDD review successfully. The broad local package run was blocked only by unrelated Windows temp-database cleanup locks; a candidate-specific substitute verifier passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search reliability for titles containing double quotes.
  * Queries with embedded quotation marks now return the expected matches without errors.

* **Tests**
  * Added coverage for quoted titles in candidate finding and search matching.

### DIFF
--- a/internal/store/relations.go
+++ b/internal/store/relations.go
@@ -1080,7 +1080,20 @@ func sanitizeFTSCandidates(title string) string {
 	for _, w := range words {
 		w = strings.Trim(w, `"`)
 		if w != "" {
-			quoted = append(quoted, `"`+w+`"`)
+			var escaped strings.Builder
+			for i := 0; i < len(w); i++ {
+				if w[i] == '"' {
+					// FTS5 escapes a literal quote by doubling it. Preserve an
+					// already escaped pair so callers do not get double escaped.
+					escaped.WriteString(`""`)
+					if i+1 < len(w) && w[i+1] == '"' {
+						i++
+					}
+					continue
+				}
+				escaped.WriteByte(w[i])
+			}
+			quoted = append(quoted, `"`+escaped.String()+`"`)
 		}
 	}
 	return strings.Join(quoted, " OR ")
--- a/internal/store/relations_test.go
+++ b/internal/store/relations_test.go
@@ -110,6 +110,25 @@ func TestFindCandidates_HappyPath(t *testing.T) {
 	}
 }
 
+func TestFindCandidates_EscapesInteriorQuotes(t *testing.T) {
+	s := setupRelationsStore(t)
+	_, _ = addTestObs(t, s, `hello"world candidate`, "decision", "testproject", "project")
+	savedID, _ := addTestObs(t, s, `hello"world source`, "decision", "testproject", "project")
+
+	candidates, err := s.FindCandidates(savedID, CandidateOptions{
+		Project:   "testproject",
+		Scope:     "project",
+		Limit:     3,
+		BM25Floor: ptrFloat64(-10.0),
+	})
+	if err != nil {
+		t.Fatalf("FindCandidates: %v", err)
+	}
+	if len(candidates) != 1 {
+		t.Fatalf("expected 1 candidate, got %d", len(candidates))
+	}
+}
+
 // TestFindCandidates_EarlyBreakDoesNotSelfBlockWithSingleConnection verifies
 // that FindCandidates closes its FTS rows before follow-up QueryRow/Exec calls.
 // With SetMaxOpenConns(1), leaving rows open after the early-break path can
--- a/internal/store/store_test.go
+++ b/internal/store/store_test.go
@@ -9396,6 +9396,35 @@ func TestSearchMatchMode_Any(t *testing.T) {
 	}
 }
 
+func TestSearchMatchMode_AnyEscapesInteriorQuotes(t *testing.T) {
+	s := newTestStore(t)
+	if err := s.CreateSession("s-matchmode-quotes", "engram", "/tmp"); err != nil {
+		t.Fatalf("create session: %v", err)
+	}
+	if _, err := s.AddObservation(AddObservationParams{
+		SessionID: "s-matchmode-quotes",
+		Type:      "decision",
+		Title:     `hello"world`,
+		Content:   "quoted search target",
+		Project:   "engram",
+		Scope:     "project",
+	}); err != nil {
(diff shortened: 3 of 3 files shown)


Alan-TheGentleman (approved):
Approved. The raw interior-quote failure is fixed and CI is green. As a non-blocking follow-up, consolidate quote escaping behind a shared helper and cover runs of one, two, and three quotes.

PR #830, opened 2026-08-27
fix(store): close sqlite handles after migration failure

### LINKED ISSUE #728: internal/store tests fail on Windows: t.TempDir cleanup races SQLite handles
### 📋 Pre-flight Checks

### 🐛 Bug Description

`go test ./internal/store/...` fails on Windows on current `main`. Two tests are
red on every run:

- `TestMigrate_Idempotent`
- `TestNewErrorBranches`

The cause is not the code under test — it is how the temp directory is cleaned
up. These tests use `t.TempDir()`, which removes the directory as soon as the
test returns. On Windows a file cannot be unlinked while a handle to it is still
open, and SQLite still holds `engram.db` at that moment, so `RemoveAll` fails.
Since Go treats a failed `t.TempDir()` cleanup as a test failure, the test goes
red even though the assertions passed.

On Linux and macOS the same code is fine, because unlinking an open file is
allowed there. That is probably why CI has not caught it.

This came up while I was looking at #550 — that PR touched the same area, so I
wanted to write the underlying problem up separately rather than leave it
implicit.

### 🔁 Steps to Reproduce

On Windows:

```
go test ./internal/store/... -count=1
```

### ✅ Expected Behaviour

`internal/store` passes on Windows as it does on Linux and macOS.

### ❌ Actual Behaviour

```
--- FAIL: TestMigrate_Idempotent (1.89s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat
    C:\...\Temp\TestMigrate_Idempotent3559259003\001\engram.db:
    The process cannot access the file because it is being used by another process
--- FAIL: TestNewErrorBranches (1.92s)
FAIL    github.com/Gentleman-Programming/engram/internal/store
```

Reproduced on consecutive runs, so it is deterministic rather than flaky.

Environment: Windows 11, `go test` run locally against `main`.

### 📦 Affected Area

`internal/store` — test helpers only. No production code is involved.

### 📎 Additional Context

I have a fix ready if it is useful. It is test-only and about 70 lines:

- A `closeAndRemoveStore(t, dir, s)` helper that closes the store first, then
  retries `RemoveAll` for up to 5s to cover the window before the OS releases
  the handle.
- The affected tests switch from `t.TempDir()` to `os.MkdirTemp` plus that
  helper.
- If the retries are exhausted it logs rather than fails, so a leftover temp
  directory never turns into a spurious red build.

With it, `go test ./internal/store/...` passes on Windows, and behaviour on
Linux and macOS is unchanged since `RemoveAll` succeeds on the first attempt
there.

I may well be missing a simpler angle — if you would rather solve it a
different way, or skip it because Windows is not a supported target, that is
completely fine and I would rather know than guess. Thanks either way, and
thanks for engram.

### LINKED ISSUE #825: bug(store): migration error paths retain SQLite handles on Windows
### 📋 Pre-flight Checks

### 📝 Bug Description
SQLite handles remain open on migration/error paths. `Store.New` does not close the database after a post-open migration failure, and migration tests can leave query rows open, causing `t.TempDir` cleanup to fail on Windows.

### 🔄 Steps to Reproduce
1. Run the affected store migration/error tests on Windows.
2. Trigger a migration failure after `sql.Open` and an idempotency query.
3. Observe TempDir cleanup.

### ✅ Expected Behavior
Production closes the DB on every post-open failure and tests close query rows before cleanup.

### ❌ Actual Behavior
`engram.db` remains in use and cleanup fails.

### Operating System
Windows

### Engram Version
engram 1.20.1-0.20260720141607-763a6ba43271

### Agent / Client
OpenCode

### 📋 Relevant Logs
```shell
The process cannot access the file because it is being used by another process.
```

### 💡 Additional Context
The selected scope includes production error cleanup and explicit test row ownership in one store workspace.

dnlrsls:
## 🔗 Linked Issue

Closes #728
Closes #825

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix

---

## 📝 Summary

- Closes the SQLite database on every post-open constructor failure.
- Explicitly closes idempotency-test query rows.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | Close the SQLite database after pragma or migration setup failures. |
| `internal/store/store_migration_test.go` | Close rows returned by the idempotency migration query. |

## 🧪 Test Plan

- [x] Focused package test passed locally: `go test ./internal/store -count=1` (34.451s)
- [ ] Full unit suite not run locally: `go test ./...`
- [ ] E2E tests not run locally: `go test -tags e2e ./internal/server/...`
- [ ] Manual testing not run locally.

---

## 🤖 Automated Checks

| **Check Issue Reference** | PR body contains `Closes #N` / `Fixes #N` / `Resolves #N` | Pending |
| **Check Issue Has status:approved** | Linked issue has `status:approved` label | Pending |
| **Check PR Has type:* Label** | PR has exactly one `type:*` label | Pending |
| **Unit Tests** | `go test ./...` passes | Pending |
| **E2E Tests** | `go test -tags e2e ./internal/server/...` passes | Pending |

---

## ✅ Contributor Checklist

- [x] I linked approved issues above (`Closes #728` and `Closes #825`)
- [ ] I ran the full unit suite locally: `go test ./...`
- [x] Docs not required

---

## 💬 Notes for Reviewers

Native four-lens RDD approved this exact candidate before delivery.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a resource leak that could occur when database setup or migration failed.
  * Improved cleanup during unsuccessful database initialization.


dnlrsls:
Thanks. I checked this against the current test seams, and I’m keeping the broader initialization-error matrix as follow-up coverage rather than expanding this PR.

`openDB` only injects failures before a database handle exists. Deterministic post-open pragma and migration cleanup assertions for both constructors would require a custom SQL driver or a new constructor seam, which is a much larger test abstraction than this localized fix. The original Windows cleanup regression was red before the change and green after it, `go test ./internal/store -count=1` passed locally, and Unit/E2E CI are green. Empty-database initialization is already exercised broadly through `newTestStore`, and the existing `%w` wrapping preserves original causes.

The docstring percentage warning is also non-actionable for these internal constructors and test functions. Given the minimal-risk result, I’m treating both items as non-blocking follow-ups.

PR #844, opened 2026-08-27
fix(sync): propagate prompt deletes to the cloud

### LINKED ISSUE #837: Prompt deletes are dropped during cloud chunk materialization, so deleted prompts linger on the dashboard
## 📝 Bug Description

A prompt deleted locally never propagates its delete to the cloud. The prompt stays on the dashboard indefinitely, and no error is surfaced anywhere.

Found while adding the end-to-end `/prompts` and project-scoped sync evidence required for #706. It is **not** the cause of #706 — that reports too *few* prompts reaching the dashboard, this causes too *many* — so it is filed separately rather than folded into that work.

## Mechanism

`materializedChunkMutations` (`internal/cloud/cloudstore/cloudstore.go:402-405`) skips every `chunk.Mutations` entry whose entity is not `relation`:

```go
for i, mutation := range chunk.Mutations {
    if strings.TrimSpace(mutation.Entity) != store.SyncEntityRelation {
        continue
    }
```

The stated assumption in the comment just above is that "Session/observation/prompt mutations are already materialized from the typed collections above, so they are skipped to avoid duplicate `cloud_mutations` rows."

That assumption holds for upserts. It breaks for prompt deletes:

1. `Store.DeletePrompt` (`internal/store/store.go:2984`) hard-deletes the row: `DELETE FROM user_prompts WHERE id = ?`.
2. With the row gone, `filterByPendingMutations` cannot place the prompt in `chunk.Prompts`.
3. So the delete can only travel as a `chunk.Mutations` entry — where the loop above discards it.

Observations are not affected the same way, which is why this has stayed hidden: `DeleteObservation(id, hardDelete bool)` (`internal/store/store.go:3117`) soft-deletes by default, keeping the row with `deleted_at` set so it still rides in `chunk.Observations`. Only prompts are hard-delete-only, so only prompts lose the delete.

Worth checking as part of the fix: `DeleteObservation` called with `hardDelete: true` removes the row the same way, so it plausibly loses its delete through the identical path.

## 🔄 Steps to Reproduce

1. Enroll a project in cloud sync and save a prompt, letting it sync to the dashboard.
2. Delete that prompt locally.
3. Let sync run.
4. The prompt is gone locally and still present on the dashboard. No error is reported.

Direct unit-level reproduction: build a chunk carrying one prompt `delete` mutation plus one session, pass it to `materializedChunkMutations`, and observe the result contains exactly one entry — the session upsert — and zero prompt entries.

## ✅ Expected Behavior

A locally deleted prompt propagates its delete to the cloud and disappears from the dashboard, the same way a deleted observation does.

## ❌ Actual Behavior

The prompt delete is silently dropped during chunk materialization. The prompt lingers on the dashboard for the life of the account, and nothing surfaces the discrepancy.

## Additional Context

There is currently no test covering prompt delete propagation, which is why the gap survived. A fix should carry regression coverage for prompt deletes reaching the cloud, and should confirm the hard-deleted-observation path while it is in there.

Alan-TheGentleman:
## 🔗 Linked Issue

Closes #837

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix

---

## 📝 Summary

- `materializedChunkMutations` skipped every `chunk.Mutations` entry whose entity was not `relation`, on the assumption that session/observation/prompt mutations always arrive through the typed collections. That holds for **upserts**, not for **deletes**: `DeletePrompt` hard-deletes the row, so the prompt cannot ride in `chunk.Prompts` and its tombstone was discarded — the prompt stayed on the dashboard forever.
- Non-relation **deletes** are now materialized from `chunk.Mutations`; non-relation **upserts** are still skipped, so the no-duplicate-`cloud_mutations`-rows invariant the skip existed to protect is preserved. Deletes are appended after the typed entries, so the tombstone gets the higher `seq` and wins the `cloud_mutations` replay.
- The same root cause was also dropping **hard-deleted observations** (`DeleteObservation(id, true)` removes the row exactly like `DeletePrompt`), and **resurrecting soft-deleted observations**: their surviving row still materializes an `upsert` whose `deleted_at` the dashboard's observation mutation payload does not read, so the replay put them back after the chunk pass had removed them. Both are fixed by the same change, as the issue asked us to check.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/cloud/cloudstore/cloudstore.go` | `materializedChunkMutations` now materializes non-relation deletes from `chunk.Mutations`; new `materializableChunkMutation` helper states the rule (relations always; session/observation/prompt only for `delete`) |
| `internal/cloud/cloudstore/cloudstore_test.go` | Unit coverage: prompt delete and observation hard delete survive materialization; soft-delete tombstone is ordered after the typed upsert; upserts still materialize exactly once |
| `internal/cloud/cloudstore/dashboard_queries_test.go` | Dashboard read-model regressions for deleted prompt / hard-deleted observation / soft-deleted observation, driven through the same materialization the cloud performs |
| `internal/cloud/cloudstore/prompt_delete_propagation_test.go` | End-to-end propagation test: real local store + real cloud exporter → chunk payloads → cloud materialization → dashboard queries |

## 🧪 Test Plan

Written test-first. The reproduction from the issue — a chunk carrying one prompt `delete` mutation plus one session — failed with exactly one entry (the session upsert) and zero prompt entries before the fix.

End-to-end verification (`TestPromptDeletePropagatesFromLocalStoreToDashboard`): save a prompt → cloud export → assert it is listed on the dashboard → `DeletePrompt` → cloud export → assert it is gone. It runs the real SQLite store and the real cloud exporter; only the network transport is replaced, and the cloud side is driven through the same `materializedChunkMutations` + dashboard read model a Postgres deployment uses, because the DB-backed tests need `CLOUDSTORE_TEST_DSN` and no Postgres was reachable here. Both propagation tests were confirmed to fail with the fix reverted and pass with it applied.

9 new tests, all green; full `go test ./...` and `go test -tags e2e ./internal/server/...` are green.

---

## ✅ Contributor Checklist

- [x] I linked an approved issue above (`Closes #837`)
- [x] I ran e2e tests locally: `go test -tags e2e ./internal/server/...`

---

## 💬 Notes for Reviewers

The soft-deleted-observation resurrection is the same mechanism, not a separate bug: the tombstone was dropped from `cloud_mutations`, and the typed `upsert` that replaced it carries `deleted_at`, which `dashboardObservationMutationPayload` does not decode. Rather than teaching the dashboard payload about `deleted_at`, the fix keeps the tombstone in the journal where the rest of the delete semantics already live — so it is folded into this PR instead of a follow-up.

Session deletes travel the identical route (`DeleteSession` hard-deletes too) and are covered by the same rule.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed synchronization of deleted prompts, observations, and sessions so deletions are correctly reflected in the dashboard.
  * Prevented soft-deleted records from reappearing during replay.
  * Ensured deleted items no longer appear in listings, detail views, or project statistics.

* **Tests**
  * Added coverage for hard and soft deletion propagation across local storage, synchronization, and dashboard views.

### DIFF
--- a/internal/cloud/cloudstore/cloudstore.go
+++ b/internal/cloud/cloudstore/cloudstore.go
@@ -396,32 +396,56 @@ func materializedChunkMutations(project string, chunk engramsync.ChunkData) ([]M
 
 	// Relations travel only as relation-entity entries in chunk.Mutations — there is no
 	// typed collection for them like Sessions/Observations/Prompts — so materialize them
-	// here to mirror the mutation-push path (#379). Session/observation/prompt mutations
-	// are already materialized from the typed collections above, so they are skipped to
-	// avoid duplicate cloud_mutations rows.
+	// here to mirror the mutation-push path (#379).
+	//
+	// Session/observation/prompt UPSERTS are already materialized from the typed
+	// collections above, so they are skipped to avoid duplicate cloud_mutations rows.
+	// Their DELETES are not, and must be materialized here (#837): a hard-deleted row is
+	// gone locally, so it cannot ride in a typed collection at all, and a soft-deleted one
+	// only rides as an upsert carrying deleted_at, which a cloud_mutations replay reads as
+	// still present. Appending the delete after the typed entries gives it the higher seq,
+	// so the tombstone wins on replay.
 	for i, mutation := range chunk.Mutations {
-		if strings.TrimSpace(mutation.Entity) != store.SyncEntityRelation {
+		entity := strings.TrimSpace(mutation.Entity)
+		op := strings.TrimSpace(mutation.Op)
+		if op == "" {
+			op = store.SyncOpUpsert
+		}
+		if !materializableChunkMutation(entity, op) {
 			continue
 		}
 		entityKey := strings.TrimSpace(mutation.EntityKey)
 		if entityKey == "" {
-			return nil, fmt.Errorf("cloudstore: materialize chunk: mutations[%d].entity_key is required for relation", i)
-		}
-		op := strings.TrimSpace(mutation.Op)
-		if op == "" {
-			op = store.SyncOpUpsert
+			return nil, fmt.Errorf("cloudstore: materialize chunk: mutations[%d].entity_key is required for %s", i, entity)
 		}
 		payload := json.RawMessage(strings.TrimSpace(mutation.Payload))
 		if len(payload) == 0 {
 			payload = json.RawMessage("{}")
 		}
-		if field, ok := chunkcodec.ValidateRelationPayload(payload); !ok {
-			return nil, fmt.Errorf("cloudstore: materialize chunk: mutations[%d].payload.%s is required for relation", i, field)
+		if entity == store.SyncEntityRelation {
+			if field, ok := chunkcodec.ValidateRelationPayload(payload); !ok {
+				return nil, fmt.Errorf("cloudstore: materialize chunk: mutations[%d].payload.%s is required for relation", i, field)
+			}
 		}
-		entries = append(entries, MutationEntry{Project: project, Entity: store.SyncEntityRelation, EntityKey: entityKey, Op: op, Payload: payload})
+		entries = append(entries, MutationEntry{Project: project, Entity: entity, EntityKey: entityKey, Op: op, Payload: payload})
 	}
 
 	return entries, nil
+}
+
+// materializableChunkMutation reports whether a chunk.Mutations entry still needs a
+// cloud_mutations row after the typed collections have been materialized.
+func materializableChunkMutation(entity, op string) bool {
+	switch entity {
+	case store.SyncEntityRelation:
+		// Relations have no typed collection, so every relation entry materializes.
+		return true
+	case store.SyncEntitySession, store.SyncEntityObservation, store.SyncEntityPrompt:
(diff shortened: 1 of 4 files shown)


Alan-TheGentleman:
Merged. Notes for the record.

The fix is keyed on the operation rather than the entity type, which is the right shape: upserts keep coming from the typed collections so the no-duplicate `cloud_mutations` invariant the skip protects stays intact, and only tombstones get materialized. One rule covering every entity that can be deleted, instead of a prompt-shaped special case. Tombstone ordering matters and is handled — deletes append after the typed entries so they carry the higher seq and survive the `cloud_mutations` replay.

Two things this turned up beyond the reported bug. Hard-deleted observations lose their delete through the identical route, same root cause, fixed here. And soft-deleted observations were being resurrected on the dashboard: the row survives with `deleted_at` so it materializes an upsert, but the dashboard payload never decodes that field, so the chunk deletes it and the replay puts it back. Folding that in was better than teaching the dashboard payload about `deleted_at`, which would have papered over a missing journal entry.

The verification is honest about its limits: no Postgres was reachable, so instead of claiming an end-to-end cloud test, the chunks go through the same `materializedChunkMutations` and dashboard read model a Postgres deployment uses. Both propagation tests were confirmed failing with the fix reverted, so they are real regressions rather than tests that pass either way.

PR #845, opened 2026-08-27
fix(sync): key relation dead-letter rows to the mutation

### LINKED ISSUE #838: Failed relation mutations with a blank entity_key overwrite each other in sync_apply_deferred
## 📝 Bug Description

Two distinct failed relation mutations can collapse into a single `sync_apply_deferred` row, silently destroying the evidence for one of them. The surviving row can also become permanently unreachable by the success-path cleanup, leaving an orphan.

Found while auditing the dead-letter identity scheme for #798. It is **pre-existing on `main`** (introduced by `292bf23`, well before that PR) and lives on the relation path, not the session path, so it is filed separately.

## Mechanism

`recordRelationApplyFailureTx` (`internal/store/store.go` around line 4642) uses the mutation's `entity_key` **directly** as the row's `sync_id`:

```go
`, mutation.EntityKey, mutation.Entity, mutation.Payload, targetKey, project, scopeClass, status)
```

with `ON CONFLICT(sync_id) DO UPDATE`. Nothing guarantees `entity_key` is non-blank or unique, so two different relation mutations that both carry a blank `entity_key` hash to the same row and the second overwrites the first.

Observed on unmodified `main`, feeding two distinct relation mutations (`rel-a`, `rel-b`) with blank entity keys:

```
rows=1
  sync_id="" status=dead payload={"sync_id":"rel-b",…}   ← rel-a silently overwritten
```

Compounding it: the success-path cleanup deletes by the payload's own sync id (`DELETE … WHERE sync_id = p.SyncID`), which no longer matches the row's blank `sync_id`. So once a relation later applies successfully, its deferred row cannot be cleaned up and is orphaned.

## Why this is not a small fix

`sync_id` is load-bearing for the relation retry contract in the same table: `ReplayDeferredForScope` reconstructs it as `EntityKey`, and the success path deletes by `p.SyncID`. Changing the derivation touches that contract on both the write and the replay side, so it wants a deliberate design pass rather than a patch — which is why it is a separate issue instead of being folded into #798.

For contrast, #798's session dead-letter path now derives identity from the mutation's own distinguishing material (target key, entity, entity key, op and payload, each length-prefixed), which keeps distinct mutations apart by construction while collapsing genuine redelivery onto one row. Something along those lines is the likely shape here, but only if the replay and cleanup sides move with it.

## 🔄 Steps to Reproduce

1. Apply two distinct relation mutations that both fail (for example, both referencing observations that do not exist locally) and that both carry a blank `entity_key`.
2. Query `sync_apply_deferred`.
3. One row exists, holding only the second payload. The first mutation's evidence is gone.

## ✅ Expected Behavior

Every distinct failed relation mutation keeps its own deferred row, and a genuine redelivery of the same mutation collapses onto one row rather than accumulating. A row created by a failure is reachable by the success-path cleanup once that mutation applies.

## ❌ Actual Behavior

Distinct mutations sharing a blank `entity_key` overwrite each other, and the surviving row can be orphaned because the cleanup looks it up by a different identifier than the one it was written under.

## Additional Context

Incidental, unreviewed observation from the same audit, worth a glance but not part of this report: `chunkcodec.ChunkID` truncates SHA-256 to 8 hex characters (32 bits).

Alan-TheGentleman:
## 🔗 Linked Issue

Closes #838

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix

---

## 📝 Summary

- Distinct failed relation mutations no longer overwrite each other's evidence in `sync_apply_deferred`. A `dead` row is now keyed on the discarded mutation's own material — target key, entity, entity key, op and payload, each length-prefixed — the same way `pulledSessionDeadLetterSyncID` already keys discarded session mutations. A genuine redelivery of the same mutation still lands on the one row.
- A `deferred` row keeps the relation's own `sync_id`. `applyRelationUpsertTx` validates the wire contract before it ever reports a missing endpoint, so a retryable mutation is proven to carry a non-blank `entity_key` equal to its payload `sync_id` — and that is the key the whole retry contract is written in. The two statuses record different things (retry state vs. evidence), so they carry different identities; only evidence rows change.
- The success-path cleanup no longer orphans rows. It now deletes a relation's rows under all three ways they can be keyed: the row key (rows written before `entity_key` was stored), the `entity_key` column, and the relation identity inside the payload — the only handle on a row whose mutation carried a blank or mismatched entity key.
- Replay reconstructs the mutation from the row's own `entity_key` and `op`, falling back to the row key and `upsert` for rows written before those columns were populated.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | Added `relationApplyFailureSyncID`; `recordRelationApplyFailureTx` writes the derived key plus `entity_key`/`op` and retires the legacy row for the same mutation; `applyRelationUpsertTx` cleanup matches row key, `entity_key` and payload identity; `ReplayDeferredForScope` reads `entity_key`/`op` with a legacy fallback |
| `internal/store/relation_deferred_identity_test.go` | New: collision, shared-entity-key collision, redelivery idempotency, orphan cleanup, FK-miss replay end to end, and four old-scheme row cases |
| `internal/store/sync_apply_test.go` | Existing dead-row lookups resolve the key through a `deadRelationRowKey` helper instead of restating the derivation |
| `cmd/engram/conflicts.go` | `engram conflicts deferred` prints the mutation's `entity_key` (and `op` on `--inspect`), since a derived key does not name the mutation it holds |
| `cmd/engram/conflicts_test.go` | New: the inspect surface names the mutation behind a derived key |

## 🧪 Test Plan

**Written test-first.** On unmodified `main` the new tests fail exactly as the issue reports:

```
--- FAIL: TestRelationDeadLetter_DistinctBlankEntityKeysKeepSeparateRows
    deferred payload identities: want [rel-a rel-b], got [rel-b]
--- FAIL: TestRelationDeadLetter_SuccessfulApplyCleansUpBlankKeyedRow
    orphaned deferred rows after successful apply: want 0, got 1
--- FAIL: TestRelationDeadLetter_LegacyRowForOtherMutationIsPreserved
    deferred payload identities: want [rel-a rel-b], got [rel-a]
```

**Empirical run**, a built harness feeding two distinct relation mutations with blank entity keys through `ApplyPulledChunk`. Before:

```
--- after two distinct failed relation mutations with blank entity_key ---
  sync_id="" status=dead payload.sync_id="rel-b"          ← rel-a erased
  rows=1
--- after rel-a applies successfully (cleanup) ---
  sync_id="" status=dead payload.sync_id="rel-b"          ← orphan
  rows=1
```

After:

```
--- after two distinct failed relation mutations with blank entity_key ---
  sync_id="relation-dead-364ceb83…" status=dead payload.sync_id="rel-a"
  sync_id="relation-dead-cee507c6…" status=dead payload.sync_id="rel-b"
  rows=2
--- after redelivering the same two mutations ---
  rows=2                                                   ← redelivery collapses
--- after rel-a applies successfully (cleanup) ---
  sync_id="relation-dead-cee507c6…" status=dead payload.sync_id="rel-b"
  rows=1                                                   ← rel-a's row cleaned, rel-b's kept
```

**Upgrade path**, a database seeded by the old binary and then opened by the new one:

```
--- database as written by the old code ---
  sync_id="" status=dead entity_key="" payload.sync_id="rel-b"
  sync_id="rel-pending" status=deferred entity_key="" payload.sync_id="rel-pending"
--- same database opened by the new code ---
  (identical; ListDeferred reachable rows: 2)
replay: retried=1 succeeded=1 failed=0 dead=0
--- after replaying the legacy pending row ---
  sync_id="" status=dead entity_key="" payload.sync_id="rel-b"
```

---

## 💬 Notes for Reviewers

**Why the identity depends on the status.** Following #798 literally and hashing every relation failure would have re-keyed the entire live retry population, so an existing deferred row and a redelivery of its mutation would sit in the table as two rows for the same relation until one of them applied. Splitting on status avoids that without leaving any collision class open: retry state is keyed by the relation identity the contract already uses and is proven valid before the row is written, and everything unproven — blank entity key, entity key disagreeing with the payload, undecodable payload, unsupported op — is evidence and gets the derived key.

**Existing rows.** No schema change and no rewrite on open. Deferred rows keep their identity exactly, so replay and cleanup are untouched for them. Legacy dead rows stay listable through `ListDeferred`/`GetDeferred` and `engram conflicts deferred`, and the cleanup now reaches them through their payload identity. Redelivering a mutation whose old row is still present retires that row rather than duplicating it — guarded on identical payload and `apply_status='dead'`, so a row that collapsed *another* mutation's evidence is never destroyed by an unrelated redelivery, and pending retry state is structurally out of reach. `TestRelationDeadLetter_RedeliveryRetiresLegacyRowForSameMutation`, `TestRelationDeadLetter_LegacyRowForOtherMutationIsPreserved`, `TestReplayDeferred_LegacyRowWithoutEntityKeyStillApplies` and `TestGetDeferred_LegacyRelationRowStaysReachable` cover these.

**Evidence that was already lost** to a pre-upgrade collapse cannot be recovered — the overwritten payload is gone from the database. The surviving row is preserved as-is.

The `chunkcodec.ChunkID` note at the end of the issue is out of scope here and untouched.

## Summary by CodeRabbit

- **New Features**
  - Deferred-conflict inspection now displays each mutation’s entity key and operation.
  - Default deferred-conflict listings now include entity keys.

- **Bug Fixes**
  - Prevented distinct failed relation mutations from overwriting one another.
  - Improved cleanup, retry, and replay behavior for deferred mutations, including legacy records.
  - Preserved separate deferred records when mutations share or lack entity keys.

- **Tests**
  - Added coverage for deduplication, retries, cleanup, replay, migration, and legacy deferred records.


Alan-TheGentleman:
Merged after four review rounds, each of which found something real. Recording them because the later ones were introduced by the fix itself.

**Round 1 — the reported collision.** Solved by splitting identity on apply status rather than hashing everything: `deferred` rows keep the relation's `sync_id`, `dead` rows get a content-derived key. The reasoning that makes this sufficient is that the wire contract is validated before the FK check, so every retryable row provably carries the relation's own id — hashing the whole population would have re-keyed live retry state instead.

**Round 2 — the fix reopened the bug it closed.** The widened cleanup matched on `entity_key` with no scoping, and the same change had started persisting the discarded mutation's raw `entity_key` onto dead rows — whose `entity_key` may disagree with their payload, which is *why* they are dead. So a valid apply of relation X erased evidence about an unrelated mutation that happened to carry key X. Now the cleanup never consults `entity_key` at all: a row belongs to the applied relation only when the relation it *describes* is that one.

**Round 2b — the cleanup became a full scan on the hot path.** It had grown a `json_extract` term, not index-satisfiable, running inside the apply write transaction on every successful upsert. Fixed by persisting the payload identity in an indexed column. Measured against a 20k-row backlog: 24.1 ms/op → 8.7 µs/op, linear to flat. The plan assertion EXPLAINs the actual SQL constant production uses, so it cannot drift.

**Round 3 — the migration could skip its backfill forever.** The `ALTER` auto-commits separately from the backfill `UPDATE`; one failure between them left the column present and every legacy row blank, and the next open saw `added=false` and skipped it permanently. Blank meant "known only by its key", which re-enabled the round-2 evidence loss. Now unconditional and idempotent, which is also the repo's existing idiom. Worth noting a version marker would *not* have fixed it: an older binary opening the same database after the column exists writes rows with the `DEFAULT ''`, and the marker would already be advanced.

**Round 4 — filed as #849, not fixed here.** Keying dead rows by payload digest means logically identical failing mutations that differ by a byte no longer collapse, and nothing reaps them. That is the deliberate cost of not collapsing distinct evidence, and the unreachable-row case the reviewer names is correct behavior — such a row is evidence, and no unrelated apply should erase it. What is missing is a retention policy, which is a product decision rather than a mechanical fix. The session dead-letter path from #798 shares the characteristic and should be covered by the same policy.

PR #846, opened 2026-08-27
fix(doctor): scan nullable session project on legacy databases

### LINKED ISSUE #841: engram doctor dies on legacy databases with NULL session projects, exactly where it is needed most
## 📝 Bug Description

`engram doctor` fails outright on any database that still holds legacy sessions with a NULL `project`, which is exactly the population that most needs to run it:

```
sql: Scan error on column index 1, name "project": converting NULL to string is unsupported
```

This matters more than it would on its own because #813's new ownership errors point users toward diagnosis on precisely these databases. A user hitting `project_ownership_required` is told to inspect and repair, and `doctor` is the tool that cannot run there.

## Mechanism

`ListDiagnosticSessions` (`internal/store/diagnostic.go:92`) selects `project` raw and scans it into `DiagnosticSessionEvidence.Project`, which is a plain `string` (`diagnostic.go:18`):

```go
query := `SELECT id, project, ifnull(directory, ''), id FROM sessions`
```

Note the query already guards `directory` with `ifnull()` but not `project`, which reads like an oversight rather than a deliberate choice.

`ListInvalidSessionIdentityEvidence` (`internal/store/diagnostic.go:135`) has the same latent shape:

```go
query := `SELECT id, project FROM sessions WHERE ` + sqlSessionIDBlank("id")
```

The migration does not rewrite `sessions.project` to NOT NULL, so rows from the era when that column was nullable survive an upgrade intact.

## 🔄 Steps to Reproduce

1. Take a database carrying legacy rows where `sessions.project IS NULL` (or seed one against the older nullable schema).
2. Run `engram doctor`.
3. It dies with the scan error above instead of reporting on the store.

Reproduced against a binary built from pristine `main` at `de69c288`, so this is not introduced by any of the recently landed PRs — the code is identical in the merge base.

## ✅ Expected Behavior

`doctor` runs on a legacy database and reports the unowned sessions as findings, since surfacing exactly this kind of legacy state is what the tool is for.

## ❌ Actual Behavior

`doctor` aborts with a scan error and produces no report at all, leaving the user with no diagnostic path on the databases that need one most.

## Suggested Fix

Mirror what #813 already did for `GetSession`: scan the nullable column into a `sql.NullString` (or add `ifnull(project, '')` to both queries, consistent with how `directory` is already handled). Worth auditing the other diagnostic queries for the same pattern while in there, and adding a fixture built on the legacy nullable schema so this class of regression is caught.

Alan-TheGentleman:
## 🔗 Linked Issue

Closes #841

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix

---

## 📝 Summary

- `engram doctor` no longer dies with `sql: Scan error on column index 1, name "project": converting NULL to string is unsupported` on a database that still holds sessions with a NULL project — the population #813's `project_ownership_required` remedy sends to diagnosis, and the one place diagnosis could not run.
- Fixed the same raw scan of the same column in **six more places** the audit turned up: `RecentSessions`, `AllSessions`, `Export`, `DeleteSession`, `EndSession`, and the `CreateSession` readback. A single legacy row was enough to deny the user `engram context`, `mem_context`, and the TUI session list, and to make that session impossible to delete, end, or re-register.
- Added the `unowned_session_project` doctor check, because surviving the read is only half the ask: doctor now **reports** each session that identifies no project and hands back the exact `engram projects rescue-ownership --project <name> --session <id>` repair.

### Why `ifnull()` and not `sql.NullString`

The query that crashed already guarded its sibling `directory` with `ifnull(directory, '')`, and three other queries in `internal/store/diagnostic.go` already do the same for `project`. Keeping the SQL guard leaves `DiagnosticSessionEvidence.Project` and `InvalidSessionIdentityEvidence.Project` as plain strings, so the JSON envelope doctor emits does not change shape. NULL and blank both mean "identifies no project" to every caller involved, exactly as `sessionOwnershipTx` already treats them, so collapsing them loses nothing. `GetSession` uses `sql.NullString` because it is a single-row scan with no `ifnull` in its query; both idioms are in-tree and each fits its own site.

### Root cause, and why an upgrade does not fix it

`sessions.project` is the only column in the schema that is nullable in practice while being declared `NOT NULL`. The table is only ever created with `CREATE TABLE IF NOT EXISTS` (`store.go:782`), which is a no-op on an existing table, so the `NOT NULL` declaration never reaches a row written before it. There is no `addColumnIfNotExists`, no table rebuild, and no backfill `UPDATE` for `sessions` — contrast `migrate()`, which does backfill `user_prompts.project` and `observations.scope`. `createSessionTx` cannot heal such a row on the way through either: its `ON CONFLICT … WHEN sessions.project = ''` never matches, because `NULL = ''` is NULL, not true.

### Audit scope

All 177 SQL scan destinations across `internal/store`, `internal/cloud`, and the dashboard were checked against the schema and its historical snapshots. `sessions.project` is the **only** legacy-nullable-into-non-nullable column. Everything else is already correct: nullable columns are `ifnull`/`coalesce`-wrapped, scanned into `*string`/`sql.NullString`, structurally excluded by a `project = ?` or `IS NOT NULL` filter, or were added via `ALTER TABLE ADD COLUMN … NOT NULL DEFAULT`, which backfills existing rows.

Two things the audit surfaced that are **not** in this PR, since they are a different problem rather than the same root cause:

1. `sessions.directory` is `NOT NULL` in the current schema and in all three historical DDL snapshots, yet three queries defensively wrap it in `ifnull(directory, '')`. I could not establish from the repo whether legacy databases actually carry NULL directories, so I left it alone rather than guessing.
2. `createSessionTx`'s `NULL = ''` comparison means a legacy NULL is never healed by a re-registration, only by explicit rescue. That is repair semantics, not a scan crash.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/diagnostic.go` | `ListDiagnosticSessions` and `ListInvalidSessionIdentityEvidence` read `project` through `ifnull(project, '')` — the bug in the issue |
| `internal/store/store.go` | Same guard at the six other raw scan sites: `CreateSession` readback, `EndSession`, `RecentSessions`, `AllSessions`, `DeleteSession`, `exportWithProjectScope` |
| `internal/diagnostic/checks.go` | New `UnownedSessionProjectCheck` reporting each session that identifies no project, with the `rescue-ownership` repair; listing deliberately unscoped |
| `internal/diagnostic/registry.go` | Register `unowned_session_project` |
| `internal/store/legacy_null_project_test.go` | **New.** Single regression fixture for the class: eight tests, each driving one caller against a seeded legacy NULL row |
| `internal/diagnostic/diagnostic_test.go` | Legacy-nullable store fixture; doctor-runs-on-legacy-DB test; three tests for the new check; MVP-count assertion now derives from `RegisteredCodes()` so it cannot rot again |
| `internal/store/store_test.go` | Coverage test's query-substring matcher updated for the rewritten export query |
| `docs/DOCTOR.md`, `CHANGELOG.md` | Document the new check and both fixes |

## 🧪 Test Plan

**Failing first (TDD).** All eight regression tests were written before the fix and reproduced the reported error verbatim:

```
--- FAIL: TestListDiagnosticSessionsReadsLegacyNullProject
    ListDiagnosticSessions on legacy NULL project = sql: Scan error on column
    index 1, name "project": converting NULL to string is unsupported
--- FAIL: TestRecentSessionsReadsLegacyNullProject
--- FAIL: TestAllSessionsReadsLegacyNullProject
--- FAIL: TestExportReadsLegacyNullProject
--- FAIL: TestDeleteSessionReadsLegacyNullProject
--- FAIL: TestEndSessionReadsLegacyNullProject
--- FAIL: TestCreateSessionReadbackReadsLegacyNullProject
```

**Manual verification with a built binary**, against a database seeded through raw SQL on the nullable schema (the public API normalizes, so it cannot produce one).

Before, on a binary built from the merge base:

```
$ engram doctor
engram doctor failed: sql: Scan error on column index 1, name "project":
converting NULL to string is unsupported
```

After:

```
$ engram doctor
Engram Doctor: warning
Checks: 6 ok=5 warnings=1 blocked=0 errors=0
...
[warning] unowned_session_project — 2 finding(s) detected.
  why: A session left over from the schema where sessions.project was nullable
       owns no project, so its memories stay outside every project-scoped
       retrieval and writes to it must resolve ownership before they can land.
  next: Assign ownership with `engram projects rescue-ownership --project <name>
        --session legacy-null-a` after confirming which project the session
        belongs to.
  - unowned_session_project: Session "legacy-null-a" identifies no project.
    evidence: {"directory":"/home/user/work/alpha","session_id":"legacy-null-a","session_project":""}
  - unowned_session_project: Session "legacy-null-b" identifies no project.
    evidence: {"directory":"/home/user/work/beta","session_id":"legacy-null-b","session_project":""}
```

The full referral loop now closes on the same database — `engram context` and `engram export` read it, doctor names the rows, the repair it prints works, and a re-run confirms the change:

```
$ engram projects rescue-ownership --project alpha --session legacy-null-a
Rescued ownership into "alpha": 1 sessions, 0 observations, 0 prompts

$ engram doctor
[warning] unowned_session_project — 1 finding(s) detected.
  - unowned_session_project: Session "legacy-null-b" identifies no project.
```

`gofmt -l` is clean on every file touched.

---

## 💬 Notes for Reviewers

- The new check is a `warning`, not `blocking`. An unowned session is a legacy state to resolve deliberately, not something that should fail a health check; each finding sets `requires_confirmation=true` because assigning ownership moves data.
- `engram context` renders an unowned session with an empty project label (`**** `). That is cosmetic and pre-existing in shape — before this PR it was a hard crash — so I left presentation alone rather than widening the diff.
- Unrelated to this change, `engram doctor` exits `0` even when it fails to produce a report. Worth a separate issue if you want CI to be able to depend on it.

## Summary by CodeRabbit

- **New Features**
  - Added a diagnostic check that identifies sessions without an assigned project and provides a repair command.
  - Diagnostic reports now include unowned sessions without project filtering.

- **Bug Fixes**
  - Prevented crashes when reading databases upgraded from schemas allowing sessions without projects.
  - Session listing, exporting, deletion, ending, and creation operations now handle missing project assignments safely.

- **Documentation**
  - Documented the new diagnostic check and ownership-repair guidance.


Alan-TheGentleman:
Merged. Notes for the record.

This turned out to be much larger than the reported one-line scan. The audit found six more sites with the identical raw scan of `sessions.project`: `RecentSessions`, `AllSessions`, `Export`, `DeleteSession`, `EndSession`, and the `CreateSession` readback. So a single legacy row did not just break `doctor` — it denied the user `engram context`, `mem_context`, the TUI session list and `export`, and made that session impossible to delete or end. `EndSession` failed *after* its UPDATE, rolling the whole thing back.

Root cause is worth writing down: `sessions` is only ever created with `CREATE TABLE IF NOT EXISTS`, which is a no-op on an existing table, so the `NOT NULL` declaration never reaches older rows and there is no backfill for it the way there is for `user_prompts.project`. That makes `sessions.project` the only column in the schema nullable-in-practice while declared `NOT NULL` — confirmed by checking all 177 scan destinations across store, cloud and dashboard.

`ifnull(project, '')` over `sql.NullString` is the right call here: the crashing query already guarded its sibling `directory` that way, three other queries in the same file already do it for `project`, and it keeps the doctor JSON envelope shape unchanged. `GetSession` uses `sql.NullString` because it is a single-row scan with no `ifnull` — both idioms are in-tree and each fits its site.

The new `unowned_session_project` check closes the referral loop that made this urgent: doctor now names the unowned sessions and prints a repair command that actually works on the same database, which is where #813's `project_ownership_required` remedy was sending people.

Three things noted in the PR and deliberately not fixed here: whether legacy databases really carry NULL `sessions.directory` (three queries defensively guard a column that is `NOT NULL` in every historical DDL snapshot — I could not establish it either way), `createSessionTx`'s `WHEN sessions.project = ''` never healing a NULL since `NULL = ''` is NULL, and `engram doctor` exiting 0 even when it fails to produce a report.

PR #883, opened 2026-08-30
feat(store,mcp): add ranged and find-window reads to mem_get_observation

### LINKED ISSUE #812: feat(store,mcp): add partial reads to mem_get_observation (offset/limit + find/context)
### 📋 Pre-flight Checks

> Searched: `mem_get_observation offset range`, `search within observation`, `partial update`, `large observation`, `observation length limit`. The closest matches are **#602 / PR #774** (`find`/`replace` on `mem_update`) and **#752** (opt-in search env vars). #602 addresses the **write** side of the same cost; this issue addresses the **read** side, and the two compose. Filing separately rather than as a comment on #602 so it can be scoped, approved and reverted independently.

### 🔍 Problem Description

`mem_get_observation` returns the whole observation or nothing — the only parameter is `id` (`internal/mcp/mcp.go:613-623`). For small memories that is exactly right. For a large one, an agent that needs a single section still pays for the entire body.

This bites hardest in the workflow the tool is built for: an agent that edits a stored document repeatedly.

Real case, from an SDD design cycle where four design artifacts were kept in engram:

| artifact | chars |
|---|---|
| design (Part A) | 47,051 |
| design (Part B) | 45,688 |
| design (Part C) | 23,446 |
| dispositions | 16,833 |
| **total** | **133,018** |

One of those observations carries `revision_count = 7` — it was rewritten seven times across review rounds. Every round, an agent had to pull all four artifacts in full (**~33,000 tokens of input**) in order to correct a few dozen paragraphs. On disk the same job is `rg` for the section, then read ~2,000 chars around it.

The cost is not the editing. It is that **locating** the passage requires downloading everything around it.

Two agent runs were lost to session limits mid-task before the artifacts were moved out of engram to plain files — which also gave up the recall, `topic_key` upsert and `memory_relations` conflict detection that were the reason to use engram in the first place. That trade should not be necessary.

**#602 fixes half of this.** With `find`/`replace`, writing a correction becomes cheap. Reading stays at ~33,000 tokens, because the agent still has to see the text before it can know what to replace. Ranged reads are the other half; together they make engram viable for documents that are revised, not just facts that are recalled.

### 💡 Proposed Solution

Add optional parameters to `mem_get_observation`. Two independent mechanisms — either can ship alone:

**1. Mechanical paging — `offset` + `limit`**

```
mem_get_observation(id: 475, offset: 12000, limit: 2000)
```

Returns runes `[offset, offset+limit)` of `content`. Mirrors how every file-read tool works.

**2. Match-scoped read — `find` + `context`**

```
mem_get_observation(id: 475, find: "test 7", context: 600)
```

Returns each window of `context` runes around every occurrence of `find`, with its rune offset, so the agent can locate and read in **one** round-trip instead of guessing an offset. This is the mode agents actually need: you rarely know an offset, you know the text you are looking for.

The returned offsets compose directly with #602 — locate with `find`, patch with `mem_update(find:…, replace:…)`.

**Response shape** (mode 2), each window prefixed so the agent knows what it did *not* get:

```
#475 "design Part A" — 47,051 runes total, 3 matches for "test 7"

[offset 21,340]
…text…

[offset 28,902]
…text…
```

**Validation rules:**
- `offset`/`limit` and `find`/`context` are mutually exclusive → error if both groups are given
- `context` without `find` → error: "context requires find"
- `limit` without `offset` → treated as `offset: 0`
- `offset` beyond the end → empty result, no error
- `find` not found → zero windows, no error (mirrors #602's no-op-on-miss rule)
- no parameters → current behaviour, unchanged

**Defaults:** `limit` and `context` default to a bounded value rather than unlimited, so a partial read cannot accidentally cost more than the full read.

**⚠ Offsets must be counted in runes, not bytes.** #697, #751 and #809 are all live reports of byte/rune confusion in the truncation and preview paths. A byte-indexed `offset` would split multi-byte characters and reproduce that same class of bug in a new surface. `[]rune` conversion, or `utf8.RuneCountInString` plus a rune-aware slice, keeps this consistent with how those issues are being resolved.

### 📦 Affected Area

- MCP Server (tools, transport)
- Store (database, queries)

### 🔄 Alternatives Considered

**Comment on #602 instead of filing separately.** Rejected: #602 is already approved with an open PR (#774). Adding read-side scope now would widen an approved issue after the fact and complicate a PR that is otherwise passing its checks.

**Use `mem_search` to locate, then `mem_get_observation`.** Does not work for this. `mem_search` ranks *across* observations and returns a 300-char preview (`internal/mcp/mcp.go:1063`); it cannot enumerate matches *within* one body or report where they are. There is no "grep inside this observation" today.

**Split large documents into many small observations.** This is what the truncation warning already advises, and it is right for facts. It fails for documents: a design document split across four observations still has to be read in full to be revised, and the split itself caused a real defect in our case — the same content ended up duplicated across two artifacts instead of moved, so a single edit then needed two updates that could silently drift apart.

**Raise `MaxObservationLength`.** Orthogonal. A bigger cap makes the read cost worse, not better.

**Keep documents on disk, keep facts in engram.** The current workaround, and what we did. It works, but it forfeits recall, `topic_key` upsert and `memory_relations` for exactly the artifacts that most need conflict detection.

**SQL-level `substr()`.** Viable for mode 1 and avoids loading the full body into Go, but SQLite's `substr()` on TEXT counts characters, not runes, and mode 2 needs match positions anyway. Worth considering as an optimisation once behaviour is settled, not as the initial shape.

### 📎 Additional Context

**Token math** at ~4 chars/token, for the real case above — one review round that corrected passages across all four artifacts:

| | read cost | write cost (with #602) |
|---|---|---|
| today | ~33,000 tok | ~cheap |
| with match-scoped reads | ~1,500 tok | ~cheap |

The read saving is roughly **95%**, and it scales with document size while the write saving does not — the larger the observation, the more of the remaining cost is reading.

**Implementation reference files:**
- Tool registration: `internal/mcp/mcp.go:613-623` (currently `WithNumber("id")` only)
- Handler: `internal/mcp/mcp.go:1821` — `handleGetObservation`
- Store read: `internal/store/store.go:3003` — `GetObservation(id int64)`
- Rune-safety precedent: `internal/store/store.go:2357-2358`, and the open reports #697 / #751 / #809

**Relationship to #602:** complementary, not overlapping. #602 makes an edit cheap once you know what to change. This makes finding out what to change cheap. Neither depends on the other, and shipping both is what removes the reason to keep large documents outside engram.

astraltrekkin:
## 🔗 Linked Issue

Closes #812

---

## 🏷️ PR Type

- [ ] `type:bug` — Bug fix

---

## 📝 Summary

- Add optional `offset`/`limit` and `find`/`context` arguments to `mem_get_observation` so an agent can read a rune window instead of the full observation body.
- Keep the existing id-only call unchanged: it still returns the complete untruncated content.
- Count every offset and length in runes so multi-byte characters are not split.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/observation_partial.go` | Rune-indexed range and find-window contract, including validation and defaults (limit 2000, context 600 per side). |
| `internal/store/observation_partial_test.go` | Store coverage for both modes, rune cuts, defaults, and error cases. |
| `internal/store/store.go` | Sentinel errors for exclusive modes, missing find, and negative values. |
| `internal/mcp/mcp.go` | Tool schema, request parsing, and formatted responses for range and find windows. |
| `internal/mcp/observation_partial_test.go` | Handler coverage for unchanged full reads, range, find, and validation errors. |
| `DOCS.md` | Document the new `mem_get_observation` parameters and rules. |
| `docs/ARCHITECTURE.md` | Update the tool table and progressive-disclosure example. |

## 🧪 Test Plan

- [ ] Unit tests pass locally: `go test ./...`
- [ ] E2E tests pass locally: `go test -tags e2e ./internal/server/...`

Focused handler and store cases:

```
go test ./internal/store ./internal/mcp -count=1 -timeout 180s -run 'TestResolveObservationRead|TestHandleGetObservation'
ok  	github.com/Gentleman-Programming/engram/internal/store	0.396s
ok  	github.com/Gentleman-Programming/engram/internal/mcp	1.106s
```

Package suites:

```
go test ./internal/store ./internal/mcp ./internal/server -count=1 -timeout 8m
ok  	github.com/Gentleman-Programming/engram/internal/store	11.885s
ok  	github.com/Gentleman-Programming/engram/internal/mcp	12.219s
ok  	github.com/Gentleman-Programming/engram/internal/server	2.205s
```

Manual checks exercised through those tests: id-only full body; `offset`/`limit` slice; `limit` without `offset` starts at 0; offset past the end is empty; `find` windows and zero-match; exclusive-mode and `context requires find` errors; UTF-8 rune slice (`aé😊z` → `é😊`).

---

## 🤖 Automated Checks

| **Check PR Has type:\* Label** | PR has exactly one `type:*` label | ⏳ |
| **Plugin Tests** | `npm test` passes in `plugin/pi` | ⏳ |

---

## ✅ Contributor Checklist

- [x] I linked an approved issue above (`Closes #N`)

---

## Summary by CodeRabbit

* **New Features**
  * Added partial reading for observation content using rune-based offsets and limits.
  * Added literal-match searches with configurable context windows around matches.
  * Added match counts, range details, and rune totals to partial-read results.
  * Full, untruncated content remains available when no partial-read options are provided.
  * Added validation for incompatible options, missing required parameters, negative values, and unsafe numeric inputs.

* **Documentation**
  * Expanded tool documentation with supported options, defaults, behavior, and validation rules.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -931,7 +931,19 @@ The optional project filter is enforced: an observation owned by another project
 
 ### mem_get_observation
 
-Get full untruncated content of a specific observation by ID.
+Get the content of a specific observation by ID.
+
+With no extra parameters this returns the full untruncated body (unchanged). Optional partial-read parameters reduce token cost on large observations. Offsets and lengths are counted in **runes**, not bytes.
+
+**Mechanical paging — `offset` + `limit`**
+
+`mem_get_observation(id: 475, offset: 12000, limit: 2000)` returns runes `[offset, offset+limit)`. `limit` without `offset` starts at `0`. `offset` past the end returns an empty window (no error). Default `limit` is 2000.
+
+**Match-scoped read — `find` + `context`**
+
+`mem_get_observation(id: 475, find: "test 7", context: 600)` returns a window of `context` runes on each side of every literal match, prefixed with the window's rune offset. `find` with no matches returns zero windows (no error). `context` without `find` is an error. Default `context` is 600.
+
+`offset`/`limit` and `find`/`context` are mutually exclusive.
 
 ### mem_session_summary
 
--- a/DOCS.md
+++ b/DOCS.md
@@ -939,9 +939,11 @@ With no extra parameters this returns the full untruncated body (unchanged). Opt
 
 `mem_get_observation(id: 475, offset: 12000, limit: 2000)` returns runes `[offset, offset+limit)`. `limit` without `offset` starts at `0`. `offset` past the end returns an empty window (no error). Default `limit` is 2000.
 
+Optional partial-read numeric values must be safe integers; negative values are rejected.
+
 **Match-scoped read — `find` + `context`**
 
-`mem_get_observation(id: 475, find: "test 7", context: 600)` returns a window of `context` runes on each side of every literal match, prefixed with the window's rune offset. `find` with no matches returns zero windows (no error). `context` without `find` is an error. Default `context` is 600.
+`mem_get_observation(id: 475, find: "test 7", context: 600)` returns merged windows of `context` runes on each side of every literal match, prefixed with each window's rune offset. Overlapping or adjacent contexts are merged; the header still reports literal occurrence count. `find` with no matches returns zero windows (no error). `context` without `find` is an error. Default `context` is 600.
 
 `offset`/`limit` and `find`/`context` are mutually exclusive.
 
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -61,7 +61,7 @@ Next session starts → Previous session context is injected automatically
 | `mem_session_summary` | Save end-of-session summary |
 | `mem_context` | Get recent context from previous sessions |
 | `mem_timeline` | Chronological context around a specific observation |
-| `mem_get_observation` | Get full content of a specific memory |
+| `mem_get_observation` | Get observation content by ID (full body, or optional rune-ranged / find windows) |
 | `mem_save_prompt` | Save a user prompt for future context |
 | `mem_stats` | Memory system statistics |
 | `mem_session_start` | Register a session start |
@@ -83,7 +83,7 @@ Token-efficient memory retrieval — don't dump everything, drill in:
 ```
 1. mem_search "auth middleware"     → compact results with IDs (~100 tokens each)
 2. mem_timeline observation_id=42  → what happened before/after in that session
-3. mem_get_observation id=42       → full untruncated content
+3. mem_get_observation id=42       → full body, or offset/limit / find/context windows
 ```
 
 ---
--- a/internal/mcp/mcp.go
+++ b/internal/mcp/mcp.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -594,7 +595,7 @@ Examples:
 	if shouldRegister("mem_get_observation", allowlist) {
 		srv.AddTool(
 			mcp.NewTool("mem_get_observation",
-				mcp.WithDescription("Get the full content of a specific observation by ID. Use when you need the complete, untruncated content of an observation found via mem_search or mem_timeline."),
+				mcp.WithDescription("Get the content of a specific observation by ID. Omit partial-read params for the full untruncated body. Use offset/limit for a rune-ranged slice, or find/context to return windows around matches. The two groups are mutually exclusive."),
 				mcp.WithTitleAnnotation("Get Observation"),
 				mcp.WithReadOnlyHintAnnotation(true),
 				mcp.WithDestructiveHintAnnotation(false),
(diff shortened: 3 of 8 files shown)


danielgap (commented):
Reviewed against #812 end to end. No blockers from my side; the nits below are non-blocking.

**What I verified:**

- Validation matrix matches the spec exactly: mutual exclusion, context-requires-find, limit-without-offset starts at 0, past-end offset is an empty window with no error, a miss returns zero windows with no error, and no params keeps the full body unchanged.
- Rune correctness: slicing and offsets are rune-indexed everywhere, with multi-byte coverage in both modes. The #697/#751/#809 byte/rune class does not reproduce here.
- The KMP is correct including the subtle parts: overlapping matches are counted (4 for `aa` in `aaaaa`) while their windows merge, adjacent contexts merge, and `matched = prefix[matched-1]` keeps the search alive after a hit.
- Overflow safety on extreme inputs: a huge context clamps in the `contextRunes >= len(runes)-end` branch before any addition, and a huge limit short-circuits in `sliceRunes`; the maxInt tests cover both.
- Full-mode output keeps the exact old shape, and the golden tool contract plus DOCS.md/ARCHITECTURE.md are updated consistently.

**Nits (non-blocking):**

1. `find: ""` currently returns "0 matches" rather than an error. The spec is silent, but an empty needle is almost certainly a caller bug; I would reject it or document the no-op. Either way it deserves a test.
2. Partial modes drop the `[type]` field from the header (`#id "title"` vs full mode's `#id [type] title`). Fine if intentional to keep partial headers short; agents that key on `[type]` lose it exactly when reading large bodies.
3. `limit: 0` silently returns an empty window. Consistent with past-end-offset being silent, but DOCS.md does not mention it.
4. Range mode fills both `Content` and `Windows[0]` in `ObservationReadResult`. Harmless, but consumers get two places holding the same thing; worth a comment pinning which is canonical.

One logistical note: the branch is BEHIND main, so it will need a rebase before merge.

This has been sitting for 15 days without a human look, and it is a careful implementation of an approved issue. @dnlrsls worth a pass when you have a window.

dnlrsls:
Thank you for the thoughtful implementation, tests, and documentation. We revisited the premise of #812 and are closing this PR because the feature is no longer aligned with Engram’s product boundary, not because of implementation quality.

Engram observations are curated memory units rather than document blobs. Ranged reads, match windows, offsets, and paging would move `mem_get_observation` toward document-navigation semantics and encourage storing artifacts that should remain files. The preferred workflow is to keep large documents in the repository and store their summaries, decisions, paths, state, and relations in Engram.

Issue #812 has been closed as not planned and its prior approval withdrawn. We appreciate the care put into this contribution.

Issue #858, opened 2026-08-28
bug(sync): enrolled repair still scales by projects x rows before first sync


dnlrsls:
### 📋 Pre-flight Checks

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/engram/issues) and this is not a duplicate
- [x] I understand this issue needs `status:approved` before a PR can be opened

### 📝 Bug Description

PR #796 correctly removed `repairEnrolledProjectSyncMutations` from synchronous `Store.New`, fixing the MCP startup timeout reported in #665. However, the same repair algorithm still runs before the first cloud export or autosync push.

`repairEnrolledProjectSyncMutations` iterates every enrolled project and calls four completeness checks in `projectNeedsBackfill`. On an already-backfilled store, these checks repeatedly evaluate sessions, observations, prompts, and relations for each project. The remaining fast path therefore scales with enrolled projects multiplied by stored rows, even when no journal mutations are missing.

### 🔄 Steps to Reproduce

1. Check out `main` at `975490b20e7f07171a9cfb753029c2afb10238ee`.
2. In an isolated Store benchmark fixture, create `E` enrolled projects and `O` total sessions plus live observations.
3. Add the corresponding local `sync_mutations` rows so every project is already fully backfilled.
4. Invoke the production `repairEnrolledProjectSyncMutations` path with fixture setup excluded from timing.
5. Measure the 3x3 matrix `E = 1, 10, 100` and `O = 1K, 10K, 100K` with `-benchtime=1x`.
6. Compare it with a test-local set-based detector that scans each source once and returns only enrolled projects with missing mutations.

### ✅ Expected Behavior

An already-backfilled repair check should scale approximately with the number of source rows plus enrolled projects. Increasing enrolled projects should not cause every source table to be reevaluated once per project.

Repair must still complete before export/push drains the journal, preserve effective-project fallback and relation ownership semantics, and remain idempotent and retryable.

### ❌ Actual Behavior

The current path retains a projects-by-rows work term:

| Enrolled projects / rows | Current per-project path | Test-local set-based detector | Improvement |
| --- | ---: | ---: | ---: |
| 1 / 1K | 1.93 ms | 1.57 ms | 1.2x |
| 10 / 10K | 97.71 ms | 21.24 ms | 4.6x |
| 100 / 10K | 930.87 ms | 19.89 ms | 46.8x |
| 100 / 100K | 12.09 s | 194.59 ms | 62.1x |

[...]

Issue #503, opened 2026-06-15
Mutation-push returns an opaque 500 and fails the whole project batch on a single empty-content observation (no repairable 400, no quarantine)


forNerzul:
### 📋 Pre-flight Checks

- [x] I have searched existing issues and this is not a duplicate
- [x] I understand this issue needs `status:approved` before a PR can be opened

### 📝 Bug Description

`POST /sync/mutations/push` accepts an empty-content observation upsert at the validation gate but then fatally rejects it during chunk canonicalization, surfacing as an opaque HTTP 500 that fails the entire per-project batch.

The push gate is lenient: `validateLegacyPayload` is a no-op for `observation`/`session`/`prompt` (`internal/cloud/cloudserver/mutations.go:340-353`), so an empty-content observation passes. The same batch is then canonicalized, where empty content is a hard error — `"observation payload content is required for upsert"` (`internal/cloud/chunkcodec/chunkcodec.go:412`) — which returns to the client as a bare `http.Error(..., StatusInternalServerError)` (`mutations.go:204`). It fires before any transaction (`materializedMutationBatchChunks`, `cloudstore.go:734`, before `BeginTx` at `:739`). Because the client is all-or-nothing (`transport.go:347-350`; autosync refuses partial acks, `autosync/manager.go:532-534`), a single poison row blocks the whole per-project batch and autosync stalls into degraded/backoff (`MaxConsecutiveFailures=10`, `manager.go:144`) until the row is removed by hand.

### The defect (precise)

Not batch atomicity — that's by design (BW3). The defect is: (1) validation-vs-canonicalization inconsistency (the gate accepts what canonicalization rejects); (2) an opaque 500 instead of an actionable error; (3) no push-side dead-letter/quarantine/repair.

### Asymmetry with chunk-push

The chunk-push path handles the identical empty-content case with a repairable 400: `validateImportableChunkPayload` → `writeActionableError(w, StatusBadRequest, ...Repairable, ...PayloadInvalid, ...)` (`internal/cloud/cloudserver/cloudserver.go:464`). So the two ingest paths disagree on the same bad input: chunk-push → repairable 400; mutation-push → opaque 500.

### 🔄 Steps to Reproduce

1. Self-hosted engram-cloud from `main`; enroll project `P`.
2. Ensure one observation in `P` has empty content (e.g. a pre-#424 `session_summary`).
[...]

PR #914, opened 2026-09-01
test(store): bound scan page race fixture (#857)

### LINKED ISSUE #857: Slow test: TestScanProject_PageProof costs 26x under -race, breaks default-timeout race runs of internal/store
### Bug Description

`TestScanProject_PageProof` (`internal/store/scan_page_test.go`, introduced in 9ab5e35) is 26x slower under the race detector than the rest of the `internal/store` package combined, which makes the standard local validation command `go test -race ./internal/store/` exceed the default 10-minute package timeout and fail with a `panic: test timed out` that looks like a hang. CI never sees this because `ci.yml` runs `go test ./...` without `-race`.

### Steps to Reproduce

1. Checkout `main` @ 975490b.
2. `go test -race -count=1 ./internal/store/` (default 10m timeout).
3. The run dies at 10m with `panic: test timed out`, reporting `TestScanProject_PageProof (8m+)` still running.

Isolated timing on one Linux x86_64 machine:

```
# without -race:
ok  github.com/Gentleman-Programming/engram/internal/store  19.542s   (targeted -run TestScanProject_PageProof)

# with -race, same single test:
ok  github.com/Gentleman-Programming/engram/internal/store  510.677s  (completes in ~8.5 minutes)

# full package with -race: earlier tests consume ~2 minutes, then this test
# needs ~8.5 more → total >10m → default-timeout panic.
```

It is not a deadlock: the test goroutine stays `[runnable]`, burning CPU inside `modernc.org/sqlite/lib._sqlite3VdbeExec` during the seeding loop, and the run completes green when given enough time.

### Expected Behavior

`go test -race ./internal/store/` completes within the default timeout, or the fixture is cheap enough that race runs stay viable for contributors and CI.

### Actual Behavior

```
panic: test timed out after 10m0s
running tests:
	TestScanProject_PageProof (8m10s)
...
goroutine N [runnable]:
modernc.org/sqlite/lib._sqlite3VdbeExec(...)
modernc.org/sqlite/lib._sqlite3Step(...)
modernc.org/sqlite.(*conn).ExecContext(...)
database/sql.(*Tx).Exec(...)
.../internal/store.(*Store).AddObservation.func1(...)  store.go:2829
.../internal/store.scanPageAdd(...)                   scan_page_test.go:97
.../internal/store.scanPageStore(...)
```

### Environment

- OS: Linux x86_64
- Engram version: `main` @ 975490b (unreleased); test introduced in 9ab5e35
- Go: 1.25.10 toolchain, modernc.org/sqlite v1.45.0
- Agent / Client: n/a (test-suite issue)

### Root Cause

`scanPageStore` seeds 5000 observations one by one through the full `AddObservation` write path. Each insert pays for the pre-insert dedupe `SELECT` plus the `obs_fts_insert` FTS trigger, all through the hooked transaction layer. The race detector multiplies that per-insert cost ~26x, so the seeding alone costs minutes under `-race`. Because `ci.yml` only runs `go test ./...` (no `-race`), main stays green and the cost only surfaces for anyone validating locally with `-race` (as happened while validating PR #613).

### Suggested Fix

Two independent directions, either helps:

1. Make the fixture cheap: seed with a bulk insert path (or a reduced count with an explicit pagination proof) so the test measures `ScanProject` paging, not 5000 hooked FTS writes. The 5000-row corpus is the fixture's choice, not the behavior under test.
2. Give CI one `-race` job for `internal/store` (even `-short`), so this cost class stays visible instead of being discovered per-contributor.

Happy to send a PR for either once a direction is picked.

### Additional Context

Observed while validating the PR #613 rebase: the full-package `-race` run failed at the default timeout and initially looked like a hang; the isolated 20-minute run above shows it completes (510s). Stack excerpt and timings available if useful.

Reaan06:
## 🔗 Linked Issue

Fixes #857

## 🏷️ PR Type

- [x] `type:bug` — Bug fix
- [ ] `type:refactor` — Code refactoring

## 📝 Summary

- Keep the page-proof fixture bounded under the race detector.
- Preserve the same pagination and dry-run assertions while avoiding unnecessary per-row setup overhead.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/scan_page_test.go` | Seed the bounded fixture transactionally and use the approved smaller E×O test size. |

## 🧪 Test Plan

- [x] `go test ./internal/store -run ^TestScanProject_PageProof$ -count=1`
- [x] `go test -race ./internal/store -run ^TestScanProject_PageProof$ -count=1`
- [x] `git diff --check`

## ✅ Contributor Checklist

- [x] Linked the approved issue #857
- [ ] Added exactly one `type:*` label: maintainer action required
- [x] Regression coverage remains in the same test file
- [x] Commit follows Conventional Commit format

## Chain Context

| Field | Value |
|-------|-------|
| Chain | approved-issue-fixes
| Tracker PR | Requested upstream tracker branch
| Position | Focused test-performance slice
| Base | `main` (temporary fallback)
| Depends on | None
| Follow-up | Retarget after the upstream tracker exists
| Review budget | 20 changed lines / 400
| Starts at | current `main`
| Ends with | standalone #857 race-fixture stabilization

### Scope
- Includes: #857 test fixture stabilization only.
- Excludes: #858 enrolled repair, #503 cloud validation, import/alias behavior, MCP, Codex, and project/setup stabilizers.

## Request for maintainers

Please add the required `type:bug` label and create the upstream tracker branch for the requested feature-branch chain. The contributor account can publish fork branches but cannot create upstream branches or labels.

## Summary by CodeRabbit

* **Performance**
  * Improved the efficiency and speed of processing page scan observations, particularly for larger batches.
  * Reduced overhead during observation collection.
* **Tests**
  * Streamlined scan validation with a smaller observation set while preserving coverage.
  * Continued benchmark coverage for larger observation batches.


Reaan06:
This is the dedicated #857 bounded scan-page race fixture. The normal and race page-proof tests passed.

Please add the missing `type:bug` label and create the upstream tracker branch.

danielgap (changes requested):
**Blocking: `BenchmarkScanProject_Page5000` now benches a 250-row pool.**

The fixture constant is shared, so lowering `scanPageTestObservations` to 250 also shrinks this benchmark's corpus. Measured on the PR branch with identical code: 162 ms/op at 250 rows vs 1.64 s/op at 5,000 rows. A silent 10x workload cut makes the name false and would hide future ranking-cost regressions behind the smaller pool.

Cheapest fix: give the benchmark its own constant (`scanPageBenchmarkObservations = 5_000`). Seeding 5,000 rows through the new transactional path costs ~3s, and normal test runs (including `-race`, and CI's `go test ./...`) never execute benchmarks without `-bench`, so the original timeout problem does not come back.

**Follow-up worth tracking: CI still can't see this class of failure.**

#857 notes `ci.yml` runs `go test ./...` without `-race`, which is why this regressed silently. I measured the cost basis on `danielgap/engram@fix/857-scan-page-race-cost`: full `go test -race ./...` lands around 4.4 min locally, roughly double on a GitHub runner, inside a 30 min job. Happy to open that as a separate slice.

**Verified:** re-ran the scan-page suite under `-race` after the `main` merge (12.2s for the four tests), and the full package earlier at 196.7s. The fixture bound and transactional seeding both work, and the page-proof shape holds (100-row page, cursor at `ids[99]`, 150-row continuation).

Reaan06:
Addressed in commit ffa241a.

`BenchmarkScanProject_Page5000` now uses a dedicated `scanPageBenchmarkObservations = 5_000` constant, while the race-proof tests keep their smaller shared fixture. This preserves the benchmark corpus at 5,000 rows without reintroducing the normal-test setup cost. Focused scan-page `-race` tests pass, and one benchmark iteration passed at approximately 4.40 s/op.

The CI `-race` follow-up remains separate.

PR #918, opened 2026-09-01
fix(pi): honest empty search - do not report HTTP 200 null body as unreachable (#672)

### LINKED ISSUE #672: bug(pi): mem_search fails with 'could not reach' for queries with underscores/hyphens returning 0 results
## Bug Description

`mem_search` from Pi (gentle-engram v0.1.10 + engram v1.20.0) reports `"gentle-engram could not reach the Engram HTTP server"` for specific queries — but Engram is actually running and healthy. The failure is **deterministic by query pattern**, not intermittent.

## Environment

- **gentle-engram**: 0.1.10
- **engram**: 1.20.0
- **OS**: Windows 11
- **Agent**: Pi (el Gentleman)
- **Shell**: Git Bash

## Steps to Reproduce

```
// These work:
mem_search(query: "hello", project: "my-project")        // ✅
mem_search(query: "simple", project: "my-project")       // ✅

// These FAIL with "could not reach":
mem_search(query: "snake_case_var", project: "my-project")        // ❌
mem_search(query: "some-tag value", project: "my-project")        // ❌
mem_search(query: "mixed_case-name test", project: "my-project")  // ❌
mem_search(query: "kebab-case-query", project: "my-project")      // ❌
```

## Expected Behavior

Return `"No memories found"` (empty result) like the CLI does.

## Actual Behavior

```
gentle-engram could not reach the Engram HTTP server at http://127.0.0.1:7437.
The Pi-native mem_* tools are registered, but the native memory provider is not
currently responding. Run mem_doctor or restart Engram.
```

### Engram CLI — works perfectly

```
$ engram search "snake_case_var" --project my-project
→ No memories found for: "snake_case_var"    (instant)

$ engram search "kebab-case-query" --project my-project
→ No memories found for: "kebab-case-query"    (instant)
```

### HTTP API — works perfectly

```
$ curl "http://127.0.0.1:7437/search?q=snake_case_var&project=my-project&limit=5"
→ HTTP 200 | 26ms | body: "null"
```

### Health check — always OK

```
$ curl http://127.0.0.1:7437/health → 200
$ mem_doctor → 4/4 checks OK
$ mem_context → works always
$ mem_save → works always
```

## Root Cause Hypothesis

The bug is in the gentle-engram Pi plugin layer (bridge between Pi `mem_search` tool and Engram HTTP API), not in Engram core. The plugin appears to mishandle HTTP responses for queries that:

1. Return **zero results** (empty/null response)
2. AND contain special characters (`_`, `-`)

The plugin interprets a valid empty response as a connection failure and reports "could not reach" instead of returning empty results.

## Pattern Analysis

| Query | Has special chars? | Returns results? | `mem_search` |
|---|---|---|---|
| `"hello"` | No | Yes | ✅ |
| `"simple"` | No | Yes | ✅ |
| `"snake_case_var"` | Yes (`_`) | No | ❌ |
| `"kebab-case-query"` | Yes (`-`) | No | ❌ |

This was discovered while investigating intermittent Engram failures. The initial suspicion was parallelism (multiple `mem_search` calls in one `<invoke>` block), but systematic testing proved the bug is query-dependent, not concurrency-dependent.

Seinra:
## Summary

Delivers the plugin-side half of the two-layer fix proposed in #672 (comment 2026-08-26). The server-side half is already merged (#862) — this PR closes the remaining gap.

## Problem recap

`engramFetch` returns `null` for three distinct outcomes:

1. No HTTP response at all (server unreachable / timed out) — genuine failure
2. HTTP 200 whose body fails to parse — server bug
3. HTTP 200 with body being the JSON literal `null` — what older `/search` servers legitimately return for zero hits

The caller (`executeMemoryTool`) treated **every** `null` as outcome 1, so a zero-hit search was reported to the agent as "could not reach Engram" — inviting pointless retries and hiding the honest "no memories found" answer.

## Fix (plugin/pi/index.ts only)

- Track whether the most recent `engramFetch` actually received an HTTP response (`lastFetchSawResponse`), reset per call.
- `data === null` **without** a seen response → unreachable error, unchanged semantics.
- `data === null` **with** a seen response → honest empty result: `No memories found.` with empty `details.data`, plus the standard status-bar update.

## Why still needed after #862

- #862 fixed the server for new builds, but installed servers already in the field still return `null` bodies.
- The client-side distinction is defense in depth: any current or future endpoint that legitimately responds with an empty body must not be misreported as a connection failure.

## Testing

- `esbuild` parse/type-stripping pass on `plugin/pi/index.ts` (no build infrastructure exists in-repo for the plugin; it is loaded directly by Pi).
- Behavior verified live on Windows since 2026-08-26 (local install running this patch): zero-hit `mem_search` queries return "No memories found." instead of the false "could not reach" error.

Closes #672 (together with the already-merged #862).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of server connectivity failures versus successful responses with no results.
  * Requests returning no memories now display a clear “No memories found” status.
  * Connectivity errors are reported only when the server cannot be reached.

### DIFF
--- a/plugin/pi/index.ts
+++ b/plugin/pi/index.ts
@@ -181,6 +181,11 @@ function isTimeoutError(error: unknown): boolean {
 // treat a null registration response as unacknowledged and stop before writing;
 // other callers retain the existing null fallthrough contract.
 let lastFetchTimeoutMethod: string | undefined;
+// Whether the most recent engramFetch actually got an HTTP response. Distinguishes a
+// genuine unreachable server (no response) from a successful HTTP 200 whose body is the
+// JSON literal `null` -- which /search returns for zero results and must NOT be reported
+// as "could not reach".
+let lastFetchSawResponse = false;
 
 function takeLastFetchTimeoutMethod(): string | undefined {
   const method = lastFetchTimeoutMethod;
@@ -195,6 +200,7 @@ async function engramFetch<TResponse = unknown>(path: string, opts: FetchOptions
   // on the first leg would mislabel an unrelated failure on the second as "may already have
   // been applied", telling the agent not to retry a write that never left the machine.
   lastFetchTimeoutMethod = undefined;
+  lastFetchSawResponse = false;
   let res: Response | undefined;
   let timedOut = false;
   for (let attempt = 0; attempt < ENGRAM_FETCH_MAX_ATTEMPTS; attempt += 1) {
@@ -223,6 +229,7 @@ async function engramFetch<TResponse = unknown>(path: string, opts: FetchOptions
   // not reach" invites the caller to retry a write whose outcome is genuinely unknown. Record
   // which it was so the tool layer can say what we do and do not know.
   if (timedOut) lastFetchTimeoutMethod = method;
+  if (res) lastFetchSawResponse = true;
   if (!res) return null;
 
   let data: unknown = null;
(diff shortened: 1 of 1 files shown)


Seinra:
Closing this PR for now — the branch remains available on my fork (Seinra/engram, branch fix/672-pi-honest-empty-search). Will resubmit through the agreed coordination flow once issue #672 is triaged. Thanks for merging the server-side half in #862.

Issue #1013, opened 2026-09-05
feat(sync): explicit inbox for mutations from unenrolled projects


Albe841114:
## Summary

Forward-looking design for what happens to sync mutations whose project is **not enrolled**. Today the target space is open: every cwd-derived key creates its own `sync_state` row, and mutations from unenrolled keys accumulate silently in the legacy `cloud` target. This proposal replaces that with a **closed target space**, a single reserved `cloud:inbox`, explicit human operations (list/claim/drop), and warnings that surface **at save time**.

The design was developed and reviewed against a real machine (Linux, engram 1.20.0, self-hosted cloud). All counts below were measured on that machine on 2026-09-05; project names and paths are anonymized.

## Problem (measured)

1. A mutation whose derived key is not enrolled lands in the legacy `cloud` target with `project` = derived key. The worker only delivers enrolled projects → the row stays unacked **forever**.
2. The sync layer creates an empty `sync_state` row per derived key (`cloud:<key>`): lifecycle `pending`, no failures, no error of any kind — invisible by construction. *(Measured on 1.20.0 — on 2.0.0-rc.6 the invisibility is gone: a new blocking doctor check fires on exactly this pathology. What it does and does not do: rc.6 note below the list.)*
3. Measured result: **892 rows stuck + 35 phantom targets accumulating for ~3 months with zero alarms**, because nothing failed — nothing was broken, nobody was looking.
4. Key derivation fragments. On one machine, within one backlog: one folder produced up to 3 keys (casing/separator variants); one non-repo working dir produced **two different keys at different times** (24 sessions under key A from July, 1 under key B in August — context-dependent resolution, not a typo variant); the basename fallback for non-repo cwds is exactly what generated the noise keys (`/` → `/`, scratch dirs → their basenames, a UUID-named project dir, the user home dir → 538 sessions under its basename).
5. 880 of the 892 rows were session/prompt **telemetry** that does not go to cloud by policy — pure churn that should never have been enqueued.
6. **And the per-project lane never delivers.** Measured: the only lane the worker has ever delivered observations from is the legacy `cloud` target, grouped by the `project` column (a fresh observation for an enrolled project enqueued there was delivered in <60 s). The per-project `cloud:<enrolled>` targets hold 135 rows in their entire history — **not one of them has ever been delivered**. The lane this design uses as its central target model has a measured delivery record of zero.

[...]

Issue #1026, opened 2026-09-05
bug(doctor): repair for sync_mutation_required_fields reports applied:true while doing nothing, and unenrolled machines still enqueue cloud mutations (v2.0.0-rc.5)


dyool17:
### 📋 Pre-flight Checks

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/engram/issues) and this is not a duplicate
- [x] I understand this issue needs `status:approved` before a PR can be opened

### 📝 Bug Description

On **v2.0.0-rc.5**, a machine where `engram cloud status` reports `not configured` and `sync_enrolled_projects` is empty is still permanently `blocked` by `sync_mutation_required_fields`, and the local repair path shipped for exactly this class cannot clear it.

This is not the original poison-payload report. The capture-side and source-side fixes **did** land and **are** in my build:

- rc.2 (2026-08-31) — #892 / #814, "invalid observation titles are repaired before they can stall sync queues"
- rc.4 (2026-09-02) — #919 / #629, "`doctor repair` detects active observations with missing required source fields and previews recoverable title repairs by default"

Two gaps survive them:

**1. `doctor repair --check sync_mutation_required_fields` is a silent no-op on already-enqueued payloads.**
The shipped repair looks at the **source** `observations` rows. My four source observations were repaired long ago and all carry correct non-empty titles today. What is still poisoned is the **frozen `sync_mutations` payload** captured back when the title was empty. Because the source rows are clean, `source_repairs` is empty, `repairs` is empty, `actions` is empty — and `--apply` cheerfully returns `"applied": true` while `doctor` stays `blocked` on the same 7 findings. Nothing in the repair surface reaches a stale payload whose source has since been fixed.

**2. The remedy loop is still circular for an unenrolled machine** (the shape #699 described):

```
doctor                       → blocked; "Run `engram cloud upgrade doctor --project <p>`"
cloud upgrade doctor         → blocked; class: repairable; "run `engram cloud upgrade repair --project <p> --apply`"
cloud upgrade repair --apply → blocked; reason_code: upgrade_blocked_manual;
                               "project <p> is not enrolled; run doctor/bootstrap guidance first"
```

`cloud repair materialize-mutations` is not an exit either — it tries to reach Postgres, which does not exist on an unenrolled machine.

[...]

PR #1098, opened 2026-09-09
feat(sync): reserve the cloud inbox target and enforce the closed sync-target space (#1091)

### LINKED ISSUE #1091: feat(sync): cloud inbox foundations — reserved target, telemetry exclusion, closed-space doctor invariant
### 📋 Pre-flight Checks

### 🔍 Problem Description

Slice 1 of the cloud inbox design (#1013). Three decision-independent gaps exist today:

1. Nothing reserves an inbox target. Every cwd-derived project key mints its own `sync_state` row silently (measured on a real deployment: 35 phantom targets accumulating for ~3 months, 892 stuck rows, zero alarms — see #1013 "Problem").
2. The reserved name is not defended: nothing stops a project named `inbox` from being enrolled and colliding with the reserved target.
3. No doctor check enforces a closed target space. A `sync_state.target_key` outside the legitimate set is invisible to every current check (rc.6's blocking check fires on unenrolled *pending mutations*, not on foreign *targets*).

### 💡 Proposed Solution

Land the three decision-independent foundations from #1013 (none depends on the lane decision pending there):

1. **Reserved `cloud:inbox` target.** A `cloud:inbox` constant and a `sync_state` row seeded with its own lifecycle `inbox` (not `pending`: no delivery cursor). The worker only ever reads its single configured target key, so the inbox row is never delivered by construction; it is also excluded from `cloud:%` project aggregations (e.g. `CloudSyncSummary`) so it never renders as a project. Lifecycle setters are guarded so the inbox row can never transition to `pending`/`healthy`.
2. **`enroll inbox` rejection.** The reserved name `inbox` cannot be enrolled, enforced at the store level so every enrollment path (CLI and programmatic) is defended.
3. **Doctor closed-target invariant.** New check: any `sync_state.target_key` outside `{cloud, cloud:inbox, local} ∪ {cloud:<enrolled project>}` is a finding (severity: error). This is the guardrail that makes a future routing bug visible in the first cycle instead of after three months.

Out of scope: routing unenrolled mutations to the inbox, the two-stage derivation pipeline, the delivery-lane change, the `inbox list/claim/drop` CLI, and **telemetry enqueue exclusion** — moved to slice 2 with the protocol finding documented in https://github.com/Gentleman-Programming/engram/issues/1013#issuecomment-5594050631 (sessions are load-bearing in the current sync protocol: server-side chunk validation and client-side FKs both depend on them, so exclusion needs its own protocol decision; the churn the design measured is unenrolled telemetry, which slice 2's enrollment-aware routing addresses).

### 📦 Affected Area

Sync (multi-instance)

### 🔄 Alternatives Considered

Extending rc.6's blocking `sync_mutation_required_fields` check instead: rejected — it detects stuck rows after the fact rather than enforcing the target-space invariant that prevents them, and it halts on noise instead of routing decisions.

### 📎 Additional Context

Part of #1013 (slice 1 of 4, per the slicing agreed in that thread). All measurements cited above are from the #1013 design body, taken on a real single-machine deployment.

danielgap:
## 🔗 Linked Issue

Closes #1091

---

## 🏷️ PR Type

- [ ] `type:bug` — Bug fix

---

## 📝 Summary

- Reserve the `cloud:inbox` sync target with its own `inbox` lifecycle: seeded `sync_state` row, excluded from `cloud:%` project aggregations, and guarded so no lifecycle setter or refresh helper can transition it (no delivery cursor by design).
- Reject enrolling the reserved project name `inbox` at the store level, defending every enrollment path (CLI and programmatic).
- Add the `sync_target_closed_space` doctor check (severity: error): any `sync_state.target_key` outside `{cloud, cloud:inbox, local} ∪ {cloud:<enrolled project>}` is a foreign-target finding with evidence and a safe next step.
- Pin the protocol constraint that session/prompt telemetry still journals to both targets: server-side chunk validation (`hasSession`) and the enforced `observations.session_id` FK make sessions load-bearing until routing lands (see [#1013 (comment)](https://github.com/Gentleman-Programming/engram/issues/1013#issuecomment-5594050631)).

Slice 1 of the cloud inbox design (#1013): pure foundations, zero behavior change for existing flows.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | Inbox target/lifecycle/reserved-name constants, seeded inbox row, `CloudSyncSummary` exclusion, Mark*/refresh no-op guards, `EnrollProject` reserved-name rejection, `ListSyncStates` store query |
| `internal/cloud/constants/constants.go` | `TargetKeyCloudInbox` alias |
| `internal/diagnostic/checks.go` | `SyncTargetClosedSpaceCheck` + `ReasonForeignSyncTarget` |
| `internal/diagnostic/registry.go` | Register the new check |
| `internal/store/cloud_inbox_test.go` | New: inbox seed/reseed, summary exclusion, enroll rejection, telemetry protocol pin, observation/relation journaling, `ListSyncStates` |
| `internal/store/sync_state_test.go` | Inbox lifecycle setter/refresh no-op pins |
| `internal/diagnostic/checks_closed_space_test.go` | New: closed-space check table tests |
| `internal/diagnostic/diagnostic_test.go` | Registry ordering + healthy-fixture enrollment |
| `cmd/engram/cloud_project_input_test.go` | CLI `enroll inbox` rejection test |

## 🧪 Test Plan

- [x] `go build ./...` — PASS
- [x] `go test ./internal/store/... ./internal/diagnostic/...` — ok (both packages)
- [x] `go test ./cmd/...` (excluding `TestCmdServeSignalClosesUnixSocket`, a pre-existing umask-002 environmental failure reproduced at pristine HEAD `a51b80d`) — ok
- [x] `gofmt -l` on all touched files — empty

---

## ✅ Contributor Checklist

- [x] Linked an approved issue (#1091; awaiting `status:approved` from a maintainer)
- [ ] Added exactly one `type:*` label — maintainer-side: I cannot add labels in the org; requesting `type:feature`
- [x] Ran shellcheck on modified scripts — N/A (no shell scripts touched)
- [x] Skills tested in at least one agent — N/A (no skills touched)
- [x] Docs updated if behavior changed — no user-visible behavior change in this slice; the reserved-name rejection message is self-documenting

## Summary by CodeRabbit

* **New Features**
  * Added a reserved Cloud Inbox sync target with stable lifecycle handling.
  * Added diagnostics to identify unsupported or unassociated sync targets.
  * Added sync target health and activity reporting.

* **Bug Fixes**
  * Prevented enrollment of projects named “inbox.”
  * Prevented regular lifecycle updates from changing Cloud Inbox state.
  * Excluded Cloud Inbox activity from regular cloud sync summaries.
  * Reported each foreign sync target separately.
  * Preserved and repaired the reserved Cloud Inbox state when opening existing databases.


danielgap:
Addressing the CodeRabbit findings from the first pass in b34f421:

- **Repair existing `cloud:inbox` state during migration (Major)** — fixed: the migration now pins the lifecycle to `inbox` and resets the delivery cursors for a legacy conflicting row, with a WHERE guard so already-clean state is untouched. Covered by `TestInboxRowRepairPinsLegacyConflictingState`.
- **Add error-path and multiple-row coverage (Minor)** — fixed: `TestSyncTargetClosedSpaceReportsEveryForeignTarget` (one finding per foreign row, evidence verified per key) and `TestSyncTargetClosedSpaceReturnsStoreReadError` (closed store surfaces the read error).
- **Count default-cloud project journal rows in `CloudSyncSummary` (Major)** — skipping in this PR with reason: the `target_key LIKE 'cloud:%'` restriction on the pending count is pre-existing base behavior (this PR only added the `cloud:inbox` exclusion to it); standard project backlog living under the default `cloud` key was never counted by this summary. Changing that counting semantic belongs to its own fix with its own tests — happy to open a follow-up issue if maintainers want it tracked.

The three new `errcheck\> failures from the first CI run are also fixed in b34f421 (checked closes + `closeRowsWithError` idiom).

Albe841114:
Running the promised real-data check for slice 1. Caveat first: **I did not build this branch.**
What follows applies the closed-space rule exactly as the PR states it to the live
`sync_state` of a three-month single-user deployment. It is a projection of what
`sync_target_closed_space` will find, not an execution of the new code — happy to
re-run it as a real build if you'd rather have that.

### What the check would find here: 38 of 49 targets are foreign

```
targets in sync_state : 49
enrolled projects     : 10
allowed by the rule   : 13   ({cloud, cloud:inbox, local} ∪ {cloud:<enrolled>})
foreign               : 38
```

**Every one of the 38 carries zero mutations.** They are empty shells left behind by
target creation, not stalled work.

A sample of what they actually are, because the shape matters more than the count:

| target | lifecycle | what it is |
|---|---|---|
| `cloud:-project` | **degraded** | a CLI flag captured as a project name |
| `cloud:rev_risk`, `cloud:rev_readability`, `cloud:rev_reliability` | pending | gentle-ai review lens names |
| `cloud:cli_agentes` | pending | separator variant of enrolled `cli-agentes` |
| `cloud:open_design` | pending | separator variant of enrolled `open-design` |
| `cloud:via lactea` + `cloud:via-lactea` | pending | both variants coexist |
| `cloud:workspace` | **healthy** | foreign target sitting in a healthy lifecycle |
| `cloud:/`, `cloud:unknown`, `cloud:prueba`, `cloud:497020aa-cd6d-4bc7-851c-f29f735274c5` | pending | directory names, a literal `/`, a uuid |

Two of these are worth separating from the noise:

- **`cloud:-project` reached `degraded`**, which means it was not merely created — it was
  handed to the transport and retried. A parsed flag became a sync target.
- **`cloud:workspace` is `healthy`** while being outside the allowed set, so the closed-space
  invariant and the lifecycle state currently disagree on the same row.

This is good news for the PR's premise: the closed-target space is not a theoretical
tidiness rule, it is describing a real and large drift. The derivation fallback that
#1013 removes is visibly the source of most of these.

### The one thing I'd flag before merge: severity `error` with no remedy

The check ships at severity `error`. On this deployment that turns a `doctor` that is
`blocked` today into one carrying **38 error findings** the moment slice 1 lands, and
slice 2 (which removes the basename fallback) does not retroactively remove rows that
already exist in `sync_state`.

There is no shipped way to clear them: `doctor repair --check sync_mutation_required_fields`
is a no-op for the neighbouring unenrolled-project finding (#1026, confirmed on rc.8 today),
and `projects prune` deletes projects without touching `sync_state` or `sync_mutations`
(measured on an isolated copy: 8114 mutation rows before and after, `max(seq)` unchanged).

So an operator upgrading into slice 1 gets 38 red findings, a correct diagnosis, and
nothing to run. That is the same shape as the loop described in #1026.

Not asking for scope creep — the foundations are right and I would land them. Two
options that keep this slice pure:

1. Ship the check at `warning` for now and raise it to `error` when the cleanup path
   from a later slice exists.
2. Keep `error` and state the remedy explicitly in `safe_next_step`, even if the remedy
   is "these are inert; a later slice removes them" — so the finding does not read as
   an actionable failure the operator cannot act on.

Either way the invariant itself is the right one, and the numbers above are what it
will meet in the field.

### Offer stands

Send me slice 2 when it is up and I will run the same measurement against the same
store, including a real build if you want the check executed rather than projected.

### Environment

engram 2.0.0-rc.8 client, cloud backend 2.0.0-rc.6, Linux (CachyOS), single-user
deployment, ~3 months of history: 984 observations, 8114 sync mutations, 10 enrolled
projects, 27 projects in the store. All reads above were read-only (`mode=ro`).

danielgap:
Went with option 2: the check stays at `error`, and the foreign-target `safe_next_step` now states the remedy explicitly at 2b0b31d.

It now reads: enroll if the target belongs to a project you want synced, otherwise no action is required because the row is inert drift from the removed derivation fallback, it cannot advance, no data is at risk, and a later cloud-inbox slice removes these legacy rows automatically.

That keeps the signal strength you validated (38 findings that are a correct diagnosis of real drift) while making each one explicitly non-actionable-but-benign instead of a dead end, which was the #1026-shaped trap you flagged.

Two things on your measurement, because they were genuinely useful:
- `cloud:-project` reaching `degraded` after being handed to the transport is a sharper datapoint than the count itself. It confirms these rows are not just cosmetic residue, they were actively consumed by the old fallback path.
- The `cloud:workspace` disagreement (healthy lifecycle outside the allowed set) is exactly the invariant/state divergence slice 1 exists to surface, so it landing as an error finding is the check doing its job on real data.

Slice 2 is next on my side once this lands. I'll take you up on the re-measurement against the same store, and yes please with a real build if the offer still stands, so the check runs instead of being projected.

Albe841114:
Built `2b0b31d` and ran it against the same store. **The projection held exactly.**

```
CGO_ENABLED=0 go build -trimpath -ldflags="-s -w -X main.version=$(git describe --tags --always)" ./cmd/engram
```

Run on an isolated copy of the live store (`VACUUM INTO` + separate `HOME`); the real store was never touched.

### The check, executed

```
[error] sync_target_closed_space — 38 finding(s) detected.
```

**38, and not just the count** — I diffed the executed target list against the one I projected, element by element:

```
projected: 38 | executed: 38 | identical set: True
```

No target I predicted was missing, and the check found none I had not predicted. The closed-space rule as implemented matches the rule as documented.

### The new safe_next_step renders as intended

```
next: If the target belongs to a project you want synced, run `engram cloud enroll <project>`.
      Otherwise no action is required: the row is inert drift left by the removed derivation
      fallback, it cannot advance and no data is at risk, and a later cloud-inbox slice removes
      these legacy rows automatically.
```

That reads correctly against the 38 real findings here. Every one of them is a `0`-mutation shell, so "inert drift" is accurate for this store, and an operator hitting this now has somewhere to go instead of a dead end. This addresses what I raised.

### Slice 1 is non-disturbing — measured, not assumed

Same store, same copy, two binaries:

| | checks | ok | warning | blocked | error |
|---|---|---|---|---|---|
| `2.0.0-rc.8` (installed release) | 7 | 5 | 1 | 1 | 0 |
| `2b0b31d` (this PR) | 8 | 5 | 1 | 1 | **1** |

The seven pre-existing checks return identical verdicts. The only delta is the new check. The PR body's claim — *"pure foundations, zero behavior change for existing flows"* — holds on a three-month real store.

### Reserved-name rejection verified

```
$ engram cloud enroll inbox
engram: project name "inbox" is reserved for the cloud inbox and cannot be enrolled
```

### One build-from-source note, not a defect in this PR

Building from a plain checkout, `git describe --tags --always` resolves to a `pi-*` tag rather than a `v2.*` one:

```
describe: pi-v0.1.12-3-g2b0b31d
$ engram --version
engram pi-v0.1.12-3-g2b0b31d
```

The consequence is cosmetic but confusing: the update checker parses `0.1.12` out of that string and prints

```
Update available: pi-v0.1.12-3-g2b0b31d -> 1.20.0
```

recommending a downgrade from a 2.x build. Release artifacts are unaffected (goreleaser injects `{{ .Version }}` explicitly), so this only hits people building from source to test a PR — which is exactly what this thread is asking for. Mentioning it so the next person reproducing does not read that banner as real.

### Standing offer

Send slice 2 when it is up and I will run the same measurement on the same store, real build again.

### Environment

engram built from `2b0b31d`, Go 1.27.1, Linux (CachyOS). Store: 984 observations, 8114 sync mutations,
10 enrolled projects, 49 `sync_state` targets, ~3 months of single-user history. Cloud backend 2.0.0-rc.6.

Albe841114:
Correction to my previous comment: **"Every one of them is a 0-mutation shell" is wrong for one of the 38.**

The check's own evidence reports:

```
cloud:bin -> {"lifecycle": "pending", "unacked_mutations": 1}
```

37 of 38 carry `unacked_mutations: 0`. `cloud:bin` carries 1.

I had counted `sync_mutations` filtered by `target_key`, which returns 0 for that row. The check counts project-scoped pending mutations under the default `cloud` key, so the one pending session mutation whose `project` is `"bin"` (seq 7940, enqueued 30 h before my run) is attributed to `cloud:bin`. Two different counting rules; the one a reproducer sees is the check's, so my universal claim was falsifiable against your own output.

Nothing else in that comment changes: 38 findings, the target set matches my projection element by element, the seven pre-existing checks return identical verdicts, and the reserved-name rejection reproduces. The `safe_next_step` guidance also still holds for this row — an unenrolled cwd-derived key with one stranded telemetry mutation is exactly the inert drift the text describes.

Flagging it because a claim that specific should be reproducible, and that one is not.

danielgap:
Thanks for the correction, and for holding the claim to that bar. Confirmed in `ListSyncStates` (internal/store/store.go): for `cloud:<project>` targets the count unions the per-target ledger with the project-scoped journal rows living under the default `cloud` key (`sm.project = substr(ss.target_key, 7)`), so the seq-7940 telemetry mutation you identified lands on `cloud:bin`. My earlier read of your numbers used only the per-target arm, same as your first pass.

The verified picture is then 37 inert 0-mutation rows plus one stranded-mutation row, and the `safe_next_step` guidance holds for both: an unenrolled cwd-derived key cannot advance, nothing is at risk, and the cloud-inbox slice removes them.

dnlrsls:
@coderabbitai resume

PR #1126, opened 2026-09-10
fix(pi): end the Engram session on session_shutdown

AlejandroTatum:
## Problem

The Pi plugin `session_shutdown` handler only cleared in-memory maps and never told the server the session ended. Every Pi session stayed open (`ended_at IS NULL`), so the omitted-session resolver later failed with `multiple active runtime sessions` for every agent sharing the project and directory (Claude Code included). On one machine this accumulated 556 open sessions across 15 projects in a week.

## Fix

- `session_shutdown` now issues a best-effort `POST /sessions/<id>/end` before clearing state. An unreachable server never makes shutdown throw or hang.
- Tests: the fake `engram serve` records incoming requests; one test asserts the end request after start+shutdown, another asserts shutdown completes when the server is unreachable.

## Verification

- `cd plugin/pi && npm test`: 127 pass, 0 fail
- Manual: with the patched plugin loaded from a local path, a `pi -p` run leaves its session with `ended_at` set.

https://claude.ai/code/session_018CQsjiGFNfGEewfQ8HSPN4

## Summary by CodeRabbit

* **Bug Fixes**
  * Sessions are now ended on the server when the application shuts down.
  * Shutdown remains non-blocking and completes safely if the server is unreachable.
  * Sessions that were not previously registered are also handled during shutdown.

### DIFF
--- a/plugin/pi/index.ts
+++ b/plugin/pi/index.ts
@@ -1317,6 +1317,12 @@ export default function registerEngram(pi: ExtensionAPI) {
   pi.on("session_shutdown", async (_event: unknown, ctx: SessionContext) => {
     const sessionId = observeRuntimeSessionID(ctx);
     if (!sessionId) return;
+    // Best-effort: an unreachable server must never keep shutdown from completing, and a
+    // session this process never registered still gets the same idempotent end request.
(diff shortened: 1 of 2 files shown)


dnlrsls:
closed bc no issue related

PR #1189, opened 2026-09-14
fix(mcp): exit gracefully when the stdio parent pipe closes (#886)

### LINKED ISSUE #886: engram.exe leaves orphaned processes on Windows, blocking Claude Desktop updates
On Windows, `engram.exe` (v1.20.0) survives after its parent process (Claude Desktop) exits — confirmed 2 orphaned instances running simultaneously with no parent Claude process alive. Because it isn't tied to the parent's lifetime (no Job Object / no stdin-EOF exit handling on Windows), it keeps a file lock that causes the Claude Desktop auto-updater to fail with a generic "file in use" error. Killing `engram.exe` manually (`taskkill /F /IM engram.exe`) before updating resolves it every time.

Expected: the process should exit when its stdio pipe to the parent closes (or be added to a Windows Job Object at spawn time so it dies with its parent), same as the SIGTERM handling already present for autosync (`releaseLease()` in `Run()`).

Repro:
1. Launch Claude Desktop with engram configured as MCP server
2. Close Claude Desktop
3. Check Task Manager — `engram.exe` is still running

Environment: Windows 11, engram 1.20.0, installed at `%LOCALAPPDATA%\engram\bin\engram.exe`

danielgap:
## 🔗 Linked Issue

Closes #886

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix
- [ ] `type:question` — Question requiring tracked work

---

## 📝 Summary

- `engram mcp` now owns the stdio transport lifecycle so EOF and SIGINT/SIGTERM unwind cleanly instead of reaching the fatal exit path.
- EOF gives mcp-go a bounded five-second window to drain accepted `tools/call` work and write the final JSON-RPC response. If a context-aware handler remains stuck, the deadline cancels it; autosync releases its lease only after `Listen` returns.
- Autosync startup and shutdown now use an explicit readiness handshake, preventing immediate EOF from racing startup or closing the store while the sync loop is still registering.

Scope note: this implements #886's primary stdin-close shutdown contract. The separate Windows pipe-retention/job-object hardening remains tracked in #1190 for cases where the parent never delivers EOF.

## 📂 Changes

| File | Change |
|------|--------|
| `cmd/engram/main.go` | Own stdio EOF/signal lifecycle, preserve final tool responses, bound EOF drain time, and defer autosync release until transport shutdown completes. |
| `cmd/engram/mcp_stdio_eof_test.go` | Cover EOF, SIGTERM, data-plus-EOF buffering, final real `tools/call` response delivery, bounded response waits, deferred autosync release, and context-waiting handler cancellation. |
| `internal/cloud/autosync/manager.go` | Add a startup readiness handshake and make `Start` a no-op when the manager is already running. |
| `internal/cloud/autosync/manager_test.go` | Cover immediate stop, startup registration, and bare `Run` followed by `Start`. |

## ✅ Test Plan

- [x] `git diff --check`
- [x] `gofmt -d cmd/engram/main.go cmd/engram/mcp_stdio_eof_test.go`
- [x] `go test ./cmd/engram/ -run 'TestCmdMCPStdio|TestEOFShutdownReader' -race -count=1`
- [x] `go test ./cmd/engram/ -run TestCmdMCPStdioFinalToolCallDrainsBeforeEOFShutdown -race -count=20`
- [x] Full `go test ./cmd/engram/... -count=1` attempted locally; the unrelated host-project expectation in `TestCmdSyncDefaultProjectNoData` fails identically on the pre-change PR head.
- [x] Current GitHub CI is green for `72053e2` (performance ratchet skipped by policy).

## 🔍 Review Provenance

Native high-risk four-lens review covered risk, resilience, readability, and reliability. It identified two candidate-caused shutdown-ordering defects; the bounded correction keeps autosync active through queue drain and adds the EOF cancellation deadline. Targeted validation approved the corrected candidate and the acknowledgement was burned (`gentle-ai.review-acknowledged/v1`, lineage `review-10e3d392591aeeef`). Remaining advisories are informational and non-blocking.

CodeRabbit's follow-up request to bound the initialize scan was implemented in `72053e2`, verified under `-race` for 20 repetitions, and separately approved and acknowledged by native reliability review (`review-3456a6d00f173019`). No review threads remain unresolved.

## 📏 Review Workload

The cumulative PR diff is 917 changed lines. The maintainer explicitly accepted a `size:exception`: one honest slicing pass found no cohesive sub-400 split that keeps the MCP lifecycle behavior with its deterministic real-transport regression coverage.

## ✅ Contributor Checklist

- [x] Linked approved issue #886.
- [x] Added exactly one `type:*` label (`type:bug`).
- [x] Used Conventional Commits with no `Co-Authored-By` trailer in the new commits.
- [x] Changed paths comply with the transient-artifact policy.
- [x] Documentation is not required because no CLI syntax, public API, or configuration contract changed.
- [x] Maintainer approved the review-size exception.

## 🤖 AI Assistance

Implementation and correction work were delegated under strict bounded edit surfaces, verified with focused race tests, and reviewed through the native review lifecycle, orchestrated by el Gentleman under maintainer direction.


danielgap:
Thanks for the review. All three findings are addressed in 3248065:

1. **Autosync startup/shutdown handshake.** `Manager.Start` now installs a `runReady` channel that closes only after `Run` registers with the wait group, and `Stop` waits for that handshake before reading `cancelFn` (re-read under the lock after the wait, since a pre-wait read can see nil). An immediate-EOF shutdown can no longer return before the goroutine is even scheduled. `tryStartAutosync` launches through the new `Start` capability; the injectable test fakes keep the plain goroutine fallback.

2. **Deferred EOF unwind.** `eofShutdownReader` retains a data-plus-EOF read (`pendingEOF`) and returns the bytes with nil error. Since bufio does not cache an error for a nil-error read, it must call back once the buffer drains; that next call surfaces the retained EOF and runs the graceful sequence, so the final buffered request is processed and answered before the context is cancelled. Immediate EOF and read errors still unwind right away, and the EOF stays sticky with the sequence once-guarded.

3. **Deterministic test waits.** `requireGracefulOutcome` now uses a bounded `waitForSignal` helper for both `runStarted` and `stopCalled` instead of the non-blocking selects.

New coverage: handshake tests in `internal/cloud/autosync` (ready-after-registration, Stop blocks until registration, re-entry does not re-close), plus wrapper-level and bufio-level subtests for the deferred EOF path.

One note for the record: `TestCmdServeSignalClosesUnixSocket` fails in my environment on both this branch and its parent commit. It is the strict unix socket directory check rejecting the group-writable `t.TempDir()` produced under umask 0002, unrelated to this diff.

danielgap:
Could a maintainer add `type:bug` here? The `Check PR Has type:* Label` workflow is failing, and I don't have permission to add labels on this repo.

Everything else is green, and the three CodeRabbit findings from the last review are addressed in 3248065.

danielgap:
Addressed in f5d4c91. `Start` now checks `cancelFn` under the lock and returns as a no-op when the manager is already running, so the bare Run-then-Start sequence can no longer install a readiness channel that no Run will close. The regression test pins it: bare `Run`, wait for registration, `Start`, then assert `runReady` stays nil and `Stop` returns within the timeout guard.

dnlrsls:
@coderabbitai resume

Issue #1247, opened 2026-09-17
Sessions never end: 1000+ open sessions accumulate and break mem_save with 'multiple active runtime sessions'


Albe841114:
## What happened

Sessions opened by runtime activity are almost never marked ended. After ~3 months both our nodes carry over a thousand open sessions, and omitted-session `mem_save` calls fail with `multiple active runtime sessions` because the runtime cannot pick a unique active session for a (project, directory).

## Environment

- `engram 2.0.0-rc.12`, Linux x86_64
- Two independent machines, same behavior:
  - node A (`~/.engram/engram.db`, user `pc_albe`): **1037** open sessions (1426 total, 389 ever closed)
  - node B (user `albe`): **1023** open sessions

## Evidence (node A)

```sql
select ownership_mode, count(*), min(substr(started_at,1,10)), max(substr(started_at,1,10))
from sessions where ended_at is null group by ownership_mode;
```

| ownership_mode | n | oldest | newest |
|---|---|---|---|
| shared | 1032 | 2026-06-14 | 2026-09-16 |
| project_owned | 5 | 2026-07-06 | 2026-09-09 |

Sessions started on **June 14 are still open on September 16**. Only 27 % of all sessions were ever
closed (52/257 in June, 239/578 in July, 88/439 in August, 10/152 in September), i.e. closing works
when a client calls it, but in practice most clients never do. New sessions arrive at roughly
5–15/day on this node. One project alone (`infraestructura`) has **739** open sessions, so the
ambiguity is guaranteed there.

## Failure mode

With many open sessions matching the same (project, directory), a `mem_save` without an explicit
`session_id` fails closed with:

> multiple active runtime sessions match the current project and directory; provide session_id, end other active matching sessions, or save independently

`engram doctor` reports the same condition as `ambiguous_active_runtime_sessions`, but by design it
does not select, end, or modify sessions.

## What exists today, and what is missing

- The MCP tool **`mem_session_end`** does end a session while keeping its observations. So the
  primitive exists — but only from inside an agent turn, one session at a time, and only if the
  client remembers to call it before exiting. Most runtimes/plugins do not.
- `engram delete session <id>` refuses any session that has observations (174 of the 1037 open
  sessions on node A), so it cannot clean the ones that matter.
- There is no CLI `session end/close`, no bulk or by-age operation, and no reclamation of sessions
  whose owning process is gone.

## Expected

- Sessions marked ended when the owning runtime process exits (or stale open sessions reclaimed when
[...]

PR #1250, opened 2026-09-18
feat(session): enforce safe closure semantics

jemanuelp:
## 🔗 Linked Issue

Closes #1192

---

## 🏷️ PR Type

- [ ] `type:bug` — Bug fix
- [x] `type:feature` — New feature
- [ ] `type:docs` — Documentation only
- [ ] `type:refactor` — Code refactoring (no behavior change)
- [ ] `type:chore` — Maintenance, dependencies, tooling
- [ ] `type:breaking-change` — Breaking change

---

## 📝 Summary

- Preserve the first session closure and return explicit missing/busy closure errors.
- Map closure failures to HTTP 404/409 responses with focused route coverage.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | Preserve closure state and report invalid closure targets. |
| `internal/server/server.go` | Map closure errors to HTTP responses. |
| `internal/store/store_test.go`, `internal/server/server_test.go` | Cover closure and conflict behavior. |

## 🧪 Test Plan

- [x] Focused tests: `go test ./internal/store -count=1`
- [x] Focused route test: `go test ./internal/server -run TestHandleEndSession -count=1`
- [ ] Full server package: blocked by pre-existing Unix-socket permission failures.
- [x] `git diff --check`

## ✅ Contributor Checklist

- [x] I linked an approved issue above.
- [x] I added exactly one `type:*` label.
- [x] Docs are not needed for this core-only slice.
- [x] Commits follow conventional commits.
- [x] No `Co-Authored-By` trailers.

## Chain Context

| Field | Value |
|-------|-------|
| Chain | OpenCode archive session closure |
| Tracker PR | Not needed |
| Position | 1 of 4 |
| Base | `main` |
| Depends on | None |
| Follow-up | sync delayed closures |
| Review budget | 383 / 400 changed lines |
| Starts at | `main` |
| Ends with | idempotent core closure and HTTP conflict mapping |

### Chain Overview

```text
main
  └── 📍 PR 1: core closure semantics
       └── PR 2: delayed sync closures
            └── PR 3: archive on idle
                 └── PR 4: archive hardening
```

### Scope
- Includes: core store/session closure semantics and endpoint mapping.
- Excludes: delayed sync, archive lifecycle, and plugin hardening.

### Autonomy
- [x] CI is expected to pass for this PR branch.
- [x] This PR has one deliverable scope.
- [x] This PR can be rolled back without unrelated changes.
- [x] Tests cover this unit.

[...]

PR #1252, opened 2026-09-18
fix(session): add session end CLI with bulk dry-run and stale doctor check

### LINKED ISSUE #1266: feat(cli): add safe stale-session cleanup and diagnostics
### 🔍 Problem Description

Long-lived Engram stores can accumulate open runtime sessions whose observations must be preserved. The MCP `mem_session_end` tool can close one known session, but operators outside an agent turn have no supported CLI path to end a confirmed stale session or safely preview a bounded cleanup. Doctor also lacks a dedicated warning before stale open sessions accumulate enough to make omitted-session writes ambiguous.

This is operational containment for #1247, not the owner-liveness root fix. It must remain independently safe even when a genuinely live runtime exists.

### 💡 Proposed Solution

Add an operator-facing session cleanup surface:

- `engram session end <id>` ends one session while preserving its observations, with explicit `ended`, `already_ended`, and `not_found` outcomes.
- `engram session end --by-age <duration> [--project <name>]` previews bounded stale candidates by default and mutates only with `--apply`.
- Bulk apply must always require an explicit positive `--by-age`; omitting the age bound must never select every open session.
- Add a read-only `stale_open_sessions` doctor check that reports old open sessions and points to the bounded dry-run command.
- Keep the operation transactional and journal session-end mutations only when a session actually transitions to ended.

### 📦 Affected Area

CLI (commands, flags)

### 🔄 Alternatives Considered

`mem_session_end` works one session at a time from an agent turn, but it is not an operator CLI and requires the caller to discover IDs manually. `engram delete session` is intentionally unsuitable because sessions with observations must retain those memories. Automatic owner-liveness selection is tracked separately in #1247 and does not replace an explicit recovery command.

### 📎 Additional Context

This issue isolates the cleanup and diagnostic scope already implemented in PR #1252. The root liveness correction remains under #1247. The explicit age requirement captures the safety regression reported in #1253 so cleanup cannot end live sessions by omission.

danielgap:
## 🔗 Linked Issue

Closes #1266

---

## 🏷️ PR Type

---

## 📝 Summary

- Adds `engram session end <id>` as the CLI counterpart of `mem_session_end`: marks a session ended while keeping its observations, with idempotent statuses (`ended` / `already_ended` / `not_found`) and `--summary` / `--json` support. Unlike the store's legacy `EndSession`, the strict path never silently succeeds on an unknown ID and never overwrites an existing `ended_at` or summary.
- Adds bulk cleanup: `engram session end --by-age <dur> [--project X]` selects open sessions by last activity (observation fallback to `started_at`), defaults to a dry-run preview, and only mutates with `--apply`. All ends and their sync-journal entries commit in one transaction.
- Adds a read-only `stale_open_sessions` doctor check (30d activity threshold, per-project counts, oldest/newest) whose safe next step points at the new command, so the accumulation reported in #1247 becomes visible before it breaks `mem_save` with `multiple active runtime sessions`.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/store/store.go` | `EndSessionStrict` (status tri-state, COALESCE summary guard, journals on end only); `StaleOpenSessions` read-only selection (last-activity proxy, LOWER(project) match, NULL-project excluded); `EndSessionsBulk` single-tx bulk end with per-session journal entries |
| `internal/store/store_test.go` | 14 subtests: status paths, ended_at/summary preservation, age boundary, project matching, journal presence |
| `cmd/engram/session.go` | New `session end` command: single/bulk modes, table-tested arg parser rejecting unknown tokens before store open, `72h`/`30d`/`2w` durations, dry-run default with `--apply`, `--json` for both modes |
| `cmd/engram/session_test.go` | 48 subtests: parser tables, 3 single-end paths, dry-run no-mutation, apply, JSON shapes, pre-store-open rejections, dispatch, usage |
| `cmd/engram/main.go` | `session` dispatch case and usage block (mirrors `delete`) |
| `internal/diagnostic/checks.go` | `stale_open_sessions` check: warning severity, evidence with per-project counts/oldest/newest, read-only |
| `internal/diagnostic/checks_test.go` | Scope.Now time-travel tests: stale flagged, fresh by activity not flagged, ended ignored, registry membership |
| `internal/diagnostic/registry.go` | Registers the new check |
| `internal/diagnostic/diagnostic_test.go` | One-line sorted want-list update for the registry ordering test |

## 🧪 Test Plan

- [x] Unit tests pass locally: `go test ./internal/store ./internal/diagnostic` and `go test ./cmd/engram -skip TestCmdServeSignalClosesUnixSocket` (the skipped test is a known environmental failure on this machine: socket parent-hierarchy permissions; it passes under `umask 022`)
- [ ] E2E tests pass locally: `go test -tags e2e ./internal/server/...` — not run; no server or HTTP surface changed (`git diff` confirms `internal/server`, `internal/mcp`, `plugin/` untouched)
- [ ] Lint passes locally: `make lint` — golangci-lint not available locally; `go vet` and `staticcheck` are clean on the three touched packages
- [x] Manually tested the affected functionality: built the binary and exercised all paths against a scratch DB — single end (fresh/already-ended/unknown), bulk dry-run listing without mutation, bulk `--apply` with case-insensitive project match, mutual-exclusion and unknown-flag rejections, doctor check with stale and clean states, sync journal entries for ended sessions (single and bulk), and an already-ended session's original `ended_at`/summary preserved across bulk runs

## New Features

- Added `session end` commands for ending individual sessions or managing stale sessions in bulk.
- Individual sessions support summaries and JSON output; bulk operations support age and project filters, dry-run previews, apply mode, and JSON output.
- Added support for day- and week-based duration values.
- Diagnostics now identify stale open sessions, group findings by project, and provide guidance for resolving them.

## Bug Fixes

- Session-ending results distinguish ended, already-ended, and unknown sessions while preserving existing summaries.


danielgap:
Maintainers: I'm pull-only, so I can't apply the `type:bug` label myself. Could someone add it (commit type is `fix`) so the `Check PR Has type:* Label` gate can pass? Thanks!

danielgap:
Non-blocking follow-up notes from the native review (13 informational advisories; the two substantive ones are now tracked as #1253 — unbounded `--project`-only bulk window — and #1254 — bulk lock hold / stale-scan cost). The remaining polish items, listed so they are not lost; all are small and confined to this PR's surface:

- [ ] `internal/store/store.go` (`staleOpenSessionQuery` comment): a curly quote glyph slipped into "neither `LOWER(NULL)` nor ["] equals a named project" — normalize to ASCII
- [ ] `EndSessionStrict` comment accuracy: the SQL is `summary = COALESCE(?, summary)`, so a provided summary **does** replace a previously stored one when re-reading an open row that already carries a summary (possible via sync import or legacy data). The comment currently says it never clobbers; either fix the comment or, if fill-only was intended, use `COALESCE(summary, ?)` and adjust the test
- [ ] Usage text is duplicated between `printSessionEndUsage` and the `cmdSession` no-args branch — extract one constant/printer
- [ ] The 30d threshold lives three times (`staleOpenSessionsThreshold`, `staleOpenSessionsThresholdDays`, and the literal `720h` inside the SafeNextStep string) — single-source it so the guidance cannot drift from the check
- [ ] `parseSessionEndAge`: compact forms like `999999999d` overflow `float64 -> time.Duration` silently; the `d <= 0` guard only catches negative overflow — reject values beyond `math.MaxInt64` nanoseconds
- [ ] Single-end `--json`: a `storeGetSession` failure after a successful end goes through `fatal`, so the mutation happened but no JSON was emitted — degrade to printing the status with `ended_at` omitted instead of exiting ugly

Happy to take these as a small follow-up PR once #1253/#1254 are triaged.

danielgap:
@dnlrsls ready for your review. This closes #1247 (status:approved, priority:high).

All functional checks are green (unit, E2E, lint, plugin, Windows); the only failing check is "Check PR Has type:* Label", which needs the `type:bug` label applied on your side.

Follow-ups from the native review are already filed as #1253 and #1254, so they don't block this one.

dnlrsls:
The current head already fixes #1253: bulk `session end --project` now requires `--by-age`, and the regression coverage rejects both project-only and project-plus-apply invocations before the store opens.

Please update this branch against current `main` and resolve its conflicts while preserving the guard in `cmd/engram/session.go` and the associated tests. The downstream zero-age store behavior still selects every open session, so losing this CLI validation during conflict resolution would reintroduce the unsafe path. Once the branch is clean, we can re-review it.

danielgap:
Branch updated against `main@9ac5cef` (force-push `e6d1968 → de2eae1`). One real conflict during the rebase, in `internal/store/store_test.go` (both sides appended test blocks at the same anchor) — resolved by keeping both: your six runtime-lease tests first, then the session-end tests.

The `--by-age` guard survived intact: `TestCmdSessionEndRejectsInvalidInvocationsBeforeStoreOpen` still asserts that `--project`-only and `--project --apply` without a window are rejected before the store opens, and it passes on the rebased head. CI is green.

Two pre-existing main-side notes, not introduced here: `gofmt` drift in `internal/diagnostic/diagnostic_test.go` and `cmd/engram/doctor.go` (already on `main`), and `TestCmdServeSignalClosesUnixSocket` is umask-sensitive locally (fails under 0002, passes under 022; CI unaffected). Happy to take either as a separate issue if useful.

danielgap:
Status update: branch updated again, now `023dba8` on current main, keeping the full review round plus the stale-session timestamp normalization and the non-positive window guard. All checks green (unit, E2E, lint, plugin, Windows, CodeRabbit). Still ready whenever you are; closes #1247.

dnlrsls (changes requested):
Thanks for turning the stale-session recovery problem into an explicit, dry-run-first operator workflow. I reviewed immutable head `023dba827c02c16218c71123832a0bac6139f1df` against current `main@70870987f56ae662acf3cfa36fa7e9a99efa9fb3`. The synthetic merge is conflict-free, but the current-main runtime contract exposes two destructive-selection blockers.

## Blocking correctness

1. **Never bulk-end a session with a live runtime lease.**

   `staleOpenSessionQuery` (`internal/store/store.go:3030-3057`) only checks `ended_at IS NULL` and observation/start timestamps. Current main has authoritative `runtime_lease_expires_at` semantics, but this query ignores them, so an old session with an unexpired lease can appear in the dry run and be ended by `--apply`.

   This contradicts approved issue #1266's requirement that cleanup remain safe when a genuinely live runtime exists. Exclude valid unexpired leases using the caller-supplied `now` so preview/apply and tests stay deterministic. Treat a non-empty malformed lease conservatively rather than guessing that the runtime is dead. Add store, CLI dry-run/apply, and doctor coverage.

2. **Malformed activity timestamps must fail closed.**

   `MAX(datetime(o.created_at))` ignores malformed values, then `COALESCE(..., datetime(s.started_at))` treats the session as if it had no observations. An old session whose only/latest observation has an unparseable timestamp can therefore be selected and ended. The comment currently describes the opposite behavior.

   Preserve uncertainty: if any relevant activity value cannot be ordered safely, do not auto-select that session for destructive cleanup. Add malformed observation and malformed session timestamp cases for both preview and apply.

3. **Keep preview and apply on one validated store contract.**

   `EndSessionsBulk` rejects non-positive windows, while public `StaleOpenSessions` does not. The CLI parser currently protects the dry run, but the store APIs can disagree if called elsewhere. Apply the same positive-window validation at the shared selection boundary and test parity.

4. **Restore the standard race gate.**

   The full `go test -race ./internal/store ./internal/diagnostic ./cmd/engram -count=1` command failed on both the candidate and synthetic current-main merge at `TestCmdMCPStdioSIGTERMRunsGracefulShutdown`; the identical full command passes exact current main. The MCP test also passes when isolated on all three trees, so this is a candidate-suite interaction, load, or leaked-state problem rather than a direct MCP behavior regression. Please identify and remove the interaction or make the synchronization deterministic; the eventual combined tree must pass the ordinary race command, not only focused tests.

## Integration ownership

Open PR #1250 already changes the canonical `EndSession` contract to reject missing IDs, preserve terminal timestamps/summaries, journal only real transitions, and classify busy-store failures. This PR adds a parallel `EndSessionStrict` implementation. A synthetic chain is textually conflict-free, but it leaves two close primitives with overlapping semantics and different error behavior.

Please stack the single-session CLI work on the canonical closure contract after #1250 is dispositioned instead of introducing a second store API that can drift.

## Documentation and review size

This adds a destructive operator command but no user-facing command documentation outside `--help`; please document single end, dry-run, `--apply`, project scoping, JSON output, live-lease exclusion, and recovery/rollback expectations in the canonical CLI/operations docs.

At 2,124 changed lines across nine files, this cannot receive a blanket size exception. Split it by behavior with tests kept beside each contract. A safe chain is:

1. canonical single-session closure/CLI on top of #1250;
2. stale selection plus transactional bulk store behavior;
3. bulk CLI and operator documentation;
4. read-only doctor diagnostic.

If one test-heavy slice still crosses 400 lines, request a narrow evidenced exception for that slice rather than the monolith.

## Verification

- synthetic merge with current main: conflict-free (`cd0ce11d6c0162f8cde28298b82e609248fa0603`)
- `git diff --check`: passed
- candidate `go test ./internal/store ./internal/diagnostic ./cmd/engram -count=1`: passed
- candidate `go vet ./internal/store ./internal/diagnostic ./cmd/engram`: passed
- synthetic current-main merge focused tests: passed
- full candidate and synthetic `-race` commands both fail at `TestCmdMCPStdioSIGTERMRunsGracefulShutdown`; the same full command passes exact current main, while the focused MCP test passes on all three trees

danielgap:
Split executed per your chain. All four slices are built, verified, and pushed to the fork; this PR is superseded by the stack below (issue #1266 stays open until the stack lands).

| # | Branch (`danielgap/engram`) | Commit | Lines | Contents |
|---|---|---|---|---|
| 1 | `fix/1247-session-end-single` | `0f7462b` | +631 | single-session end CLI composed on #1250's canonical contract (`GetSession` pre-check, `EndSession` transition, authoritative re-read; no `EndSessionStrict` anywhere in the chain) |
| 2 | `fix/1247-stale-bulk-store` | `221381c` | +549 | stale selection (`StaleOpenSessions`) + transactional `EndSessionsBulk` with per-row journaling and window guard |
| 3 | `fix/1247-bulk-cli-docs` | `24bf92c` | +472/−21 | bulk CLI (dry-run default, explicit `--apply`, `--json` envelopes, `--by-age`/`--project`) + operator docs (CLI reference entry + Session End CLI section) |
| 4 | `fix/1247-stale-doctor` | `c05f435` | +271/−1 | read-only `stale_open_sessions` doctor check + DOCTOR.md catalog entry |

Slices 1-3 are test-heavy and cross the 400-line budget (422, 387, and ~330 test lines respectively of each total). Requesting the narrow evidenced exception you allowed per slice rather than the monolith; slice 4 lands at 272.

Sequencing: fork PRs can only target upstream branches, so the stack opens sequentially. PR 1 targets `main` as soon as #1250 is dispositioned, since its composition already speaks that contract (`ErrSessionNotFound`, `ErrSessionBusy`, already-ended short-circuit with preserved terminal state). Each later PR opens when its parent merges, retargeted so only its own slice shows.

Verification at the chain head: `go build ./...`, `go vet` on all three touched packages, and `go test ./internal/store ./internal/diagnostic ./cmd/engram -count=1` all pass (one known local socket-permission test skipped locally, unrelated).

The four blockers land as focused commits on their owning slices: live-lease exclusion in the stale selection (slice 2, once the chain rebases onto current main's `runtime_lease_expires_at`), malformed-timestamp fail-closed (slice 2), shared positive-window validation (slice 2), and the MCP race-gate interaction (chain level).

Issue #1269, opened 2026-09-18
Pi memory facade cannot resolve local server identity while MCP remains healthy


despinosaqualit:
### 📝 Bug Description

In Pi, the injected `mem_*` facade fails to resolve the local Engram server identity, even though the same local Engram instance is reachable and healthy through the MCP adapter. This prevents the standard memory facade from saving session summaries and observations, while an MCP-tool fallback succeeds.

### 🔄 Steps to Reproduce

1. Install Engram for Pi and start a Pi session with the Engram integration enabled.
2. Invoke an injected memory operation such as `mem_save` or `mem_session_summary`.
3. Observe that the operation fails because the local server identity cannot be resolved.
4. Verify that the local endpoint is listening, connect to the `engram` MCP server, and invoke `engram_mem_doctor` or `engram_mem_session_summary`.
5. Observe that the MCP connection and doctor check succeed, and the MCP summary save succeeds.

### ✅ Expected Behavior

The injected Pi memory facade and the MCP adapter should resolve the same healthy local Engram instance. A successful MCP connection should not coexist with an identity-resolution failure from `mem_save` or `mem_session_summary`.

### ❌ Actual Behavior

The injected facade reports that it cannot resolve the local Engram server identity and refuses the save. The connected MCP server remains usable and its doctor check reports no errors.

### Operating System

Windows (WSL)

### Engram Version

Engram 1.20.0; gentle-engram Pi extension 0.1.14; pi-mcp-adapter 2.34.0.

### Agent / Client

Other

### 📋 Relevant Logs

```shell
Injected memory facade: could not resolve the local Engram server identity.
MCP connection: successful.
Engram MCP doctor: healthy; no warnings or errors.
```

### 💡 Additional Context

The failure occurred on every attempted direct `mem_save` and `mem_session_summary` call in one Pi session. A summary was subsequently persisted through the connected Engram MCP tool. No credentials, user paths, repository names, or local endpoint addresses are included here.

danielgap:
Taking this one, and crediting @despinosaqualit for the report.

The version pin in your report (Engram 1.20.0 server with the v2-capable gentle-engram 0.1.14 facade) places this squarely on the identity surface already mapped in #1255: `probeEngramHealth()` and `localInstanceID()` in `plugin/pi/index.ts` resolve server identity without a version-aware path, so a healthy local endpoint can still be refused by the injected facade while the MCP adapter succeeds.

Relationship to the open work:

[...]

PR #1279, opened 2026-09-19
fix(pi): classify legacy pre-v2 servers and retry deterministic failures on a fixed 5s window

### LINKED ISSUE #1255: bug(pi): stale pre-v2 server is reported as ownership mismatch after upgrade
### 📝 Bug Description

After upgrading the Engram CLI from 1.20.0 to 2.0.0, an already-running detached `engram serve` process can remain on 1.20.0. The Pi plugin (`gentle-engram` 0.1.14) probes that server with the v2 ownership check. Because the legacy `/health` response has no `instance_id`, the plugin classifies it as a foreign server and reports only:

`Engram server ownership mismatch at http://127.0.0.1:7437`

The safety check is correct, but the diagnostic does not distinguish a legacy server left running after an upgrade from a genuinely foreign v2 server. Memory remains unavailable until the user discovers and restarts the old process manually.

### 🔄 Steps to Reproduce

1. Install Engram 1.20.0 and let the Pi plugin start its detached `engram serve` process on the default port.
2. Upgrade the Engram CLI to 2.0.0 and the Pi plugin to `gentle-engram` 0.1.14 without terminating the existing server process.
3. Start or reload Pi and invoke any `mem_*` tool.
4. Observe `Engram server ownership mismatch at http://127.0.0.1:7437`.
5. Compare the installed CLI version with `GET /health`: the CLI is 2.0.0, while the running server is still 1.20.0 and its response has no `instance_id`.
6. Stop the 1.20.0 process and start Engram 2.0.0. Memory works again after the plugin's startup retry backoff expires.

### ✅ Expected Behavior

The plugin should continue to fail closed, but distinguish a legacy health response from a mismatched v2 `instance_id`. The error should state that an older Engram server is still running, include the detected server and CLI versions when available, and provide safe restart guidance. It should not terminate an unverified process automatically.

After the server is replaced with the expected v2 instance, recovery should be immediate or the remaining retry delay should be made explicit.

### ❌ Actual Behavior

`probeEngramHealth()` reads only the optional `instance_id` and maps both a missing ID and a different ID to `foreign`. `initializeEngramServer()` then throws the generic ownership-mismatch error, with no indication that the listener may be a pre-v2 process left behind by an upgrade.

After manually replacing the server, the cached startup failure can continue to be returned until the exponential retry backoff expires (up to 60 seconds), making the successful repair appear ineffective temporarily.

### Operating System

macOS

### Engram Version

2.0.0 (running stale server: 1.20.0)

### Agent / Client

Other

### 📋 Relevant Logs

```shell
Engram server ownership mismatch at http://127.0.0.1:7437
```

### 💡 Additional Context

This is a follow-up to #1120 rather than a request to weaken its ownership protection. That issue correctly added `instance_id` verification so adapters do not adopt another user's server. This report covers the upgrade path where the responding process is the same user's pre-v2 server and therefore cannot provide the new identity field.

A safe implementation could add a separate `legacy`/`version-mismatch` health classification when `/health` succeeds without `instance_id`, then emit actionable restart guidance while preserving the existing hard failure for a non-matching v2 identity.

### LINKED ISSUE #1256: Pi provider 0.1.14 broken against server v2: 'instance-id' command no longer exists
## Summary

Pi memory tools (`mem_save`, `mem_session_summary`, `mem_doctor`, …) are fully broken against engram server v2: every call fails with `Engram could not resolve its local server identity`, even with a healthy server answering `/health` 200.

## Root cause

`plugin/pi/index.ts` → `localInstanceID()` shells out to `ENGRAM_BIN instance-id` and requires a 32-hex-char id:

```ts
const result = spawnSync(ENGRAM_BIN, ["instance-id"], ...);
const id = result.status === 0 ? result.stdout.trim() : "";
if (!/^[a-f0-9]{32}$/.test(id)) throw new Error("Engram could not resolve its local server identity");
```

But engram `v2.0.0-rc.10` has **no `instance-id` command** (`unknown command: instance-id`), so the id is always empty and every memory tool fails closed. Additionally, `ENGRAM_BIN`/`ENGRAM_URL` are unset in a default Pi session, so even discovery relies on ambient PATH.

## Environment

- `gentle-engram 0.1.14` (Pi provider)
- `engram v2.0.0-rc.10` (`engram serve`, data dir via `ENGRAM_DATA_DIR`, `/health` → 200)
- Server was briefly on a stale v0.1.0 binary (separate ops issue, since fixed) — the failure reproduces identically against the correct v2 server, proving this is a client/server contract break, not a stale-server issue.

## Expected

- Pi provider works against engram v2 out of the box: either resolve instance identity through a v2-supported surface (e.g. `/health` version/identity field, `engram version`, or a new `instance-id` equivalent), or drop the spawn-based identity gate where the HTTP server is already proven healthy.
- Document required env (`ENGRAM_BIN`, `ENGRAM_URL`, `ENGRAM_DATA_DIR`) for Pi sessions.

## Workaround used

Bypassed the Pi tools entirely and saved via CLI against the running server:

```bash
export ENGRAM_DATA_DIR=/srv/fortis/data/engram
engram save "title" "msg" --type decision --project ativva
```

## Impact

Any Pi user on engram v2 has zero persistent-memory capability through the native tools (save, session summary, doctor all fail with the same identity error).

danielgap:
## 🔗 Linked Issue

Closes #1255
Closes #1256

Requested scope split per the approval thread: this PR closes #1255 and the remaining #1256 slice (environment documentation plus the pre-rc.11 CLI guidance wording; the original v2 `instance-id` regression was already fixed by #1127 and contained in v2.0.0-rc.11+). It intentionally leaves the broader Pi/MCP same-instance behavior in #1269 open.

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix

---

## 📝 Summary

- `probeEngramHealth()` gains version-aware `legacy` and `identity_missing` states: only recognized SemVer releases older than `2.0.0-rc.11` are legacy; current, unknown, absent, and malformed versions without identity fail closed with distinct identity-verification guidance. The `foreign` message stays byte-identical and the `expectedID === ""` path is untouched.
- Legacy and ownership failures fail closed before any spawn with two new actionable messages (legacy-server guidance reporting detected server/CLI versions, degrading to `unknown`; pre-rc.11 CLI upgrade guidance, with a distinct not-found message for a missing binary).
- Deterministic startup failures (ownership, legacy, identity resolution) recheck on a fixed 5 s window instead of the exponential curve, and the existing 5 s self-heal probe now clears the retry window on observed identity; transient failures keep the exponential backoff.
- README gains an environment variables section (`ENGRAM_URL`, `ENGRAM_BIN`, `ENGRAM_PORT`, `ENGRAM_DATA_DIR`).

The A/B reproduction against a server built from the `v1.20.0` tag remains recorded in #1269; that broader same-instance behavior is explicitly out of scope for this PR.

---

## 📂 Changes

| File | Change |
|------|--------|
| `plugin/pi/index.ts` | Version-aware `legacy`/`identity_missing` classification + server-version capture in `probeEngramHealth()`; `DeterministicStartupError`; distinct legacy, missing-identity, pre-rc.11, and ENOENT guidance; fail-closed ownership branches in `initializeEngramServer()`; fixed 5 s deterministic recheck; self-heal retry-window clearing |
| `plugin/pi/test/index-source.test.mjs` | Probe-matrix, wording, cadence, and window-clearing unit tests |
| `plugin/pi/test/startup-lifecycle.test.mjs` | Behavioral tests against real fake binaries (legacy fail-closed, pre-rc.11 guidance, missing binary, foreign byte-identical) |
| `plugin/pi/README.md` | Environment-variable reference plus version-aware missing-identity guidance |

Current diff: +540/−29 across four files. The approved design note in #1255 explicitly scoped the behavior, tests, and README slice as one cohesive unit; a maintainer approved the `size:exception` for this PR.

---

## 🧪 Test Plan

- [x] Plugin suite: `cd plugin/pi && npm test` — 156 passed, 0 failed, 0 skipped
- [x] E2E: `go test -tags e2e ./internal/server/...` — passes under `umask 022` except `TestProjectCurrentDoctorJudgeAndCompareRoutes`, which fails identically on base `9ac5cef` (pre-existing on main, unrelated to this plugin-only change); the two Unix-socket permission tests are umask-sensitive and pass under `umask 022`
- [ ] Lint: not runnable locally (`golangci-lint v2.13.2` not on PATH); this PR touches zero Go lines and CI will run it
- [x] Manual verification: A/B repro in #1269 (v1.20.0-tag server vs main), plus independent read-only verification of all design dimensions (classification matrix, byte-exact templates, fixed cadence via injected clock, self-heal window clearing, transient backoff pinned unchanged)

---

## ✅ Contributor Checklist

- [x] I linked an approved issue above (`Closes #N`)
- [x] I added exactly **one** `type:*` label to this PR (`type:bug`)
- [x] I ran e2e tests locally: `go test -tags e2e ./internal/server/...` (see the pre-existing base failure noted above)
- [ ] I ran lint locally: `make lint` — golangci-lint not installed in this environment; CI covers it
- [x] Docs updated (README environment section)
- [x] Commits follow conventional commits format (`fix(pi): ...`)
- [x] I checked every changed path against the Transient Artifact Policy (four tracked files only)

---

## 💬 Notes for Reviewers

**Correction review provenance.** Native review completed for correction commit `7ea47f9` against prior PR head `c6f177e`: lineage `review-555945af26092eaa` reached approved state across risk, resilience, readability, and reliability lenses and was acknowledged with no blocking findings.

**One accepted wording deviation.** For a missing binary (ENOENT) the design said "keep an accurate not-found message"; the implementation adds `The Engram binary "<ENGRAM_BIN>" could not be found. Install Engram, or point ENGRAM_BIN at the current binary.` rather than reusing the generic identity error. It is a one-line revert in `instanceIDFailureMessage()` if you prefer the old wording.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of older or incompatible Engram servers and binaries, with clearer upgrade guidance.
  * Prevented interaction with servers that cannot confirm ownership or identity.
  * Added consistent five-second retry intervals for deterministic startup failures.
  * Improved recovery when the expected server identity becomes available again.

* **Documentation**
  * Documented Pi plugin environment variables, defaults, and effects on server startup, binary discovery, health checks, and data storage.


danielgap:
Thanks for both findings — addressing them separately:

**Instance-id failure diagnostics (line 418)** — addressed in a61bfe4. `instanceIDFailureMessage` now returns five distinct diagnoses: missing binary (ENOENT, wording unchanged), timeout, spawn errors with the errno echoed (covers EACCES), and a generic nonzero-exit fallback that names the exit code. The pre-v2.0.0-rc.11 upgrade message is now returned **only** for an explicit unsupported-command result: a nonzero exit whose stderr names both `unknown command` and `instance-id` (the shape every pre-rc.11 binary emits for this dispatch, including the fake binary in the behavioral tests). ENOENT and legacy wording are byte-identical to before; three new wording tests cover the new shapes, and the untouched behavioral tests still pass.

**Moving server ownership / compatibility classification / startup policy into the core Go API (lines 454-459)** — I'd like to push back on this one, for a reason specific to this bug class: the process being classified is by definition a **pre-v2 binary that cannot gain new code**. No "core Go API" running on the stale 1.20.0 server can ever classify itself as legacy — the only component that can distinguish "pre-v2 server without identity" from "foreign v2 server with a different identity" is the newer client probing `/health`. That is exactly the classification this PR adds, per the approved design in #1255 (fail closed, distinguish legacy from foreign, actionable restart guidance; the approval thread scoped "the PR with tests and the README slice in one unit").

The retry policy has the same constraint from the other side: the fixed 5s window for deterministic failures and the exponential backoff for transient ones are **in-process adapter state** feeding `sharedInitialization` and the self-heal probe (which clears the retry window on observed identity). Moving the decision into a one-shot CLI invocation would split the policy from the state it governs without removing any logic. Within the probe itself the adapter stays a thin mapper: fetch `/health`, classify, map to an actionable message.

Happy to revisit if there's a core-side shape you'd recommend for the client-side probe (e.g. a `probe` subcommand the adapter shells out to) — but that reads as a redesign of the approved scope rather than a fix to this PR, so I'd track it as a separate issue.

PR #1285, opened 2026-09-19
fix(autosync): run pull despite non-enrolled pending mutations (#1273)

### LINKED ISSUE #1273: Autosync skips the pull step while non-enrolled projects have pending mutations, blocking cross-machine replication
### 📝 Bug Description

When cloud autosync is enabled, the local daemon journals sync mutations for EVERY local project, enrolled or not. During each autosync cycle, if there is nothing to push for enrolled projects but there are pending mutations belonging to non-enrolled projects, the cycle records a degraded/blocked state and returns WITHOUT running the pull step.

Consequence: on a machine that mostly consumes a shared project (no local writes to enrolled projects), inbound replication never happens automatically. New observations pushed from another machine only arrive after a manual `engram sync --cloud --import --project <p>`.

This punishes a common workflow: users with many local throwaway projects that they deliberately do NOT enroll. Enrollment is documented as the explicit opt-in, but un-enrolled projects end up blocking replication of the enrolled ones.

Root cause (verified in source): in internal/cloud/autosync/manager.go, push() returns a nonEnrolledPendingError when there are no pushable enrolled mutations but CountPendingNonEnrolledSyncMutations > 0; the cycle then calls recordBlocked() and returns before m.pull(ctx) runs.

### 🔄 Steps to Reproduce

1. Run `engram cloud serve` (authenticated mode) and set up two machines A and B with autosync enabled (ENGRAM_CLOUD_AUTOSYNC=1, ENGRAM_CLOUD_TOKEN, ENGRAM_CLOUD_SERVER).
2. Enroll one project P on both machines; bootstrap it from A (engram cloud upgrade bootstrap). Verify sync works.
3. On machine B, make sure there are OTHER local projects with saved memories that are NOT enrolled (their mutations get journaled automatically).
4. On machine A, save a new observation in P. The push succeeds (it reaches the cloud dashboard).
5. On machine B, wait several autosync cycles: the new observation never arrives. GET /sync/status shows phase "degraded", reason_code "non_enrolled_pending_mutations", and last_sync_at frozen. Running `engram sync --cloud --import --project P` manually imports it instantly.

### ✅ Expected Behavior

Enrollment should be the explicit opt-in boundary: pending mutations of non-enrolled projects may be reported (informational), but they should not prevent the pull step for enrolled projects. The cycle should record the blocked-push state and still run pull.

### ❌ Actual Behavior

The cycle aborts after the push stage. In internal/cloud/autosync/manager.go (cycle): when push() returns nonEnrolledPendingError, the manager calls recordBlocked(...) and returns; m.pull(ctx) is skipped. On a consumer machine this means autosync never pulls, indefinitely, while any non-enrolled project has queued mutations. Verified in the v1.20.0 tag (recordBlocked at manager.go:414-417, pull at :424 unreachable in that branch) and still present on main as of 96ea92a (same logic at :494-504).

### Operating System

macOS

### Engram Version

1.20.0

### Agent / Client

Claude Code

### 📋 Relevant Logs

```shell
GET http://127.0.0.1:7437/sync/status?project=P
{
  "enabled": true,
  "phase": "degraded",
  "reason_code": "non_enrolled_pending_mutations",
  "reason_message": "pending cloud sync mutations are blocked because project(s) are not enrolled: projA=161, projB=189, projC=1672, ... Run `engram cloud enroll <project>` for each intended project or review enrollment.",
  "last_sync_at": "2026-09-17T15:00:58-06:00"   <- frozen; never advances while degraded
}
```

### 💡 Additional Context

- Reproduced on macOS (launchd service) and Ubuntu 24.04 (systemd user service), both engram 1.20.0 via Homebrew, against a self-hosted cloud server (ghcr.io image 1.20.0) with ENGRAM_CLOUD_ALLOWED_PROJECTS=*.
- The push side is unaffected: a machine that writes to an enrolled project pushes fine (pending enrolled mutations exist, so push completes and pull runs that cycle). Only consumer machines stall.
- Workaround: enroll every actively used project on each machine and/or run explicit `engram sync --cloud --import --project <p>`.
- Suggested fix: on nonEnrolledPendingError, record the blocked reason but continue to m.pull(ctx). Optionally make the non-enrolled backlog purgeable or informational, since journaling for never-enrolled throwaway projects grows forever.
- Still reproducible on main as of 96ea92a. Note that #1234 (2cdda90) adds a doctor repair path that can supersede obsolete un-enrolled journal rows, which drains the backlog, but the cycle still skips pull whenever NEW mutations for non-enrolled projects appear, so the underlying behavior remains.

josegarvinvictoria:
## 🔗 Linked Issue

Closes #1273

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix

---

## 📝 Summary

- Autosync `cycle()` did an early return after `recordBlocked` when `push()` returned `*nonEnrolledPendingError`, skipping `m.pull(ctx)` entirely.
- On consumer machines with local non-enrolled projects that have pending mutations, inbound replication for enrolled projects froze indefinitely (`last_sync_at` frozen, phase `degraded` with `reason_code: non_enrolled_pending_mutations`).
- The cycle now captures the blocked push without early-returning, always runs `pull()`, retains the degraded blocked status with enrollment guidance, and advances `LastSyncAt` when inbound sync succeeds — in memory (`GET /sync/status`) and persisted (`sync_state.last_success_at`, affecting `engram cloud status`/TUI) without marking the target healthy. A pull transport failure is still classified and recorded as a failure.

## 📂 Changes

| File | Change |
|------|--------|
| `internal/cloud/autosync/manager.go` | `cycle()` no longer early-returns on `nonEnrolledPendingError`; runs `pull()` first, then records blocked+`LastSyncAt` on success via `recordBlockedWithInboundSuccess` (new optional `blockedSuccessReportingStore` hook) or the classified transport failure on pull error |
| `internal/cloud/autosync/manager_test.go` | Updated `TestManagerBlocksWhenOnlyNonEnrolledPendingMutationsRemain` (pull now runs, `LastSyncAt` set); added `TestManagerPullsAndAppliesMutationsWhenNonEnrolledPendingMutationsExist` (pull applies mutations, `LastSyncAt` set) and `TestManagerPullFailureOverridesNonEnrolledBlocked` (`LastSyncAt` stays nil) |

## 🧪 Test Plan

- [x] Unit tests pass locally: `go test ./...` (only the pre-existing macOS baselines fail: `cmd/engram`/`internal/server` unix-socket `invalid argument`, `internal/setup` encoding, `internal/version` update, `plugin` hook/unix-socket — verified identical on `origin/main` via a temp worktree)
- [x] Targeted: `go test ./internal/cloud/...` all green
- [x] `gofmt`/`go vet` clean on changed files

Note: `status:approved` cannot be added by external contributors; the issue is currently `status:needs-review`. A maintainer will need to approve #1273 (confirmed reproducible in the issue comments) for the Issue-Approved check to pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sync cycles now continue pulling changes when pushes are blocked by pending mutations from projects that are not enrolled.
  * Successfully pulled changes are applied, the last sync time is updated, and the blocked status and enrollment guidance are preserved.
  * Pull failures record the appropriate authentication error and backoff behavior without marking the sync as successful.
  * Blocked syncs that complete a successful pull now retain their latest success timestamp.


josegarvinvictoria:
Closing this respectfully: reviewing the full issue thread, the maintainer (@Rafaeldelinares) already reproduced the bug and announced their own fix — including the proposed change and reproduction tests — offering to open the PR themselves pending `status:approved`. This PR was opened without first checking the thread and duplicates that announced work, which is not our intent. Withdrawing to leave the issue to the maintainer's branch; happy to help review whenever it is up. Apologies for the noise.

